PR #224, opened 2025-08-19
feat!: external node can read previous replay version

joonazan:
Put the whole ReplayRecord wire format in one file and diff that file.

If there have been changes to the file after the previous version tag, the wire format version must be one larger than at the version tag.


lib/storage_api/src/replay_wire_format/mod.rs
```diff
@@ -0,0 +1,32 @@
+use crate::ReplayRecord;
+
+mod conversion;
+
+// Don't change the file even if we update formatting rules
+#[rustfmt::skip]
```
RomanBrodetski, line 6: good catch!

lib/storage_api/src/replay_wire_format/v1.rs
```diff
@@ -0,0 +1,74 @@
...
+//! (This is enforced by CI)
+
+use bincode::{Decode, Encode};
+
+// It is somewhat safe to assume that these will not change
+use alloy::primitives::{B256, U256, ruint::aliases::B160};
+
+/// The format ReplayRecords are currently sent in
+#[derive(Encode, Decode)]
+pub struct ReplayWireFormat {
```
RomanBrodetski, line 15: I would call it `ReplayWireFormatV1` to be more explicit.. Otherwise we'd had files in different modules but with the same name - confusing?

RomanBrodetski (commented):
Overall I'm willing to give it a try! 
Although I want @perekopskiy to review and think it thourgh as well -thanks

Issue #292, opened 2025-08-28
Add/refine merics for RPC


RomanBrodetski:
Make sure we have exactly one `tracing::debug!` per rpc invocation. There may be additional `tracinfg::trace!`. In that `debug!`, please log and add to metrics:
  * size of input
  * maybe some input meta? (e.g. block number etc)
  * size of output
  * time it took to process this request 

Additionally, use different `target` for different RPCs so that they can be enabled/disabled individually, e.g. `tracing::debug!(target: "rpc::eth::call", ...)`

If any of this is too cumbersome to infer, feel free to skip.
Make sure the new metrics are reflected in grafana (reach out to @RomanBrodetski for coordination)

PR #306, opened 2025-09-01
feat: RPC monitoring middleware

dimazhornyk:
Adds a middleware to log report metrics about every RPC request #292


RomanBrodetski (dismissed):

lib/rpc/src/middleware.rs
```diff
@@ -0,0 +1,99 @@
...
+    ) -> impl Future<Output = Self::MethodResponse> + Send + 'a {
+        let method = request.method_name().to_owned();
+        let fut = self.inner.call(request);
+
+        async move {
+            let started = Instant::now();
+            let out = fut.await;
+            let output_size = out.as_json().get().len();
+
+            log_and_report(CallKind::Call, &method, started.elapsed(), output_size);
```
RomanBrodetski, line 43: can we also track the request size? Not overly important, may be handy if there is some malicious activity involving very large requests?
So I'd say only add it if it's trivial to add

lib/rpc/src/middleware.rs
```diff
@@ -0,0 +1,99 @@
...
+        requests: Batch<'a>,
+    ) -> impl Future<Output = Self::BatchResponse> + Send + 'a {
+        let fut = self.inner.batch(requests);
+
+        async move {
+            let started = Instant::now();
+            let out = fut.await;
+            let output_size = out.as_json().get().len();
+
+            log_and_report(CallKind::Batch, "batch", started.elapsed(), output_size);
```
RomanBrodetski, line 59: very interesting! Is this called on a batched rpc request?? I was thinking RPC batch is just a collection of method invocations - so I'd expect them to be tracked separately. But here you have just one `fut` still? I suppose it iterates over the inner requests somehow inside?

This makes our tracking work poorly for batched requests - they won't even be counted towards the individual methods. Can we somehow "open" that batch here?

I'd also be interested in some batching-specific metrics like average/heatmap number of method invokcations per batch.

Or maybe I completely misunderstood what bathc is :)

lib/rpc/src/middleware.rs
```diff
@@ -0,0 +1,99 @@
...
+            out
+        }
+    }
+
+    fn notification<'a>(
```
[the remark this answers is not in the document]
dimazhornyk: this is the same as `call` but it doesn't return the `id` to the client, basically used for fire-and-forget scenarios

I doubt that this is widely used (same as `batch`), but we have to support this interface if we want the access to a lower-level info about the requests and responses

lib/rpc/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+extern crate core;
```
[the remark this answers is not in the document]
dimazhornyk: removed

lib/rpc/src/middleware.rs
itegulov: I am pretty sure you can achieve this if you re-implement `Monitoring::batch` by copying `RpcService::batch` implementation and adapting it to call `Monitoring::call` instead of `RpcService::call`

itegulov (approved):

PR #328, opened 2025-09-03
chore: Change config option name in running_with_l1.md

mm-zk:
And added info about update_state_json tool


RomanBrodetski (approved):
oops thanks for the fix!

PR #345, opened 2025-09-09
feat: zksync os inteface/multivm

perekopskiy:


node/bin/src/block_replay_storage.rs
```diff
@@ -12,8 +12,7 @@ use tokio::pin;
 use tokio::time::{Instant, Sleep};
 use vise::Unit;
 use vise::{Buckets, Histogram, Metrics};
-use zk_ee::system::metadata::BlockMetadataFromOracle;
-use zk_os_forward_system::run::BlockContext;
+use zksync_os_interface::types::BlockContext;
```
perekopskiy, line 15: `BlockContext` has changed, it's a breaking change, so we will need to migrate storage:
- re-sync one of ENs with new image
- rotate main node
- resync other nodes

mm-zk (approved):

lib/storage_api/src/replay_wire_format/v2.rs
```diff
@@ -0,0 +1,75 @@
+//! We need to not accidentally change the replay wire format
```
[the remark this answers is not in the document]
RomanBrodetski: I want to add a CI job that would try to sync an EN against some real envs that still have older versions of replay wire format.

If you mean just for the current version - it's tested in int. tests (there is an EN sync test)

lib/storage_api/src/replay_wire_format/v2.rs
```diff
@@ -0,0 +1,75 @@
...
+//! but there is no way in Rust to get a stable unique ID for a type,
+//! so instead we define it in this separate file.
```
RomanBrodetski, line 3:
```suggestion


```

Issue #331, opened 2025-09-03
health endpoints and graceful shutdown


RomanBrodetski:

RomanBrodetski:
Only limited functionality - as discussed in DM with @dimazhornyk

PR #348, opened 2025-09-10
feat: delay the termination, expose health endpoint

dimazhornyk:
Add a status server, which only has `/health` endpoint for now
Exit the program after a delay to allow the load balancer to switch the traffic
#331


node/bin/src/main.rs
```diff
@@ -14,6 +16,8 @@ use zksync_os_observability::PrometheusExporterConfig;
 use zksync_os_state::StateHandle;
 use zksync_os_state_full_diffs::FullDiffsState;
 
+const GRACEFUL_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(10);
```
RomanBrodetski, line 19: maybe make confiugrable? I can imagine locally we'd want to wait less than on real envs

node/bin/src/main.rs
```diff
@@ -44,14 +48,57 @@ pub async fn main() {
...
 
+    let stop_receiver_copy = stop_receiver.clone();
+
     tokio::select! {
-        _ = main_task => {},
+        _ = main_task => {
+            if *stop_receiver_copy.borrow() {
+                tracing::info!("Main task exited gracefully after stop signal");
+                // sleep to wait for other tasks to finish
+                tokio::time::sleep(GRACEFUL_SHUTDOWN_TIMEOUT).await;
```
RomanBrodetski, line 58: hmm so what exactly does this timer do? I thought components there don't listen to the `stop_receiver` - otherwise they'd exit without waiting for the timeout? 

Either way, I don't see what we are waiting for on this line...

RomanBrodetski: Ah, so it's probably applied IN CASE this main_task listens to the stop_receiver and shuts down immediately. But I think this timer doesn't help anyway because RPC is not running anymore in this case?

RomanBrodetski (approved):

PR #349, opened 2025-09-10
feat: have all user-facing config values in one file

RomanBrodetski:
As discussed on the standup, we want to have one file where all the config fields with their defaults can be seen.
Individual components can still have their own configs - but if they have user-defined values, they should be duplicated in this file. Thus, component-level config serves as the component input/config itself, whereas node/bin/config.rs serves as the general, comprehensive node configurtion.


lib/l1_sender/src/config.rs
```diff
@@ -1,55 +1,20 @@
 use alloy::consensus::constants::GWEI_TO_WEI;
 use serde::{Deserialize, Serialize};
-use smart_config::value::SecretString;
-use smart_config::{DescribeConfig, DeserializeConfig};
```
itegulov, line 4: Let's get rid of smart-config deps in all `Cargo.toml`s except for bin -- to make sure we deleted all irrelevant entries

lib/l1_sender/src/config.rs
```diff
@@ -1,55 +1,20 @@
...
-    /// Can be arbitrary funded address - proof submission is permissionless.
-    // TODO: Pre-configured value, to be removed
-    #[config(default_t = "0x1590cd15674138453a4ae04393c41ea4f21306ef9cf5f2ba326d757d4d97bb8a".into())]
-    pub operator_prove_pk: SecretString,
-
-    /// Private key to use to execute batches on L1
-    /// Can be arbitrary funded address - execute submission is permissionless.
-    // TODO: Pre-configured value, to be removed
-    #[config(default_t = "0x1bf13db0eca84058660ab17913b8f6342361b20847d613b4dc241145620cc593".into())]
-    pub operator_execute_pk: SecretString,
```
itegulov, line 28: Let's leave one pk here maybe? We pass it separately from config but makes sense as a part of config IMHO

PR #370, opened 2025-09-15
feat: support L1->L2 tx gas estimation

### LINKED ISSUE #154: Estimate gas for L1->L2 transactions
We used to have a special RPC method for this in Era: `zks_estimateGasL1ToL2`. In ZKsync OS though regular `eth_estimateGas` should work for both L2 and L1->L2 transaction types. Antonio confirmed that should be the case as well from zksync-os POV.

This ticket is currently blocked by the intrinsic gas being too low making L1->L2 transactions way cheaper than they should be (e.g., deposits only consume 11k gas)

itegulov:
Fixes #154

### DIFF
--- a/lib/rpc/src/eth_call_handler.rs
+++ b/lib/rpc/src/eth_call_handler.rs
@@ -18,7 +18,7 @@ use zksync_os_interface::{
 use zksync_os_storage_api::ViewState;
 use zksync_os_storage_api::{RepositoryError, StateError};
 use zksync_os_types::{
-    L1Envelope, L1PriorityTxType, L1Tx, L1TxType, L2Envelope,
+    L1_TX_MINIMAL_GAS_LIMIT, L1Envelope, L1PriorityTxType, L1Tx, L1TxType, L2Envelope,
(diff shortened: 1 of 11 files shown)


lib/rpc/src/eth_call_handler.rs
```diff
@@ -407,7 +443,8 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
                 }
                 Err(
                     InvalidTransaction::CallGasCostMoreThanGasLimit
-                    | InvalidTransaction::OutOfGasDuringValidation,
+                    | InvalidTransaction::OutOfGasDuringValidation
+                    | InvalidTransaction::OutOfNativeResourcesDuringValidation,
```
RomanBrodetski, line 447: this method is now over 200 LoC. Others are over 100 in this file.
Not necessarily in this PR, but  maybe makes sense to refactor this module? At the very least extract some functions, but I think some new local abstractions can do well here.

RomanBrodetski (approved):

PR #404, opened 2025-09-22
feat: make batcher skip blocks that are already processed

RomanBrodetski:


node/bin/src/batcher/mod.rs
```diff
@@ -94,14 +106,17 @@ impl Batcher {
...
+                batch_metadata = ?batch_envelope.batch,
                 block_count = batch_envelope.batch.last_block_number - batch_envelope.batch.first_block_number + 1,
                 new_state_commitment = ?batch_envelope.batch.commit_batch_info.new_state_commitment,
                 "Batch created"
             );
 
+            tracing::debug!(
+                da_commitment=?batch_envelope.batch.commit_batch_info.operator_da_input,
+                "Batch da_input",
+            );
```
perekopskiy, line 118: nit: let's maybe include batch_number here as well

perekopskiy (approved):

PR #407, opened 2025-09-23
feat: add logging configuration (json/terminal/logfmt)

itegulov:
~Additionally, I added custom wrapper macros that replace `zksync_os_` with `zk::` in target for all logs. This is controversial and I am not 100% happy with how it turned out: a lot of boilerplate and it requires `proc_macro_expand` nightly feature (pulled through `string-literal-const-replace`). Still, I migrated L1 sender just to showcase how it works:~
image

~Let me know what you think, happy to rollback the last commit and just leave logging configuration~

UPD: Remove above, left just logging configuration in this PR

### DIFF
--- a/node/bin/src/config.rs
+++ b/node/bin/src/config.rs
@@ -382,10 +382,6 @@ pub struct LogConfig {
     #[config(with = Serde![str])]
     pub format: LogFormat,
 
-    /// Default log directive in format that is used in `RUST_LOG`. Applied before directives from `RUST_LOG`.
-    #[config(default_t = "info".into())]
-    pub default_directive: String,
-
     /// Whether to use color in logs.
     #[config(default_t = true)]
     pub use_color: bool,
--- a/lib/tracing/src/lib.rs
+++ b/lib/tracing/src/lib.rs
@@ -5,7 +5,6 @@ pub use formatter::LogFormat;
 
 use tracing::level_filters::LevelFilter;
 use tracing_subscriber::EnvFilter;
-use tracing_subscriber::filter::Directive;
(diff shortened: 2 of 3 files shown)


lib/tracing/src/macros.rs
```diff
@@ -0,0 +1,512 @@
...
+            $crate::__tracing::Level::ERROR,
+            $($arg)+
+        )
+    );
+}
+
+#[macro_export]
+macro_rules! transformed_module_path {
+    () => {
+        $crate::__string_literal_replace!(module_path!() ("zksync_os_" -> "zk::"))
```
perekopskiy, line 510: IMO the complexity is not worth it; also it will be confusing for new devs and they can have issues with finding logs

perekopskiy (commented):
Looks good, other than prexix replace part

node/bin/src/config.rs
```diff
@@ -368,6 +371,24 @@ pub struct FakeSnarkProversConfig {
     pub max_batch_age: Duration,
 }
 
+/// Configuration for the logging stack.
+#[derive(Debug, Clone, PartialEq, DescribeConfig, DeserializeConfig)]
+#[config(derive(Default))]
+pub struct LogConfig {
+    /// Format of the logs emitted by the node.
+    #[config(default, fallback = &fallback::Env("MISC_LOG_FORMAT"))]
```
RomanBrodetski, line 379: Where does `MISC_LOG_FORMAT` come from?  Why do we need a fallback here?

RomanBrodetski (dismissed):

PR #409, opened 2025-09-24
feat!: add `execution_version` 2

perekopskiy:
- adds `execution_version` 2 which uses new zksync-os dependency
- adds hardcoded vk to mutlivm; vk is now also stored in real proof structs
- `SnarkJobManager::pick_real_job` groups fri proofs only with the same vk

### DIFF
--- a/lib/multivm/src/apps/mod.rs
+++ b/lib/multivm/src/apps/mod.rs
@@ -1,4 +1,5 @@
 pub mod v1 {
+    use alloy::primitives::{B256, b256};
     use std::path::{Path, PathBuf};
     use std::sync::OnceLock;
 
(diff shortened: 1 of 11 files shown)


lib/multivm/src/apps/mod.rs
```diff
@@ -58,4 +59,73 @@ pub mod v1 {
         })
         .clone()
     }
+
+    pub const VERIFICATION_KEY: B256 =
+        b256!("0x0a1b2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8f9"); // TODO!
```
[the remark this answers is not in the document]
perekopskiy: I cannot use b256!(include_string!(..)), as b256 macro expects string. I means I won't be able to use `const` here and I'd prefer to keep it rather than move this string to a file

node/bin/src/prover_input_generator/mod.rs
```diff
@@ -158,42 +159,44 @@ fn compute_prover_input(
...
-                list_source,
-            )
-            .expect("proof gen failed")
-        }
-        v => panic!("Unsupported ZKsync OS execution version: {v}"),
-    };
+    let prover_input =
+        match proving_run_execution_version(replay_record.block_context.execution_version) {
+            1 => unreachable!("proving_run_execution_version does not return 1"), // we prove v1 blocks with v2, it's reflected in `proving_run_execution_version`
+            2 => {
```
AntonD3, line 165: Just to understand better, here we usually would want to support at most 2 versions, right?

lib/multivm/src/lib.rs
```diff
@@ -54,4 +75,19 @@ pub fn simulate_tx<S: ReadStorage, PS: PreimageSource>(
     }
 }
 
-pub const LATEST_EXECUTION_VERSION: u32 = 1;
+pub const LATEST_EXECUTION_VERSION: u32 = 2;
+
+pub fn proving_run_execution_version(forward_run_execution_version: u32) -> u32 {
```
itegulov, line 80: We are bringing proving-related stuff into multivm crate - this is a worrying sign

itegulov: Please see module description in `lib.rs`. This is a module providing unified interface for running blocks and simulating transactions. From my POV it was not meant to understand the nuances of proving. Just to be clear VKs and app.bins should only be stored/exposed for the last two versions. What we are doing right now seems to be that multivm holds both all historical forward systems AND also VKs/app.bins. This is a misdesign IMHO

perekopskiy: I removed apps::v1 and VKs from multivm. So now we only have `proving_run_execution_version` and `apps::v2` as a prover related code in multivm

itegulov: Thanks! FWIW I do think it's cleaner now

itegulov (approved):

PR #414, opened 2025-09-25
fix: backward compatible deserialization for proofs

perekopskiy:


lib/l1_sender/src/batcher_model.rs
```diff
@@ -100,10 +100,15 @@ pub enum FriProof {
...
-pub struct RealFriProof {
-    pub proof: Vec<u8>,
-    pub proving_execution_version: u32,
+#[serde(untagged)]
+pub enum RealFriProof {
+    V1(Vec<u8>),
+    V2 {
+        proof: Vec<u8>,
+        proving_execution_version: u32,
+    },
```
itegulov, line 111: Did you test this? Would make sense to add a unit test here that old format still works.

itegulov (approved):

PR #429, opened 2025-09-29
feat: update zksync-os to v0.0.26 and interface to v0.0.7

itegulov:
This PR:
* Adapts call tracer implementation to work on top of the new interface tracing API (introduced in https://github.com/matter-labs/zksync-os-interface/pull/20 and https://github.com/matter-labs/zksync-os/pull/309).
* Adapts multivm to pulling app.bins from Github releases in compile time instead of the crate (app.bins are no longer checked in on zksync-os side)
* Removes v0.0.23 multivm version as it doesn't have tracing capabilities and there were no forward system related changes since then
* Adapts to other changes in zksync-os-interface v0.0.7 (`gas_per_pubdata` was replaced with `pubdata_price`) which required a new replay wire format V3


lib/multivm/src/lib.rs
```diff
@@ -44,30 +38,33 @@ pub fn run_block<S: ReadStorage, PS: PreimageSource, TS: TxSource, TR: TxResultC
...
+                    tracer,
                 )
                 .map_err(|err| anyhow::anyhow!(err))
         }
         v => panic!("Unsupported ZKsync OS execution version: {v}"),
     }
 }
 
-pub fn simulate_tx<S: ReadStorage, PS: PreimageSource>(
+pub fn simulate_tx<Storage: ReadStorage, PreimgSrc: PreimageSource, Tracer: EvmTracer>(
```
0xVolosnikov, line 49: Seems like we forgot to replace `EvmTracer` here in the interface

lib/types/src/transaction/encode.rs
```diff
@@ -126,7 +120,7 @@ impl From<L2Envelope> for TransactionData {
             from: Address::ZERO,
             to: l2_tx.to().unwrap_or_default(),
             gas_limit: U256::from(l2_tx.gas_limit()),
-            pubdata_price_limit: U256::from(gas_per_pubdata_limit),
+            pubdata_price_limit: U256::from(0),
```
antoniolocascio, line 123: This change is strange. The TransactionData already had a field called "pubdata_price_limit"? How come?

itegulov: We used to pass 50k here but one of the recently added changes asserts that `gas_per_pubdata` is zero for all L2 transactions. The field name itself is likely inaccurate if that's what you mean, but we need to migrate to using zksync_os_api for encoding regardless.

perekopskiy (approved):

PR #448, opened 2025-10-05
feat: pipeline framework (PR 2/X) - `pipe()` syntax; consume `self`; migrate batcher

RomanBrodetski:
* migrates Batcher to the pipeline framework
* simplifies the framework taking @itegulov 's suggestions into account

### DIFF
--- a/lib/pipeline/src/builder.rs
+++ b/lib/pipeline/src/builder.rs
@@ -1,85 +1,88 @@
 use crate::peekable_receiver::PeekableReceiver;
-use crate::{PipelineComponent, Sink, Source};
+use crate::{PipelineComponent, Source};
 use anyhow::Result;
-use std::mem::size_of;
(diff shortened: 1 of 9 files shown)


lib/pipeline/src/builder.rs
```diff
@@ -1,85 +1,88 @@
...
+/// ```ignore
+/// Pipeline::new()
+///     .pipe(CommandSource { ... })  // Source component
+///     .pipe(Sequencer { ... })      // Transformer component
+///     .pipe(TreeManager { ... })    // Transformer component
+///     .execute()
+///     .await
+/// ```
+pub struct Pipeline {
     tasks: Vec<Box<dyn FnOnce() -> tokio::task::JoinHandle<Result<()>> + Send>>,
```
RomanBrodetski, line 18: changing to JoinSet is on the list

PR #458, opened 2025-10-06
feat: Peek batch data from State

AlexOnTheStorm:
Allowing peeking batch data (ProverInput) by batch_number.

### DIFF
--- a/node/bin/src/prover_api/prover_server.rs
+++ b/node/bin/src/prover_api/prover_server.rs
@@ -19,6 +19,12 @@ use zksync_os_l1_sender::batcher_model::FriProof;
 // ───────────── JSON payloads ─────────────
 
 #[derive(Debug, Serialize, Deserialize)]
+struct BatchDataPayload {
+    block_number: u64,
+    prover_input: String, // base64‑encoded little‑endian u32 array
+}
+
+#[derive(Debug, Serialize, Deserialize)]
 struct FriProofPayload {
     block_number: u64,
     proof: String,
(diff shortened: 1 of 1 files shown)


node/bin/src/prover_api/fri_job_manager.rs
```diff
@@ -88,6 +88,20 @@ impl FriJobManager {
...
 
+    /// Peek a batch data for a given batch number
+    pub fn peek_batch_data(&self, batch_number: u64) -> Option<ProverInput> {
+        match self.assigned_jobs.get_batch_data(batch_number) {
+            Some(prover_input) => {
+                tracing::info!("Batch data is peeked for batch number {batch_number}");
+                Some(prover_input)
+            }
+            None => {
+                tracing::debug!("No batch data found for batch number {batch_number}");
```
RomanBrodetski, line 99: this log will be confusing - it will just say "No batch data found for batch number {batch_number}" but there will be no context on why. If you want to log it, do "Trying to peel batch number {batch_number}  that is not present in `assigned_jobs`"

node/bin/src/prover_api/prover_server.rs
```diff
@@ -50,6 +44,12 @@ struct AvailableProofsPayload {
     available_proofs: Vec<String>,
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+struct BatchDataPayload {
+    block_number: u64,
+    prover_input: String,
+}
```
RomanBrodetski, line 51: nit: there was this comment ` // base64‑encoded little‑endian u32 array` on `prover_input` that is gone now. Not a big deal, but worth retaining

RomanBrodetski (approved):

PR #459, opened 2025-10-06
2FA

cytadela8:


node/bin/src/batch_verification_manager.rs
```diff
@@ -0,0 +1,265 @@
...
+                            );
+
+                            tokio::time::sleep(self.config.retry_delay).await;
+                        } else {
+                            tracing::warn!(
+                                "Batch verification failed after {} retries exceeding total timeout. Bailing. Last error: {}",
+                                retry_count,
+                                err
+                            );
+                            break Err(err);
```
cytadela8, line 178: I think this is inline with other cases when we are currently crashing the server, but maybe crashes should be limitted to truly unrecoverable errors. Here it could make sense to keep trying infinitely to verify the batch and wait for an engineer to debug the problem.

cytadela8 (commented):
A lot of tcp server/client code feels very verbose, but I'm not sure how best I could simplify this.

lib/batch_verification/src/signature.rs
```diff
@@ -0,0 +1,4 @@
+// TODO verification, generation, etc..
+pub type Signature = Vec<u8>;
+pub type Signatures = Vec<Signature>;
+
```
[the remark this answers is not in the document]
itegulov: I'd try to use [blst](https://crates.io/crates/blst) for signature primitives here

itegulov (commented):
Please add PR description and a link to ticket

node/bin/src/lib.rs
```diff
@@ -705,9 +709,19 @@ async fn run_en_pipeline<
...
+    let pipeline = if config.batch_verification_config.enabled {
+        pipeline.pipe(BatchVerificationClient::new(
+            config.batch_verification_config.signing_key.clone(),
+            config.genesis_config.chain_id.unwrap(), // TODO check unwrap
+            node_state_on_startup.l1_state.diamond_proxy,
+            config.batch_verification_config.address,
+        ))
+    } else {
+        pipeline.pipe(NoOpSink::new())
+    };
```
itegulov: This can be simplified if you pipe `Option<BatchVerificationClient>` before `NoOpSink`. See my comment [here](https://github.com/matter-labs/zksync-os-server/pull/525#discussion_r2438021671)

lib/merkle_tree/src/with_version.rs
```diff
@@ -12,6 +12,11 @@ pub struct MerkleTreeVersion<DB: Database = RocksDBWrapper, P: TreeParams = Defa
     pub block: u64,
 }
 
+pub struct BlockMerkleTreeData {
+    pub block_start: MerkleTreeVersion,
+    pub block_end: MerkleTreeVersion,
+}
```
itegulov: This does not belong to `merkle_tree` either. It's way more high-level than what `merkle_tree` operates on.

lib/types/src/batch_signature.rs
```diff
@@ -0,0 +1,63 @@
+use alloy::primitives::{Address, Signature as AlloySignature, SignatureError};
+use alloy::signers::Signer;
+use alloy::signers::local::PrivateKeySigner;
+use alloy::sol_types::SolValue;
+use serde::{Deserialize, Serialize};
+use zksync_os_contract_interface::IExecutor::CommitBatchInfoZKsyncOS;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct BatchSignatureSet(Vec<BatchSignature>);
```
itegulov: Similarly, `types` crate should only contains basic types and this is a lot more domain-specific.

I propose you move this and maybe some other batcher-related types into a separate crate (can be done separately ofc). And name it "batcher-types" or something like that to signify that these are types shared by the entire batcher subsystem.

There is an argument for moving `BlockMerkleTreeData` there too

lib/batch_verification/src/verification_client.rs
```diff
@@ -0,0 +1,306 @@
...
+            self.chain_id,
+            self.diamond_proxy,
+            request.batch_number,
+        )
+        .into_l1_commit_data(BatchDaInputMode::Rollup);
+
+        if commit_batch_info != request.commit_data {
+            let mut mismatches = Vec::new();
+
+            // I don't like this, but it works great :/
```
itegulov: Are you doing this because you want a pretty-printed diff between these two structs?

FWIW there is [structdiff](https://github.com/knickish/structdiff) that I have used once in the past. You should be able to derive it for `CommitBatchInfo` pretty easily. Another reason to use it over `CommitBatchInfoZKsyncOS` from alloy_sol :wink: 

@popzxc maybe you have a better suggestion?

lib/batch_verification/src/config.rs
```diff
@@ -0,0 +1,34 @@
+use std::time::Duration;
+
+use smart_config::{DescribeConfig, DeserializeConfig, config, value::SecretString};
+
+#[derive(Clone, Debug, DescribeConfig, DeserializeConfig)]
+#[config(derive(Default))]
```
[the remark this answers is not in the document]
cytadela8: Honestly I don't understand why we do it like that, but I will convert.

lib/batch_verification/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+pub use verification_response::BatchVerificationResponse;
+pub use verification_response::BatchVerificationResponseCodec;
+pub use verification_response::BatchVerificationResponseDecoder;
+pub use verification_response::BatchVerificationResult;
+
+mod verification_server;
+pub use verification_server::BatchVerificationRequestError;
+pub use verification_server::BatchVerificationServer;
+
+mod verification_client;
```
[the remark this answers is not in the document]
cytadela8: Reworking naming

lib/batch_verification/src/verification_client.rs
```diff
@@ -0,0 +1,427 @@
...
+                    ?dur,
+                    "retrying connection to main node for batch verification"
+                );
+            })
+            .await?;
+
+        // This makes it valid HTTP
+        socket
+            .write_all(b"POST /batch_verification HTTP/1.0\r\n\r\n")
+            .await?;
```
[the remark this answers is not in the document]
cytadela8: I'm moving some logic into a generic listener component. However retires need to be separately implemented for both replay and batch_verification (much more complex for replay) so I want to have this out of scope.

lib/storage_api/src/lib.rs
```diff
@@ -24,3 +24,6 @@ pub use metered_state::{MeteredViewState, StateAccessLabel};
 
 mod state;
 pub use state::{ReadStateHistory, StateError, StateResult, ViewState, WriteState};
+
+mod util;
+pub use util::skip_http_headers;
```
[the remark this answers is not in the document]
cytadela8: Removed

node/bin/src/lib.rs
cytadela8: Hi, looked into this. So I would need to add that implementation for Option<C>, but to have a sensible implementation for NAME and OUTPUT_BUFFER_SIZE I would need to change those from const to functions. Can we push this to a separate PR, becouse its going to touch a lot of files?

lib/batch_verification/src/server_verification_component.rs
```diff
@@ -0,0 +1,265 @@
...
+            match tokio::time::timeout(remaining_time, response_receiver.recv()).await {
+                Ok(Some(response)) => {
+                    // TODO add validation of signatures incl. uniqueness
+                    responses.push(response.signature);
+                    tracing::debug!(
+                        "Validated response for batch {} (request {}), {} of {}",
+                        batch_envelope.batch_number(),
+                        request_id,
+                        responses.len(),
+                        self.config.threshold
```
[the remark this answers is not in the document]
cytadela8: Modified it to use this kind of syntax I think everywhere in my new code

lib/batch_verification/src/client.rs
```diff
@@ -0,0 +1,336 @@
...
+use zksync_os_storage_api::ReplayRecord;
+use zksync_os_types::BatchSignature;
+
+/// Cache of blocks that are to be used for batch verification
+/// Accepts blocks only in ascending order. Old blocks are evicted when not
+/// needed through a call to remove_lower_then.
+///
+/// This may be optimized by using a ring buffer for data storage instead.
+///
+/// TODO metric for blocks in signing cache
```
cytadela8: Finish this TODO

lib/batch_verification/src/client.rs
```diff
@@ -0,0 +1,336 @@
...
+                            match verification_result {
+                                Ok(signature) => {
+                                    tracing::info!(batch_number, request_id, "Approved batch verification request");
+                                    writer.send(BatchVerificationResponse { request_id, result: BatchVerificationResult::Success(signature) }).await?;
+                                },
+                                Err(reason) => {
+                                    tracing::info!(batch_number, request_id, "Batch verification failed: {}", reason);
+                                    writer.send(BatchVerificationResponse { request_id, result: BatchVerificationResult::Refused(reason.to_string()) }).await?;
+                                },
+                            }
```
RomanBrodetski: very nice high-level flow - concise and easy to follow

node/bin/src/lib.rs
RomanBrodetski: yes, it can be improved but it's ok to do separately

lib/batch_verification/src/server.rs
```diff
@@ -0,0 +1,168 @@
...
+    ) -> anyhow::Result<()> {
+        let (recv, mut send) = socket.split();
+        let mut reader = BufReader::new(recv);
+
+        // Skip HTTP headers similar to replay_transport
+        skip_http_headers(&mut reader).await?;
+
+        // Write wire format version
+        send.write_u32(BATCH_VERIFICATION_WIRE_FORMAT_VERSION)
+            .await?;
```
popzxc: I am very concerned that the protocol implementation is 
1. Not encapsulated (we mix low-level protocol interchange with the component business logic).
2. Not documented (or at least I didn't find any docs).
3. Doesn't seem to be extendable (I worry about future compatibility somewhat).

If we want to take a look at Ethereum, it already has an established communication protocol: [rlpx](https://github.com/ethereum/devp2p/blob/master/rlpx.md). Moreover, it's extendable: you can add new capabilities and implement sub-protocols easily for different purposes.
If we take a look at engineering globally, we have protobuf/capnp/etc, which would help with a lot of stuff we currently have to worry about.

Was it a deliberate/discussed decision to implement a new protocol here?

lib/batch_verification/src/sequencer_component.rs
```diff
@@ -0,0 +1,372 @@
...
+        output: mpsc::Sender<Self::Output>,
+    ) -> anyhow::Result<()> {
+        if self.config.enabled {
+            let (server, response_receiver) = BatchVerificationServer::new();
+            let server = Arc::new(server);
+            let response_channels = Arc::new(DashMap::new());
+
+            let server_for_fut = server.clone();
+            let server_address = self.config.address.clone();
+            let server_fut = async move { server_for_fut.run_server(server_address).await }
```
popzxc: Can't we just have `server_for_fut.run_server(server_address).boxed()`? Why do we need to have `async move { <x>.await }`? Here and below

lib/batch_verification/src/server.rs
```diff
@@ -0,0 +1,168 @@
...
+                )
+                .await
+                {
+                    tracing::info!("Error handling client {}: {}", addr, e);
+                }
+            });
+        }
+    }
+
+    async fn handle_client(
```
popzxc: 1. Shouldn't we have some kind of handshake to ensure that both parties use the same version of the protocol? Right now, it looks like if client doesn't support required version of the protocol, the only option it has is to close the connection (which is not very helpful).
2. Don't we want the communication to be encrypted? E.g. during the handshake we verify that the person who connects is among `accepted_signers`, and then derive a shared ephemeral secret to encode the communication (similarly to what rlpx or noise protocol does). Otherwise, it looks like we're very prone to any kind of MITM attack where the messages in either direction can be altered (causing valid blocks to be not signed or returning malformed signatures to the server).

lib/storage_api/src/util.rs
popzxc: ?

node/bin/src/config.rs
```diff
@@ -432,6 +434,39 @@ pub struct LogConfig {
...
+    /// [server] Batch verification server address to listen on.
+    /// [en] Batch verification server address to connect to.
+    #[config(default_t = "0.0.0.0:3072".into())]
+    pub address: String,
+    /// [server] Threshold (number of needed signatures)
+    #[config(default_t = 1)]
+    pub threshold: usize,
+    /// [server] Accepted signer pubkeys
+    #[config(default_t = vec!["0x36615Cf349d7F6344891B1e7CA7C72883F5dc049".into()])]
+    pub accepted_signers: Vec<String>,
```
popzxc: Do we want to make it a part of the local configuration? I assume that later it will be taken from some other source, e.g. L1, but if we introduce it as a part of overall node configuration, it might be hard to remove later.

Probably question for @RomanBrodetski -- do we want any kind of notion for experimental/temporary configs that do not fall under the breaking changes rules?

run_en_batch_verification.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
itegulov: Also not a big fan of bash scripts. Can you add a section in the docs instead?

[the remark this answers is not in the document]
cytadela8: Sure, but I'm a fan of bash scripts becouse when I have docs I need to waste time writing bash scripts 🙃

Cargo.toml
```diff
@@ -27,6 +27,10 @@ members = [
     "lib/multivm",
     "lib/pipeline",
     "lib/gas_adjuster",
+    "lib/batch_verification",
+    "lib/socket",
+    "lib/batch_types",
+    "lib/batch_verification",
```
cytadela8: listed twice

PR #477, opened 2025-10-08
feat: pipeline framework PR 6/X - migrate l1 sender proves and SnarkJobsManager

RomanBrodetski:
Only PriorityTree and l1 executor are left after this PR.

We already benefit from the Pipeline - the rescheduleing logic is much cleaner. After other components are migrated, I'll do a nice cleanup of `lib.rs`.

I disabled running with no L1 (not used anywhere at the moment) - will reimplement it via pipleine later on


lib/pipeline/src/builder.rs
```diff
@@ -48,6 +48,27 @@ impl<Output: Send + 'static> Pipeline<Output> {
         }
     }
 
+    /// Add a transformer component to the pipeline with prepended messages
+    ///
+    /// This is useful when you need to reschedule messages at the start of the pipeline.
+    /// The prepended messages are sent to the component before any messages from the pipeline.
+    pub fn pipe_with_prepend<C>(mut self, component: C, prepend: Vec<Output>) -> Pipeline<C::Output>
```
itegulov, line 55: As a thought: this is representable even with just regular `pipe`. You just need a component that outputs the desired prefix first and then continues to output its unchanged input. This might be a good opportunity to move the rescheduling logic from `bin/lib.rs` into `CommitRescheduler`, but leaving the decision up to you

lib/l1_sender/src/pipeline_component.rs
```diff
@@ -27,7 +27,7 @@ where
     type Output = BatchEnvelope<FriProof>;
 
     const NAME: &'static str = C::NAME;
-    const OUTPUT_BUFFER_SIZE: usize = 5;
+    const OUTPUT_BUFFER_SIZE: usize = 1;
```
itegulov, line 30: Why `1`?

itegulov (approved):

lib/pipeline/src/builder.rs
RomanBrodetski: yeah good point.. note that we apply this logic in two places within pipline: 
* L1 committed but not L1 proven
* L1 proven but not L1 executed

so the suggestion would be to delegate this to the respective "previous" components? So, we'd create a CommitRescheduler and ProveRescheduler... And they may be generic just like we did for l1_sender.

I like this. I will finish the pipeline first and revisit.

lib/l1_sender/src/pipeline_component.rs
RomanBrodetski: I will revisit all buffer sizes to be more delibarate. Just putting 5 everywhere is not a good approach - e.g. gapless committer itself shouldn't have any input buffer (it has a buffer inside). 

FYI this one I set to `1` now because I added a buffer in SnarkProverPipelineStep (there is a comment there).

PR #501, opened 2025-10-10
ci: add server release binaries publication

antonbaliasnikov:
## What?

* [x] Add server binaries publication for GitHub releases.
* [x] Rename execution binary from `zksync_os_bin` to `zksync-os-server` 

After each release PR is merged, binaries will be built and added to the corresponding release.

Each binary contains [GitHub Attestation ](https://docs.github.com/en/actions/concepts/security/artifact-attestations) and is packed inside a compressed tarball.

macOS binaries are unified into one universal binary for x86_64 and aarch64 architectures.

Before publication, binaries are smoke tested on all supported versions of Ubuntu and macOS.

## Why?

To make binaries available for our users.

`genesis.json` and `l1-state.json` are also attached to the release artifacts to make sure the quick local setup for each particular release is possible.

## Tests

Updated binaries for the latest `v0.8.2` tag:
https://github.com/matter-labs/zksync-os-server/releases/tag/v0.8.2

image
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,7 +158,7 @@ jobs:
   #  Build prover integration tests  #
   ####################################
   build-prover-tests:
-    runs-on: matterlabs-ci-runner-ultra-performance
+    runs-on: matterlabs-ci-runner-high-performance
     container:
       image: nvidia/cuda:12.9.1-devel-ubuntu22.04
     steps:


itegulov (dismissed):

.github/workflows/ci.yml
```diff
@@ -158,7 +158,7 @@ jobs:
   #  Build prover integration tests  #
   ####################################
   build-prover-tests:
-    runs-on: matterlabs-ci-runner-high-performance
+    runs-on: matterlabs-ci-runner-ultra-performance
```
yorik: Why are you changing to ultra runner?

yorik (approved):

PR #525, opened 2025-10-14
feat: revm consistency checker

vladbochok:
Adding a separate pipeline item for running the revm execution environment to compare state results. The zksync-revm uses a Database with access to a custom state handler. All state changes across one block are accumulated in an in-memory (cache) database, and for each new block, the data is loaded from the server state handler. This means that each block comparison is done separately, and if a mismatch happens in block X, the block X + 1 comparison won’t be affected.

You can find the zksync-revm here:
- https://github.com/matter-labs/zksync-os-revm/pull/2

### DIFF
--- a/lib/revm_consistency_checker/src/lib.rs
+++ b/lib/revm_consistency_checker/src/lib.rs
@@ -1,5 +1,5 @@
 pub mod bytecode_hash;
 pub mod helpers;
 pub mod node;
-pub mod revm_state_db;
+pub mod revm_state_provider;
 pub mod storage_diff_comp;
--- a/lib/revm_consistency_checker/src/node.rs
+++ b/lib/revm_consistency_checker/src/node.rs
@@ -12,14 +12,14 @@ use zksync_os_revm::{DefaultZk, ZkBuilder};
 use zksync_os_storage_api::{ReadStateHistory, ReplayRecord};
 
 use crate::helpers::zk_tx_into_revm_tx;
-use crate::revm_state_db::RevmStateDb;
+use crate::revm_state_provider::RevmStateProvider;
 use crate::storage_diff_comp::CompareReport;
 
 pub struct RevmConsistencyChecker<State>
 where
     State: ReadStateHistory + Clone + Send + 'static,
 {
-    state: RevmStateDb<State>,
+    state: State,
 }
 
 impl<State> RevmConsistencyChecker<State>
@@ -27,9 +27,7 @@ where
     State: ReadStateHistory + Clone + Send + 'static,
 {
     pub fn new(state: State) -> Self {
-        Self {
-            state: RevmStateDb::new(state, 0, Default::default()),
-        }
+        Self { state }
     }
 }
 
@@ -60,14 +58,18 @@ where
             };
 
             latency_tracker.enter_state(GenericComponentState::Processing);
-            self.state.set_latest_block(
-                replay_record.block_context.block_number - 1,
-                replay_record.block_context.block_hashes,
-            );
+            let state_block_number = replay_record.block_context.block_number - 1;
+            let block_hashes = replay_record.block_context.block_hashes;
+            let state_view = self
+                .state
+                .state_view_at(state_block_number)
+                .map_err(anyhow::Error::from)?;
 
             {
                 // For each block, we create an in-memory cache database to accumulate transaction state changes separately
-                let mut cache_db = CacheDB::new(self.state.clone());
+                let state_provider =
+                    RevmStateProvider::new(state_view, block_hashes, state_block_number);
+                let mut cache_db = CacheDB::new(state_provider);
                 let mut evm = Context::default()
                     .with_db(&mut cache_db)
                     .modify_cfg_chained(|cfg| {
--- a/lib/revm_consistency_checker/src/revm_state_provider.rs
+++ b/lib/revm_consistency_checker/src/revm_state_provider.rs
@@ -8,63 +8,50 @@ use reth_revm::{
 };
 use ruint::aliases::B160;
 use zk_ee::common_structs::derive_flat_storage_key;
-use zk_os_forward_system::run::ReadStorage;
-use zksync_os_interface::{traits::PreimageSource, types::BlockHashes};
+use zksync_os_interface::types::BlockHashes;
 use zksync_os_merkle_tree::fixed_bytes_to_bytes32;
-use zksync_os_storage_api::ReadStateHistory;
-use zksync_os_storage_api::{StateError, ViewState};
+use zksync_os_storage_api::ViewState;
 
 #[derive(Debug, Clone)]
-pub struct RevmStateDb<State>
+pub struct RevmStateProvider<State>
 where
-    State: ReadStateHistory + Clone + Send + 'static,
+    State: ViewState,
 {
-    state: State,
-    latest_block: u64,
+    state_view: State,
     block_hashes: BlockHashes,
+    state_block_number: u64,
 }
 
-impl<State> RevmStateDb<State>
+impl<State> RevmStateProvider<State>
 where
-    State: ReadStateHistory + Clone + Send + 'static,
+    State: ViewState,
 {
-    pub fn new(state: State, latest_block: u64, block_hashes: BlockHashes) -> Self {
-        RevmStateDb {
-            state,
-            latest_block,
+    pub fn new(state_view: State, block_hashes: BlockHashes, state_block_number: u64) -> Self {
+        Self {
+            state_view,
             block_hashes,
+            state_block_number,
         }
-    }
-
-    pub fn set_latest_block(&mut self, latest_block: u64, block_hashes: BlockHashes) {
-        self.latest_block = latest_block;
-        self.block_hashes = block_hashes;
     }
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -141,6 +141,8 @@ reth-revm = { git = "https://github.com/paradigmxyz/reth", tag = "v1.7.0" }
 reth-storage-api = { git = "https://github.com/paradigmxyz/reth", tag = "v1.7.0" }
 reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", tag = "v1.7.0" }
 reth-execution-types = { git = "https://github.com/paradigmxyz/reth", tag = "v1.7.0" }
+zksync-revm = { git = "https://github.com/vladbochok/zksync-revm" }
+revm-inspectors = { version = "0.30" }
```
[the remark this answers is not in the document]
itegulov: Ah, I suppose it might require newer revm and hence newer reth? Feel free to ignore if so

lib/pipeline/src/builder.rs
```diff
@@ -65,6 +65,17 @@ impl<Output: Send + 'static> Pipeline<Output> {
...
+    /// Conditionally add a component if present. The component must keep the same item type.
+    pub fn pipe_opt<C>(self, component: Option<C>) -> Pipeline<Output>
+    where
+        C: PipelineComponent<Input = Output, Output = Output>,
+    {
+        match component {
+            Some(c) => self.pipe(c),
+            None => self,
+        }
+    }
```
itegulov: I think it would be cleaner if you wrote a blanket implementation for `Option<C>`. Something like this:
```rust
impl<T: Send + Sync + 'static, C: PipelineComponent<Input = T, Output = T>> PipelineComponent
    for Option<C> {
  // todo: implement
}
```

node/bin/src/revm_consistency_checker/bytecode_hash.rs
```diff
@@ -0,0 +1,87 @@
...
+    );
+}
+
+#[test]
+fn calculate_empty_bytecode_test() {
+    use alloy::hex;
+    let bytecode = Bytecode::new_legacy(hex!("").into());
+    let bytecode_hash = calculate_bytecode_hash(&bytecode);
+    assert_eq!(bytecode_hash, EMPTY_BYTE_CODE_HASH);
+}
```
itegulov: Please put tests under `#[cfg(test)] mod tests {}`. Otherwise, AFAIU, `#[test]` would still get expanded and end up in release binary.

Cargo.toml
```diff
@@ -141,6 +141,8 @@ reth-revm = { git = "https://github.com/paradigmxyz/reth", tag = "v1.7.0" }
 reth-storage-api = { git = "https://github.com/paradigmxyz/reth", tag = "v1.7.0" }
 reth-transaction-pool = { git = "https://github.com/paradigmxyz/reth", tag = "v1.7.0" }
 reth-execution-types = { git = "https://github.com/paradigmxyz/reth", tag = "v1.7.0" }
+zksync-revm = { git = "https://github.com/vladbochok/zksync-revm" }
```
itegulov: Please pin commit or even better a tag

node/bin/src/revm_consistency_checker/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod bytecode_hash;
+pub mod helpers;
+pub mod node;
+pub mod revm_state_db;
+pub mod storage_diff_comp;
```
itegulov: I'd move this entire module into its own crate. From what I have seen there is no reason for it to be in `bin`

node/bin/src/revm_consistency_checker/node.rs
```diff
@@ -0,0 +1,119 @@
...
+                .build_zk();
+
+            let revm_txs = replay_record
+                .transactions
+                .iter()
+                .zip(block_output.tx_results)
+                .filter_map(|(transaction, tx_output_raw)| {
+                    let tx_output = match tx_output_raw {
+                        Ok(tx_output) => tx_output,
+                        _ => return None, // Skip invalid transaction as they are not included in the batch
```
itegulov: I wish zksync-os did not return `None` for invalid transactions at all but we never addressed it :pensive: Nothing actionable here, just noting

node/bin/src/revm_consistency_checker/revm_state_db.rs
```diff
@@ -0,0 +1,146 @@
...
+use zksync_os_storage_api::ReadStateHistory;
+use zksync_os_storage_api::{StateError, ViewState};
+
+#[derive(Debug, Clone)]
+pub struct RevmStateDb<State>
+where
+    State: ReadStateHistory + Clone + Send + 'static,
+{
+    state: State,
+    latest_block: u64,
```
itegulov: It feels like this should hold `ViewState` instead. Then `RevmConsistencyChecker` would initialize a new instance in every loop iteration instead of modifying an existing handle via `set_latest_block`.

itegulov (commented):
Only did a surface-level review (and did not look into zksync-revm at all) but overall aligns with what I expected to see, great work!

node/bin/src/revm_consistency_checker/node.rs
```diff
@@ -0,0 +1,119 @@
...
+                    ))
+                });
+
+            evm.transact_many_commit(revm_txs)?;
+            let compare_report = CompareReport::build(
+                evm.0.db_mut(),
+                &block_output.storage_writes,
+                &block_output.account_diffs,
+            )?;
+            compare_report.log_tracing(20);
```
mm-zk: btw - I was expecting some explicit -- "compare_report.is_valid()" + maybe an exported metric (for alerting etc).

node/bin/src/revm_consistency_checker/revm_state_db.rs
```diff
@@ -0,0 +1,146 @@
...
+            .state_view_at(self.latest_block)?
+            .read(flat_key)
+            .unwrap_or_default()
+            .into_u256_be())
+    }
+
+    /// Gets block hash by block number.
+    fn block_hash_ref(&self, number: u64) -> Result<B256, Self::Error> {
+        if let Some(diff) = self.latest_block.checked_sub(number)
+            && diff < 256
```
mm-zk: shouldn't we return an error if diff > 256 ?

lib/revm_consistency_checker/src/bytecode_hash.rs
```diff
@@ -0,0 +1,95 @@
+use alloy::primitives::B256;
```
RomanBrodetski: are unsafes necesarry in this file? This is the first place in the codebase where we introduce it. Is there a way around it? what's the trade-off?

lib/revm_consistency_checker/src/revm_state_db.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn set_latest_block(&mut self, latest_block: u64, block_hashes: BlockHashes) {
+        self.latest_block = latest_block;
+        self.block_hashes = block_hashes;
+    }
+}
+
+#[derive(Debug)]
+pub struct RevmStateDbError(anyhow::Error);
+
+// add this:
```
RomanBrodetski: leftover?

lib/revm_consistency_checker/src/revm_state_db.rs
```diff
@@ -0,0 +1,146 @@
...
+use zksync_os_storage_api::ReadStateHistory;
+use zksync_os_storage_api::{StateError, ViewState};
+
+#[derive(Debug, Clone)]
+pub struct RevmStateDb<State>
+where
+    State: ReadStateHistory + Clone + Send + 'static,
+{
+    state: State,
+    latest_block: u64,
```
RomanBrodetski: Instead of building one instance of `RevmStateDb` and mutating it on every block via `set_latest_block`, I recommend building an immutable `RevmStateProvider` every block. This way you ensure there are no leftovers (dirty state) when you reuse it from block to block (I don't see an issue now, but this will evolve - having mutability here makes it more error-prone).

So, you'll only have a `State` (`ReadStateHistory + Clone + Send + 'static`) as the component dependency (`RevmConsistencyChecker`), and every block you will do 
```
RevmStateProvider::new(state.state_view_at(block_number), block_hashes, block_number)
```
and only hold it while you process the block. 
Alternatively to passing block_number, you can reindex block_hashes so that it is indexed by block numbers and not diffs - both would wrk.

lib/revm_consistency_checker/src/revm_state_db.rs
```diff
@@ -0,0 +1,156 @@
...
+        RevmStateDbError(e.into())
+    }
+}
+
+impl<State> DatabaseRef for RevmStateDb<State>
+where
+    State: ReadStateHistory + Clone + Send + 'static,
+{
+    /// The database error type.
+    type Error = RevmStateDbError;
```
RomanBrodetski: what's the advantage over just    ` type Error = anyhow::Error;`? Alternatively, if you want/need to keep a custom error, you can do 
```
#[derive(Debug, Error)]
#[error(transparent)]
pub struct RevmStateDbError(#[from] anyhow::Error);
``` 
and I think you'll get `Display`, `Error::source` and `From<anyhow::Error>`

PR #532, opened 2025-10-14
fix: Persisting some info about the failed batch

AlexOnTheStorm:
Instead of saving all the batch metadata we now save only `batch_number, last_block_timestamp, expected_hash_u32s and proof_final_register_values,`
This allows us not importing the whole crate to airbender-prover for the sake of one struct BatchMetadata.

### DIFF
--- a/node/bin/src/prover_api/prover_server.rs
+++ b/node/bin/src/prover_api/prover_server.rs
@@ -112,8 +112,7 @@ async fn submit_fri_proof(
         }) => Err((
             StatusCode::BAD_REQUEST,
             format!(
-                "FRI proof verification failed. Expected: {:?}, Got: {:?}",
-                expected_hash_u32s, proof_final_register_values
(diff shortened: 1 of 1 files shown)


node/bin/src/prover_api/fri_job_manager.rs
```diff
@@ -47,13 +46,30 @@ pub enum SubmitError {
...
+        }
+    }
+    pub fn proof_final_register_values(&self) -> [u32; 16] {
+        match self {
+            SubmitError::VerificationFailed(_, proof_final_register_values) => {
+                *proof_final_register_values
+            }
+            _ => panic!("Proof final register values not found"),
+        }
+    }
```
itegulov: I assume you need these because you know that `verify_fri_proof` can only return this variant of `SubmitError`. I would just create a separate error type `FriProofVerificationError` that is wrapped inside `SubmitError`. Then you can return it from `verify_fri_proof` and safely address fields without the need for these methods.

itegulov (approved):

Issue #534, opened 2025-10-15
Make node agnostic of RocksDB


itegulov:
A lot of pieces are already in place we need to make sure all node components rely on traits instead of specific implementations

EmilLuta:
Still in progress, most likely.

PR #535, opened 2025-10-15
feat(storage): move replay DB to storage crate

itegulov:
Part of #534 

This PR moves replay storage into `storage` crate and make all components rely on `ReadReplay/WriteReplay` traits instead of one specific implementation. PR is easily reviewable commit-by-commit if you prefer. But not a lot of changes regardless.

### DIFF
--- a/node/bin/src/prover_api/fri_job_manager.rs
+++ b/node/bin/src/prover_api/fri_job_manager.rs
@@ -27,7 +27,6 @@ use tokio::sync::mpsc::Permit;
 use tokio::sync::mpsc::error::TrySendError;
 use tokio::sync::{Mutex, mpsc};
 use zksync_os_l1_sender::batcher_metrics::BatchExecutionStage;
-use zksync_os_l1_sender::batcher_model::BatchMetadata;
 use zksync_os_l1_sender::batcher_model::{BatchEnvelope, FriProof, ProverInput, RealFriProof};
 use zksync_os_multivm::proving_run_execution_version;
 use zksync_os_observability::{
@@ -39,6 +38,11 @@ use zksync_os_pipeline::PeekableReceiver;
 pub enum SubmitError {
     #[error("proof did not pass verification")]
     VerificationFailed,
+    #[error("FRI proof verification error")]
+    FriProofVerificationError {
+        expected_hash_u32s: [u32; 8],
+        proof_final_register_values: [u32; 16],
+    },
     #[error("batch {0} is not known to the server")]
     UnknownJob(u64),
     #[error("deserialization failed: {0:?}")]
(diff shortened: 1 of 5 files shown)


lib/storage/src/db/replay.rs
```diff
@@ -288,24 +234,34 @@ impl ReadReplay for BlockReplayStorage {
...
-        let current_latest_block = self.latest_block().unwrap_or(0);
-
-        if record.block_context.block_number <= current_latest_block {
+        let current_latest_record = self.latest_record();
+        if record.block_context.block_number <= current_latest_record {
             tracing::debug!(
-                "Not appending block {}: already exists in WAL",
-                record.block_context.block_number
+                block_number = record.block_context.block_number,
+                "not appending block: already exists in WAL",
```
RomanBrodetski: nit: when I wrote this I wasn't consistent with namings.. Sometimes `WAL` sometimes `Block Replay Storage`. I'd use the latter everywhere. Can be tackled in a separete PR (I can do it)

lib/storage_api/src/replay.rs
```diff
@@ -1,20 +1,145 @@
...
 
     /// Get full data needed to replay a block by its number.
+    ///
+    /// This method:
+    /// * MUST be thread-safe
+    /// * MUST return `Some(_)` for block number 0 (genesis)
+    /// * MUST, for a given number, return the same value every time as long as it returned `Some(_)`
+    ///   at least once
+    /// * MUST return `Some(_)` for all block numbers `<N` if it returned `Some(_)` for `N`
+    /// * MUST return `Some(_)` for block number returned by [`latest_record`](Self::latest_record)
```
RomanBrodetski: nit - I'd simplify:

```
  /// * MUST be thread-safe
  /// * MUST return `Some(_)` for all block numbers in range `[0; latest_record()]`
  /// * MUST return the same value for any block number once it returns `Some(_)` at least once
  /// * MAY return `Some(_)` for block numbers after latest
```

lib/storage_api/src/replay.rs
```diff
@@ -1,20 +1,145 @@
...
+            match record {
+                Some(record) => futures::future::ready(Some(record)),
+                None => futures::future::ready(None),
+            }
+        });
+        Box::pin(stream)
+    }
+
+    /// Streams all replay records with block_number ≥ `start`, in ascending block order. On reaching
+    /// the latest stored record continuously waits for new records to appear.
```
RomanBrodetski:
```suggestion
    /// the latest stored record continuously waits for new records to appear. Used to send blocks to ENs
```

RomanBrodetski (dismissed):
All comments are optional/nits.

One more thing: did you consider putting this in a separate create (not `storage`)? How exactly to choose whether a module belongs to `storage`? By the fact that it has some persistence? But by this logic, if we build a new feature, put it to its own crate, and then add persistence, we'd need to move it to `storage`?

Same applies to `storage_api` crate.

lib/storage/src/db/replay.rs
itegulov: I renamed everything apart from column family

RomanBrodetski:
I think it's fundamentally different from DAL because here components are actually, truly free to use and manage their own persistence. In Era we used a relational DB and it wasn't possible to fully define the ownership of tables as for some (most) business logic we had to join tables from differnt components. Hence DAL as a single layer.

So, having persistence of different components in something like `storage` may encourage engineers to introduce tighter coupling between components' persistence layers, which hurts encapsulation. 

> This makes it easy to tell at a glance all things that can/must be persisted on disk

Yeah, useful - just like it'd be useful to see at a glance all the web servers (ie open ports) or all components that depend on e.g. external services - but we don't create one crate that has all this.

What do you guys think?

PR #538, opened 2025-10-15
chore: RocksDB explorer

popzxc:
While I was learning the current state of the codebase, I decided to vibe code a tool to view contents of RocksDB.
It worked pretty well (for my purposes, at least), so I spent a bit of time polishing it.

It's up to you to decide if you want to have it; I'm fine with this PR being closed, since while it _seems_ to be working fine, it's still an LLM product.

This tool allows you to check all the databases we currently have and look for specific fields. There are basic rails for editing capabilities (you can edit entries, but it's implemented only for some CFs; adding more support should be trivial if required). There is also sorting-by-field support, but it's kinda useless, since we don't load the entire database to memory.

Video demo:

https://github.com/user-attachments/assets/25a34a5d-c7a3-4300-a927-1ed7095e8f75

### DIFF
--- a/tools/state-viewer/src/app.rs
+++ b/tools/state-viewer/src/app.rs
@@ -5,11 +5,11 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::{Result, anyhow};
 use crossterm::event::{KeyCode, KeyEvent, KeyModifiers};
 
-use crate::db::{open_components, DbStore};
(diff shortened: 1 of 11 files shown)


antonbaliasnikov:
As this is a tool, I believe PR should have a `chore` commit type to not have any impact on semantic versioning.

popzxc:
Closing, as the team would prefer implementation to be linked to the storage types, but reworking storage abstractions to make it possible (properly) is too much work for now.

Issue #530, opened 2025-10-14
Support state override for `eth_call`


itegulov:
This requires replacing [this](https://github.com/matter-labs/zksync-os-server/blob/main/lib/rpc/src/eth_call_handler.rs#L192-L194) with a real implementation that overrides provided slots in state for the rest of the execution.

I envision this as a new `ViewState` implementation that proxies another `Box<dyn ViewState>` handle inside of it but returns custom values for overridden slots.

PR #539, opened 2025-10-15
feat: eth_call state overrides

dimazhornyk:
Adds state overrides support for `eth_call` #530
Supports both standard storage slot overrides and account overrides (balance/nonce/code)

### DIFF
--- a/lib/rpc/src/eth_call_handler.rs
+++ b/lib/rpc/src/eth_call_handler.rs
@@ -7,12 +7,10 @@ use alloy::consensus::transaction::Recovered;
 use alloy::consensus::{SignableTransaction, TxEip1559, TxEip2930, TxLegacy, TxType};
 use alloy::eips::BlockId;
 use alloy::network::TransactionBuilder;
-use alloy::primitives::{Address, B256, Bytes, Signature, TxKind, U256, ruint::aliases::B160};
+use alloy::primitives::{Address, B256, Bytes, Signature, TxKind, U256};
 use alloy::rpc::types::state::StateOverride;
 use alloy::rpc::types::trace::geth::{CallConfig, GethTrace};
 use alloy::rpc::types::{BlockOverrides, TransactionRequest};
-use std::collections::HashMap;
-use zk_ee::common_structs::derive_flat_storage_key;
 use zk_os_api::helpers::{get_balance, get_nonce};
 use zksync_os_interface::types::ExecutionOutput;
 use zksync_os_interface::{
@@ -21,8 +19,7 @@ use zksync_os_interface::{
 };
 use zksync_os_storage_api::ViewState;
 use zksync_os_storage_api::{
-    RepositoryError, StateError,
-    state_override_view::{AccountViewOverride, OverriddenStateView},
+    RepositoryError, StateError, state_override_view::OverriddenStateView,
 };
 use zksync_os_types::{
     L1_TX_MINIMAL_GAS_LIMIT, L1Envelope, L1PriorityTxType, L1Tx, L1TxType, L2Envelope,
@@ -228,14 +225,11 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
             .state_view_at(execution_env.block_context.block_number)?;
 
         let res = match state_overrides {
-            Some(overrides) => {
-                let (slot_overrides, account) = parse_state_overrides(overrides);
-                execute(
-                    execution_env.transaction,
-                    execution_env.block_context,
-                    OverriddenStateView::new(storage_view, slot_overrides, account),
-                )
-            }
+            Some(overrides) => execute(
+                execution_env.transaction,
+                execution_env.block_context,
+                OverriddenStateView::new(storage_view, overrides),
+            ),
             None => execute(
                 execution_env.transaction,
                 execution_env.block_context,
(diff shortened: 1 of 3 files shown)


integration-tests/tests/call.rs
```diff
@@ -154,3 +148,41 @@ async fn call_revert() -> anyhow::Result<()> {
...
+    let tester = Tester::setup().await?;
+
+    // Deploy TracingSecondary with `data = 1` stored at slot 0
+    let initial_data = U256::from(1);
+    let contract = TracingSecondary::deploy(tester.l2_provider.clone(), initial_data).await?;
+
+    // Build a TransactionRequest for multiply(1) -> returns the storage-backed value
+    let tx_req = contract.multiply(U256::from(1)).into_transaction_request();
+
+    // Baseline call without overrides (should return 7)
```
itegulov: Why 7? Looks like it should be 1

integration-tests/tests/call.rs
```diff
@@ -154,3 +148,41 @@ async fn call_revert() -> anyhow::Result<()> {
...
+    let overrides_json = json!({
+        format!("{:#x}", contract.address()): {
+            "state": {
+                // slot 0 -> 0x...02
+                "0x0000000000000000000000000000000000000000000000000000000000000000":
+                    format!("0x{:064x}", U256::from(2u64)),
+            }
+        }
+    });
+    let overrides: StateOverride = serde_json::from_value(overrides_json)?;
```
itegulov: You can do this type-safely like this:
```rust
let overrides = StateOverride::from_iter([(
    *contract.address(),
    AccountOverride {
        balance: None,
        nonce: None,
        code: None,
        state: Some(HashMap::from_iter([(
            b256!("0x0000000000000000000000000000000000000000000000000000000000000000"),
            b256!("0x0000000000000000000000000000000000000000000000000000000000000002"),
        )])),
        state_diff: None,
        move_precompile_to: None,
    },
)]);
```

lib/rpc/src/eth_call_handler.rs
```diff
@@ -571,3 +620,56 @@ pub enum EthCallError {
...
     State(#[from] StateError),
 }
+
+/// Parses the RPC `StateOverride` structure into separate maps of slot overrides
+/// and account-level overrides (balance/nonce/code).
+fn parse_state_overrides(
+    state_overrides: StateOverride,
+) -> (
+    std::collections::HashMap<B256, B256>,
+    std::collections::HashMap<Address, AccountViewOverride>,
```
itegulov: Nit: I'd just import HashMap, this is a bit noisy

lib/storage_api/src/state_override_view.rs
```diff
@@ -0,0 +1,113 @@
...
+
+            let mut base: AccountProperties =
+                inner.clone().get_account(address).unwrap_or_default();
+
+            if let Some(nonce) = acc_override.nonce {
+                set_properties_nonce(&mut base, nonce);
+            }
+
+            if let Some(balance) = acc_override.balance {
+                use zk_os_api::helpers::set_properties_balance;
```
itegulov: Why import here?

lib/storage_api/src/state_override_view.rs
```diff
@@ -0,0 +1,113 @@
...
+                set_properties_balance(&mut base, balance);
+            }
+
+            if let Some(code) = acc_override.code {
+                let bytecode_preimage = set_properties_code(&mut base, &code.to_vec());
```
itegulov:
```suggestion
                let bytecode_preimage = set_properties_code(&mut base, &code);
```

integration-tests/tests/call.rs
dimazhornyk: yep, I changed the number and forgot about the comment
fIxed

lib/rpc/src/eth_call_handler.rs
```diff
@@ -205,7 +205,18 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+
+        // If nonce is not provided and override supplies one for `from`, apply it.
+        if request.nonce.is_none()
+            && let Some(so) = &state_overrides
+            && let Some(from) = request.from
+            && let Some(acc) = so.get(&from)
+            && let Some(n) = acc.nonce
+        {
+            request.nonce = Some(n);
+        }
```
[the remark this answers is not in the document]
dimazhornyk: makes sense, removed

lib/rpc/src/sandbox.rs
```diff
@@ -18,10 +18,10 @@ pub const STACK_SIZE: usize = 1024;
 /// zksync-os ergs per gas.
 pub const ERGS_PER_GAS: u64 = 256;
 
-pub fn execute(
+pub fn execute<V: ViewState>(
     tx: ZkTransaction,
     mut block_context: BlockContext,
-    state_view: impl ViewState,
+    state_view: V,
```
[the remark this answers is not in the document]
dimazhornyk: an artifact from the previous implementation approach, removed

lib/storage_api/src/state_override_view.rs
```diff
@@ -0,0 +1,112 @@
...
+            inner,
+            overrides,
+            account_key_to_hash,
+            preimage_overrides,
+        }
+    }
+}
+
+impl<V: ViewState> ReadStorage for OverriddenStateView<V> {
+    fn read(&mut self, key: B256) -> Option<B256> {
```
RomanBrodetski: why do we have `overrides` separate from `account_key_to_hash`? I understand they are coming from different places in user input (they don't use raw storage slot addresses when overriding accounts), but it seems like we use them in the same way (checking raw overrides first aand account_key_to_hash next). Can we just merge them then? ie populate in `new`

RomanBrodetski: I think ultimately you just need a mapping from `StateOverride` to `(overrides, preimage_overrides)`. You may want to use some intermediary things like `AccountViewOverride` but I don't see a reason for them to be present in the codebase on a higher level than local variable in some function.

RomanBrodetski (approved):

PR #547, opened 2025-10-16
chore: Enhance sentry events & tracing

Artemka374:
This PR changes the way events are submitted to Sentry, so we will have the following structure of message
* Title: Error/Warning/panic
* Message: Actual message logged in trace

All the other required debugging information will be contained in tracing fields.

To prevent receiving a lot of (not always readable)text in messages, this PR also changes the way we are logging warnings and errors:
* Actual message contains only short description of what happened
* All the non-constant values are moved to tracing fields, *except* components name - so the message will clearly state which component is failed

Instead of sending `POD_NAMESPACE` as environment tag, send it as separate namespace tag.

### DIFF
--- a/node/bin/src/sentry.rs
+++ b/node/bin/src/sentry.rs
@@ -14,17 +14,12 @@ pub fn init_sentry(url: &str) -> sentry::ClientInitGuard {
             );
 
             if event.exception.is_empty() {
-                let ty = match event.level {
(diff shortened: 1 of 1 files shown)


RomanBrodetski (commented):
left a small comment

RomanBrodetski (approved):
left a small comment

PR #548, opened 2025-10-17
feat(l1-watcher): move pagination/polling into shared component

itegulov:
Reduced the amount of copy-pasted code in l1-watcher. Now main tick/poll loop lives inside the shared `L1Watcher` component and specific flavours (commit/execute/priority-tx) just provide a hook to process fetched events: `process_event`

### DIFF
--- a/lib/l1_watcher/src/watcher.rs
+++ b/lib/l1_watcher/src/watcher.rs
@@ -42,16 +42,16 @@ impl<Processor: ProcessL1Event> L1Watcher<Processor> {
     }
 
     async fn poll(&mut self) -> Result<(), L1WatcherError<Processor::Error>> {
-        // Proactively iterate while there are more blocks to process.
-        loop {
-            let latest_block = self.zk_chain.provider().get_block_number().await?;
+        let latest_block = self.zk_chain.provider().get_block_number().await?;
+
+        while self.next_l1_block <= latest_block {
             let from_block = self.next_l1_block;
             // Inspect up to `self.max_blocks_to_process` blocks at a time
             let to_block = latest_block.min(from_block + self.max_blocks_to_process - 1);
-            if from_block > to_block {
-                return Ok(());
-            }
-            let events = self.process_l1_blocks(from_block, to_block).await?;
+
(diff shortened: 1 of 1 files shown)


lib/l1_watcher/src/watcher.rs
```diff
@@ -3,64 +3,68 @@ use alloy::primitives::BlockNumber;
...
+        // Proactively iterate while there are more blocks to process.
+        loop {
+            let latest_block = self.zk_chain.provider().get_block_number().await?;
+            let from_block = self.next_l1_block;
+            // Inspect up to `self.max_blocks_to_process` blocks at a time
+            let to_block = latest_block.min(from_block + self.max_blocks_to_process - 1);
+            if from_block > to_block {
+                return Ok(());
+            }
+            let events = self.process_l1_blocks(from_block, to_block).await?;
```
RomanBrodetski: nit: seems like we use the term `process` for two different meanings - processing l1 blocks (this method), but also `process_event(event)` - that is, individual delegation to the concrete processor. Not a big deal, but a little confusing that actual `process`ing doesn't happen inside `process_l1_blocks` method.

lib/l1_watcher/src/watcher.rs
RomanBrodetski: nit let's also maybe track the l1 block range that was processed?

lib/l1_watcher/src/tx_watcher.rs
```diff
@@ -51,54 +48,19 @@ impl L1TxWatcher {
...
-        Ok(Self {
-            l1_watcher,
+        let this = Self {
             next_l1_priority_id,
             output,
-            poll_interval: config.poll_interval,
-        })
-    }
+        };
+        let l1_watcher = L1Watcher::new(
```
RomanBrodetski: I guess we could also extract the init logic (including binsearch) from individual watchers to the base l1_watcher. Currently the control flow starts inside the subclasses - and then they pass it over to the base class, passing `this` as a parameter.
Can of course be done seaprately/later

PR #549, opened 2025-10-17
feat: pubdata price calculation

perekopskiy:
- adds gas adjuster that poll for new base fees, provides pubdata price to block context
- uses true pending block context in estimate_gas in API
- replace hardcoded implementations for eth_gasPrice/eth_feeHistory


lib/sequencer/src/execution/block_context_provider.rs
```diff
@@ -56,8 +59,10 @@ impl<Mempool: L2TransactionPool> BlockContextProvider<Mempool> {
         node_version: semver::Version,
         genesis: Arc<Genesis>,
         fee_collector_address: Address,
-        base_fee_override: Option<u64>,
-        pubdata_price_override: Option<u64>,
+        base_fee_override: Option<u128>,
+        pubdata_price_override: Option<u128>,
+        pubdata_price_provider: Arc<dyn PubdataPriceProvider>,
```
popzxc: What if we replace `Arc<dyn PubdataPriceProvider>` with [`watch::Receiver<Option<PubdataPrice>>`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html)?

This way we:
1. won't need an interface for `PubdataPriceProvider` (which doesn't seem an abstraction we're confident in)
2. can turn `GasAdjuster` into a completely detached actor, and for ENs just provide a receiver that is never updated (always has `None`)

lib/gas_adjuster/src/provider.rs
```diff
@@ -0,0 +1,61 @@
+use crate::GasAdjuster;
```
popzxc: This no longer needs to be a separate module, you can move it to the main impl.

lib/gas_adjuster/src/tests.rs
```diff
@@ -0,0 +1,23 @@
+use super::GasStatisticsInner;
```
popzxc: If you will move gas statistics to a separate module, it makes sense to move the tests to the inline module within this file.

lib/gas_adjuster/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+                    );
+                }
+            } else {
+                attempts_failed_in_a_row = 0;
+            }
+            timer.tick().await;
+        }
+    }
+
+    fn gas_price_inner(&self) -> u128 {
```
popzxc: No longer "inner" I guess?

lib/gas_adjuster/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+        // Subtracting 1 from the "latest" block number to prevent errors in case
+        // the info about the latest block is not yet present on the node.
+        // This sometimes happens on Infura.
+        let current_block = self.provider.get_block_number().await?.saturating_sub(1);
+
+        let last_processed_block = self.base_fee_statistics.last_processed_block();
+
+        if current_block > last_processed_block {
+            let n_blocks = current_block - last_processed_block;
+            let fee_data = Self::base_fee_history(&self.provider, current_block, n_blocks).await?;
```
popzxc: Given that you already have to provide a field from `&self`, maybe it makes sense to add the receiver type to the method? Keeping it static doesn't seem to bring value tbh.

PR #563, opened 2025-10-25
feat: config option to run l1 senders in dry run mode

RomanBrodetski:
sometimes it's useful to run server without sending anything to L1. This mode may also be used on EN that doesn't have access to the operator private keys. 

This is particularly useful when mitigaiting an incident


lib/l1_sender/src/run.rs
```diff
@@ -0,0 +1,262 @@
+use crate::batcher_model::{BatchEnvelope, FriProof};
```
RomanBrodetski: this code is moved from `lib.rs` - but otherwise kept as is

perekopskiy:
@RomanBrodetski is it still relevant?

RomanBrodetski:
stale so stale

PR #565, opened 2025-10-29
feat: add block rebuild options

perekopskiy:
- adds configs that allows rebuilding blocks that were not yet committed effectively causing a reorg
- adjusts storages so that they drop data for newer blocks, with a couple of exceptions:
    - replay storage and FullDiffs state overrides data only in case of "block rebuild" for main node
 - panics in case some signer has a rejected tx and then accepted tx, to avoid unprovable block. It's not directly related to block reversion, but the change is small so included here

### DIFF
--- a/node/bin/src/lib.rs
+++ b/node/bin/src/lib.rs
@@ -269,6 +269,7 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
         // Starting from 1 used to work but with the current compacted state implementation it panics.
         // The approach has to be re-evaluated if we want to keep compacted storage impl and what to do in this case.
         // For now just panic early.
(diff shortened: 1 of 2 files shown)


node/bin/src/lib.rs
```diff
@@ -244,18 +253,35 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+                .map(|c| c.from_block),
+        )
         .min()
         .unwrap()
+        .max(1)
     };
 
     let starting_block = if desired_starting_block < state.block_range_available().start() + 1 {
-        tracing::warn!(
+        // Starting from 1 used to work but with the current compacted state implementation it panics.
```
popzxc: Mark with TODO?

PR #569, opened 2025-10-30
feat: JS tracer

dimazhornyk:
This PR adds a JS tracer implementation.
It uses Boa JS runtime to call the JS functions when zksync-os execution triggers a corresponding hook.

State access is implemented by using the captured StateView and storage overlays in the JS context.

## Release Notes

* **New Features**
  * Added JavaScript-based EVM transaction tracing with advanced debugging capabilities, including state inspection and call stack tracking.
  * Enabled state override support for transaction debugging and tracing operations.
  * Enhanced contract instrumentation with calculation event logging and result tracking.

* **Tests**
  * Added comprehensive integration tests for JavaScript tracer validation across multiple scenarios and node configurations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3029,7 +3029,7 @@ dependencies = [
  "iana-time-zone",
  "num-traits",
  "serde",
- "windows-link",
+ "windows-link 0.1.3",
 ]
 
 [[package]]
@@ -3146,6 +3146,7 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+ "terminal_size",
 ]
 
 [[package]]
@@ -4198,7 +4199,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "778e2ac28f6c47af28e4907f13ffd1e1ddbd400980a9abd7c8df189bf578a5ad"
 dependencies = [
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -5271,7 +5272,7 @@ checksum = "a56f203cd1c76362b69e3863fd987520ac36cf70a8c92627449b2f64a8cf7d65"
 dependencies = [
  "cfg-if",
  "libc",
- "windows-link",
+ "windows-link 0.1.3",
 ]
 
 [[package]]
@@ -6118,7 +6119,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "07033963ba89ebaf1584d767badaa2e8fcec21aedea6b8c0346d487d49c28667"
 dependencies = [
  "cfg-if",
- "windows-targets 0.52.6",
+ "windows-targets 0.53.5",
 ]
 
 [[package]]
@@ -7831,7 +7832,7 @@ dependencies = [
  "once_cell",
  "socket2",
  "tracing",
- "windows-sys 0.59.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -9184,7 +9185,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.9.4",
- "windows-sys 0.59.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -10074,6 +10075,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "smart-config-commands"
+version = "0.4.0-pre"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "51f4b028f80f6667d297c09a2c781f7f4a565c3bcdb783203553b12a6cb0d99c"
+dependencies = [
+ "anstream",
+ "anstyle",
+ "serde_json",
+ "serde_yaml",
+ "smart-config",
+]
+
+[[package]]
 name = "smart-config-derive"
 version = "0.4.0-pre"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -10422,6 +10436,16 @@ dependencies = [
  "once_cell",
  "rustix 1.0.8",
  "windows-sys 0.59.0",
+]
+
+[[package]]
+name = "terminal_size"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "60b8cb979cb11c32ce1603f8137b22262a9d131aaa5c37b5678025f22b8becd0"
+dependencies = [
+ "rustix 1.0.8",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -11713,7 +11737,7 @@ dependencies = [
  "windows-collections",
  "windows-core",
  "windows-future",
- "windows-link",
+ "windows-link 0.1.3",
  "windows-numerics",
 ]
 
@@ -11734,7 +11758,7 @@ checksum = "c0fdd3ddb90610c7638aa2b3a3ab2904fb9e5cdbecc643ddb3647212781c4ae3"
 dependencies = [
  "windows-implement",
  "windows-interface",
- "windows-link",
+ "windows-link 0.1.3",
  "windows-result",
  "windows-strings",
 ]
@@ -11746,7 +11770,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc6a41e98427b19fe4b73c550f060b59fa592d7d686537eebf9385621bfbad8e"
 dependencies = [
  "windows-core",
- "windows-link",
+ "windows-link 0.1.3",
  "windows-threading",
 ]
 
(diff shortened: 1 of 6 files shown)


0xVolosnikov:
Would be nice to compare it to the Geth/Reth implementation somehow - it's very easy to introduce some divergence in some obscure corner case

0xVolosnikov:
Also, do we want to limit this functionality somehow? 

- Protect from DDoS spamming trace requests
- Reject huge / malicious JS payloads?

lib/rpc/src/js_tracer.rs
```diff
@@ -0,0 +1,811 @@
...
+    }
+
+    format!("0x{v:x}")
+}
+
+fn anyhow_error_to_js_error(e: anyhow::Error) -> JsError {
+    JsError::from_opaque(JsValue::from(JsString::from(e.to_string())))
+}
+
+/// This would've been a better fit in the zksync-os repo, so in can be reused by other components
```
0xVolosnikov: https://github.com/matter-labs/zksync-os/blob/da0843854e5c02a71a8f25d80d065330db1b97e8/evm_interpreter/src/opcodes.rs#L189

dimazhornyk: replaced it, thanks!!

lib/rpc/src/js_tracer.rs
```diff
@@ -0,0 +1,811 @@
...
+            "gas": gas,
+            "value": match request.modifier() {
+                CallModifier::Static => JsonValue::Null,
+                _ => serde_json::to_value(request.nominal_token_value()).unwrap_or(JsonValue::Null),
+            },
+            "input": Bytes::copy_from_slice(request.input()),
+            "depth": self.current_depth,
+        });
+
+        _ = self.call_method("enter", &obj);
```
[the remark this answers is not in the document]
dimazhornyk: Geth bubbles up the error immediately
since our traits don't allow us to return an error in the tracer methods - I modified the code to record the error and then check it after the execution

integration-tests/tests/tracing.rs
```diff
@@ -414,3 +417,126 @@ async fn call_trace_block() -> anyhow::Result<()> {
...
+    call_request.max_priority_fee_per_gas = Some(1);
+    call_request.max_fee_per_gas = Some(u128::MAX);
+    call_request.set_from(tester.l2_wallet.default_signer().address());
+
+    let trace = tester
+        .l2_provider
+        .debug_trace_call(call_request, BlockId::latest(), opts)
+        .await?;
+
+    println!("trace: {trace:?}");
```
itegulov: Same here

lib/rpc/Cargo.toml
```diff
@@ -40,3 +42,7 @@ vise.workspace = true
 tower.workspace = true
 tower-http = { workspace = true, features = ["cors"] }
 hyper.workspace = true
+
+# JS engine for custom tracers
+boa_engine = { version = "0.21.0" }
+boa_gc = { version = "0.21.0" }
```
itegulov: Please move version pinning to workspace and use `workspace = true` here

lib/rpc/src/js_tracer.rs
```diff
@@ -0,0 +1,681 @@
...
+    fn before_evm_interpreter_execution_step(
+        &mut self,
+        opcode: u8,
+        _frame_state: impl EvmFrameInterface,
+    ) {
+        let obj = serde_json::json!({
+            "op": zk_os_evm_interpreter::opcodes::OPCODE_JUMPMAP[opcode as usize].unwrap_or("Invalid opcode"),
+            "depth": self.current_depth,
+        });
+        self.invoke_method("step", &obj);
```
itegulov: `log` object is missing some fields/methods compared to geth/reth. So I guess we are not fully compatible, right? If that's the case I think it's important to document the differences and leave TODOs where compromises have been made.

lib/rpc/src/js_tracer.rs
```diff
@@ -0,0 +1,681 @@
...
+        }
+    }
+
+    fn before_evm_interpreter_execution_step(
+        &mut self,
+        opcode: u8,
+        _frame_state: impl EvmFrameInterface,
+    ) {
+        let obj = serde_json::json!({
+            "op": zk_os_evm_interpreter::opcodes::OPCODE_JUMPMAP[opcode as usize].unwrap_or("Invalid opcode"),
```
itegulov: Also, this is technically supposed to be a JS object with `toString`/`toNumber`/`isPush`

lib/rpc/src/js_tracer.rs
dimazhornyk: yep, I refactored the code to match the geth API as much as possible
there is still some incompatibility, it is documented at the top of the tracer implementation

I will run some tests now to find all the cases of incompatibility

lib/rpc/src/debug_impl.rs
```diff
@@ -77,18 +70,52 @@ impl<RpcStorage: ReadRpcStorage> DebugNamespace<RpcStorage> {
...
+                            TraceResult::new_success(GethTrace::JS(value), Some(*tx_hash))
+                        })
+                        .collect()),
+                    Err(err) => {
+                        tracing::error!(?err, "Failed to trace transaction with JS tracer");
+                        Err(DebugError::InternalError)
+                    }
+                }
+            }
+            _ => unreachable!("checked above"),
```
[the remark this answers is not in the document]
dimazhornyk: makes sense, returned a proper error instead

lib/rpc/src/js_tracer/host.rs
```diff
@@ -0,0 +1,341 @@
...
+        .unwrap_or_default();
+
+    let Some(address) = parse_address(addr) else {
+        return Ok("0x0".to_string());
+    };
+
+    let balance = env
+        .state_view
+        .borrow_mut()
+        .get_account(address)
```
perekopskiy: Shouldn't there be overlay for balances as well?

dimazhornyk: good catch, added balance overlays as well

lib/rpc/src/js_tracer/tracer.rs
```diff
@@ -0,0 +1,835 @@
...
+///
+/// The JS tracer can use the `db` object to query the state via the following interface:
+///   - getBalance(address): returns balance of an address
+///   - getNonce(address): returns the nonce as hex string
+///   - getCode(address): returns code at address
+///   - getState(address, slot): returns storage value at slot
+///   - exists(address): returns true if the address exists in the state or overlays
+///
+/// Known divergences from geth tracer interface:
+/// - `stack` is not provided in step/fault logs
```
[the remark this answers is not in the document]
dimazhornyk: shall we add it to the wrapper? https://github.com/matter-labs/zksync-os-interface/pull/32

lib/rpc/src/js_tracer/tracer.rs
```diff
@@ -0,0 +1,835 @@
...
+        self.current_depth = 0;
+        self.pending_step = None;
+        self.pending_create_type = None;
+        self.last_finished_frame = None;
+
+        let config = self.tracer_config.clone();
+        self.invoke_method(TracerMethod::Setup, &config);
+    }
+
+    fn finish_tx(&mut self) {
```
0xVolosnikov: We also should cleanup accounts marked for selfdestruct at this step. 
Note: in EVM we delete accounts only if SELFDESTRUCT is called in the same tx where the corresponding contract is deployed. So we do not destruct contracts that were deployed in a different tx

lib/rpc/src/js_tracer/tracer.rs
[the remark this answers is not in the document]
0xVolosnikov: Ah, it still requires some changes in ZKsync OS to actually wrap it. Ok, it makes sense to also create a corresponding PR in zksync-os, I'll do it tomorrow

lib/rpc/src/js_tracer/tracer.rs
```diff
@@ -493,6 +532,13 @@ impl EvmTracer for JsTracer {
             self.pending_create_type = Some(CreateType::Create);
         }
 
+        let call_value = request.nominal_token_value();
+        if call_value != U256::ZERO {
+            let value_i256 = I256::from(call_value);
+            self.apply_balance_delta(request.caller(), value_i256.neg());
+            self.apply_balance_delta(request.callee(), value_i256);
+        }
```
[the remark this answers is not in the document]
0xVolosnikov: We should provide an access to the internal state to tracers. We can't do it now, but it's something to consider in the future

lib/rpc/src/js_tracer/tracer.rs
```diff
@@ -0,0 +1,916 @@
...
+    frame_stack: Vec<FrameState>,
+    last_finished_frame: Option<TxContext>,
+    tx_failed: bool,
+
+    error: Option<anyhow::Error>,
+}
+
+impl JsTracer {
+    pub fn new(state_view: impl ViewState + 'static, js_cfg: String) -> anyhow::Result<Self> {
+        if js_cfg.len() > MAX_JS_TRACER_PAYLOAD_BYTES {
```
0xVolosnikov: What about other DoS vectors, e.g. infinite loops in JS payload?

PR #577, opened 2025-10-31
feat: Protocol upgrade support for provers

EmilLuta:
RATIONALE: Allow protocol upgrades to happen, without provers trying to pick up "wrong batches" (read as batches that it cannot prove).

The changes force server <-> prover interaction to be VK based.

The new VK concept is a 1o1 mapping between `ExecutionVersion` and `VerificationKeyHash`. Provers by default will provide a list of VKs they support and server will provide jobs that correspond to such versions.

This is a "breaking" change, as in, communication between the 2 components changes. The implementation is backwards compatible (legacy API & v1 API).

The plan is to remove legacy API by end of Q4 2025. Until then, the code is a bit "dirty" as we need to support 2 versions.

1 implication/draw-back of current implementation: We won't be able to generate proof input on ExecutionVersion A and prove it with ExecutionVersion B (that implies deprecating & removing the execution_version on proofs).

### DIFF
--- a/lib/multivm/src/lib.rs
+++ b/lib/multivm/src/lib.rs
@@ -14,7 +14,7 @@ use zksync_os_interface::types::{BlockOutput, TxOutput};
 
 pub mod apps;
 
-#[derive(Debug, Clone, Copy, TryFromPrimitive)]
+#[derive(Debug, Clone, Copy, TryFromPrimitive, PartialEq)]
 #[repr(u32)]
 pub enum ExecutionVersion {
     V1 = 1,
@@ -122,6 +122,13 @@ pub fn simulate_tx<Storage: ReadStorage, PreimgSrc: PreimageSource, Tracer: AnyT
     }
 }
 
+/// Method to decide what execution version/VK should the prover use.
+///
+/// Generally speaking, we could have a single execution version, the one used by the server.
+/// There's an edge case where we have a circuit bug and it would require us to prove a batch
+/// with a different execution version than the one it was generated/sealed.
+/// For such cases, we have this mapping defined here, that will allow us to release patch versions that will say:
+/// "oh, you executed with v3? np, prove with v4 as v3 proving is bugged".
 pub fn proving_run_execution_version(forward_run_execution_version: u32) -> ExecutionVersion {
     let forward_run_execution_version: ExecutionVersion = forward_run_execution_version
         .try_into()
--- a/node/bin/src/prover_api/fri_job_manager.rs
+++ b/node/bin/src/prover_api/fri_job_manager.rs
@@ -45,8 +45,9 @@ pub enum SubmitError {
     UnknownJob(u64),
     #[error("deserialization failed: {0:?}")]
     DeserializationFailed(bincode::error::DecodeError),
-    #[error("verification key hash mismatch - server has {0}, got from prover {1}")]
-    VerificationKeyHashMismatch(String, String),
+    // server execution version, prover execution version
+    #[error("execution error mismatch - server expects {0:?}, but got {1:?} from prover")]
+    ExecutionVersionMismatch(ExecutionVersion, ExecutionVersion),
     #[error("internal error: {0}")]
     Other(String),
 }
@@ -196,9 +197,11 @@ impl FriJobManager {
                 Ok(env) => {
                     let env = env.with_stage(BatchExecutionStage::FriProverPicked);
                     let prover_input = env.data.clone();
+                    let proving_execution_version =
+                        proving_run_execution_version(env.batch.execution_version);
                     let fri_job = FriJob {
                         batch_number: env.batch_number(),
-                        vk_hash: env.batch.verification_key_hash().to_string(),
+                        vk_hash: proving_execution_version.vk_hash().to_string(),
                     };
                     tracing::info!(
                         fri_job.batch_number,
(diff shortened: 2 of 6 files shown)


node/bin/src/prover_api/prover_server/legacy/handlers.rs
```diff
@@ -1,84 +1,35 @@
...
+        AppState,
+        legacy::models::{
+            BatchDataPayload, FailedProofResponse, FriProofPayload, NextSnarkProverJobPayload,
+            ProverQuery, SnarkProofPayload,
+        },
+    },
+};
 
-async fn pick_fri_job(State(state): State<AppState>) -> Response {
+pub(in crate::prover_api::prover_server::legacy) async fn pick_fri_job(
```
perekopskiy: pub(super)

lib/l1_sender/src/batcher_model.rs
```diff
@@ -94,6 +96,20 @@ impl<E> BatchEnvelope<E> {
             latency_tracker: self.latency_tracker,
         }
     }
+
+    /// Get Verification Key Hash associated with this batch's execution version.
+    pub fn verification_key_hash(&self) -> anyhow::Result<VerificationKeyHash> {
```
perekopskiy: This method can cause a lot of confusion because there are `batch.execution_version` and also there is `proving_execution_version` inside the proof. I'm 100% sure, but you shouldn't use `batch.execution_version` for proving at all. Anyway, please make it clear which one is used in method name

lib/multivm/src/lib.rs
```diff
@@ -22,6 +25,33 @@ pub enum ExecutionVersion {
     V3 = 3,
 }
 
+// Verification Key Hash and Execution Version have a 1-to-1 mapping starting from V3 onwards.
+// We could backport for V1 and V2 if needed in the future, but it's very unlikely.
```
EmilLuta: I hardly find any value in backporting, as we'll be using the "vk versioned" system from V3 onwards. That said, happy to hear more thoughts.

EmilLuta:
Another aspect worth mentioning, given backwards compatibility, there're a bunch of expects (we should have the version all across the board and I expect us to, but maybe I missed something or there are some specific edge-cases where we might not have it). This makes me feel slightly uneasy that something might go wrong during upgrade, but I reckon it can be tested on stage. Redoing an internal protocol upgrade locally is rather convoluted ATM.

lib/multivm/src/verification_key_hash.rs
```diff
@@ -0,0 +1,50 @@
...
+#[derive(Debug, PartialEq, Copy, Clone)]
+pub struct VerificationKeyHash(&'static str);
+
+impl Display for VerificationKeyHash {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+macro_rules! define_verification_keys {
```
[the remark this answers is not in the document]
EmilLuta: I mean, yeah, but then the match is going to look rather funky:
```
match hash {
   v1 => if v1 == V1_VK_HASH.0 => Ok(V1_VK_HASH),
   ...
}
```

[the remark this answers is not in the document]
RomanBrodetski: hmmm what match? I think you'll only need the two matches I have above..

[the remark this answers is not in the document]
EmilLuta: Discussed with @RomanBrodetski who preferred no Type Enforcement here. I disagree and commit (I find it much better to know that not any &str can be used as part of APIs throughout the codebase, but it's surface is minimal when we remove the VK checking on server). Additionally will backfill the values.

lib/multivm/src/lib.rs
```diff
@@ -22,6 +22,40 @@ pub enum ExecutionVersion {
     V3 = 3,
 }
 
+impl ExecutionVersion {
+    // NOTE: V1 and V2 have a slight chance of being off as they've been backfilled.
+    // If you find a divergence in what you expect and the actual value, most likely a bug.
+
+    /// verification key hash generated from zksync-os v0.0.21, zksync-airbender v0.4.4 and zkos-wrapper v0.4.3
+    const V1_VK_HASH: &'static str =
```
EmilLuta: I actually find this suboptimal and much rather prefer the typed version (so interfaces don't depend on &'static str that may be anything). That said, can be done as a follow-up. LMK if you have any thoughts on it.

node/bin/src/prover_api/fri_job_manager.rs
```diff
@@ -58,15 +58,34 @@ pub struct FailedFriProof {
...
     pub proof_final_register_values: [u32; 16],
+    // TODO: migrate to String, once legacy is deprecated
+    pub vk_hash: Option<String>,
     pub proof_bytes: Bytes,
 }
 
 #[derive(Debug, Serialize)]
 pub struct JobState {
     pub batch_number: u64,
+    pub vk_hash: String,
```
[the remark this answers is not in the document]
RomanBrodetski: Coming back to this comment after I read the changes in fri job manager - we now have many places where we do a tuple of (u64, string) - we could use such `Job` there instead. Whether you want to change it now is up to you - can merge as is

node/bin/src/prover_api/prover_server/v1/handlers.rs
```diff
@@ -99,13 +47,26 @@ async fn submit_fri_proof(
...
+        )
+    })?;
     match state
         .fri_job_manager
-        .submit_proof(payload.block_number, proof_bytes.into(), prover_id)
+        .submit_proof(payload.block_number, proof_bytes.into(), Some(execution_version), &prover_id)
         .await
     {
         Ok(()) => Ok((StatusCode::NO_CONTENT, "proof accepted".to_string()).into_response()),
+        Err(SubmitError::VerificationKeyHashMismatch(server_vk, prover_vk)) => Err((
```
RomanBrodetski: maybe it's worth having execution versions here, not VKs? They are easier to read. We can have both if needed.

node/bin/src/prover_api/prover_server/v1/models.rs
```diff
@@ -0,0 +1,52 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub(super) struct BatchDataPayload {
+    pub block_number: u64,
```
RomanBrodetski: this is batch_number, no? I was thinking we can rename now since we introduce a new interface version

node/bin/src/prover_api/fri_job_manager.rs
```diff
@@ -172,14 +200,15 @@ impl FriJobManager {
                     let env = env.with_stage(BatchExecutionStage::FriProverPicked);
                     let batch_number = env.batch_number();
                     let prover_input = env.data.clone();
+                    let vk_hash = env.batch_metadata_verification_key_hash();
```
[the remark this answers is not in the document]
EmilLuta: As mentioned in previous comments, it's the only way to keep sequencer<->prover communication sane. Otherwise, it's impossible to tell what the prover should prove a job with. We can come back to this decision and revert it, but as I stated before, I strongly believe that execution version should be consistent between server/prover. I see no benefit to diverge.

node/bin/src/prover_api/prover_server/v1/handlers.rs
EmilLuta: It's possible that prover_vk has no equivalent for ExecutionVersion (such as `0xblabla`). We can have options, or back port it. I find VKs more obvious (it's the "id" that tells you uniquely what it means), but happy to have both (assuming we do ExecutionVersion as Option<>).

EmilLuta:
As agreed with @RomanBrodetski, server will decide the execution version at the start of the chain (FRIs) with provers driving versioning forward from there.

PR #582, opened 2025-11-03
feat!: upgrade system (part 1 of N)

popzxc:
## Breaking changes

- Changes in `ReplayRecord` stored format (new CFs: `ProtocolVersion` and `ForcePreimages`)
- Changes in replay wire format (new `ReplayWireFormatV5` to work with new replay fields)
- New mandatory config field `genesis_bytecode_supplier_address` (not used yet but needs to point to a real contract)

## What does this PR support?

- Upgrade transactions are fetched from the contract and assembled
- Upgrade transactions are executed by the server once upgrade timestamp is reached.
- Upgrade transaction will be the only tx in the block, and block with upgrade transaction will be the only one in the batch.
- Batch with the new protocol version won't be committed until the protocol upgrade is finalized on L1.
- Easy-to-implement integration tests for different upgrade flows (including EN sync).

## What's NOT implemented?

These things are planned, but not included in this PR to keep the scope reviewable.

- Matching protocol version to execution version (e.g. switching to a new multivm version based on protocol upgrade).
- VKs switch (e.g. fetching new VKs from the protocol upgrade).
- Persistent state for partially fetched upgrade data.
- New multiVM system / artifact registry.

## Caveats

- ⚠️ We introduce breaking changes to both wire format and storage, we need to think on how to roll out this feature.
- ⚠️ Protocol version for genesis is not exposed for ENs right now, so they might not know which version should be used "initially". It's unknown right now if it's a problem, since upgrade tx will be executed in the first block.
- `BytecodesSupplier` is not ready for ZKsync OS, so the implementation assumes that all the bytecodes will be deployed to L2 before the upgrade. If this step won't be done, upgrade transaction will fail as there is no way to fetch preimages from L1.
- Upgrades are expected to be sequential, e.g. node is not guaranteed to correctly process multiple upgrades if somehow the previous upgrade event was missing (e.g. if it happened a lot of time ago).
- Upgrades are expected to be executed in a reasonable time after it was added to CTM (e.g. if upgrade was added to CTM, but executed after a few months, we won't be able to find upgrade data).
- The way upgrade transactions are processed by `BlockContextProvider` is not ideal, it exposes extra interfaces to the best txs stream, which makes it a leaky abstraction. This problem will be solved once we have persistent state for upgrade data.

### Changes to the existing components

- `BlockContextProvider` is reworked to receive upgrade transactions from a channel. Old system with an `Option` for upgrade tx was removed. Protocol version is now included in the produced command, and also became a part of the replay record. Same for forced preimages (bytecodes that must be inserted to state before executing the block, aka factory deps).
- Genesis upgrade is processed in the same way as any other upgrade transaction.
- `PersistentPreimages` no longer need genesis to be initialized; it receives forced preimages from the upgrade transaction.
- `block_executor` will process upgrade transactions and seal the block immediately after it's processed.
- `batch_builder` will seal the batch if it receives a block with upgrade transaction (on both ends -- it will seal the previous batch before starting a new one with upgrade tx, and one with upgrade tx will be sealed with just 1 block)

### New components

- `UpgradeTxWatcher` - part of `l1_watcher`, looks for the upgrade txs and sends them to the stream
- `UpgradeGatekeeper` - part of `l1_sender`, waits until protocol on diamond proxy matches protocol version of the batch

## Testing

This PR introduces a way to (somewhat) declaratively do upgrade tests -- you can build the upgrade yourself, including setting a version, adding force deployed contracts, etc, and then execute. `DefaultUpgrade` is assumed but not forced. More functionality is planned to be added in the future prs (e.g. setting VKs).
For now, 2 tests are added:
- "simple" upgrade (no force deployments)
- "normal" upgrade (with force deployments).

The latter also has a check for ENs to be able to sync.

### DIFF
--- a/lib/l1_watcher/src/upgrade_tx_watcher.rs
+++ b/lib/l1_watcher/src/upgrade_tx_watcher.rs
@@ -347,7 +347,7 @@ async fn find_l1_block_by_protocol_version(
 
     util::find_l1_block_by_predicate(Arc::new(zk_chain), move |zk, block| async move {
         let res = zk.get_raw_protocol_version(block.into()).await?;
-        Ok(res == protocol_version)
+        Ok(res >= protocol_version)
     })
     .await
 }


lib/l1_sender/src/batcher_model.rs
```diff
@@ -34,6 +35,7 @@ pub struct BatchMetadata {
     pub tx_count: usize,
     #[serde(default = "default_execution_version")]
     pub execution_version: u32,
+    pub protocol_version: ProtocolSemanticVersion,
```
RomanBrodetski: note for myself - does it break S3?

integration-tests/src/upgrade/builder.rs
```diff
@@ -0,0 +1,165 @@
...
+        self.timestamp = timestamp;
+        self
+    }
+
+    /// Builds the `ProposedUpgrade` struct.
+    pub fn build(self) -> ProposedUpgrade {
+        let (delegate_to_address, delegate_to_calldata) = self.delegate_to;
+
+        const DEPLOYER_ADDRESS: &str = "0x0000000000000000000000000000000000008007";
+        const COMPLEX_UPGRADER_ADDRESS: &str = "0x000000000000000000000000000000000000800f";
```
itegulov: Nit: might make sense to use `address!` here

itegulov: Just saw that you use it with `L2CanonicalTransaction` later on. Might be awkward then if it expects U256 for these two values...

integration-tests/src/upgrade/builder.rs
```diff
@@ -0,0 +1,165 @@
...
+
+        const REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_LIMIT: u32 = 800;
```
itegulov: There is `zksync_os_types::REQUIRED_L1_TO_L2_GAS_PER_PUBDATA_BYTE` already that represents the same thing

integration-tests/src/upgrade/tester.rs
```diff
@@ -0,0 +1,352 @@
...
+
+    async fn fund_wallets(&self) -> anyhow::Result<()> {
+        for wallet in [self.bridgehub_owner] {
+            self.tester
+                .l1_provider
+                .send_transaction(
+                    TransactionRequest::default()
+                        .with_to(wallet)
+                        .with_value(U256::from(10).pow(U256::from(18u64))), // 1 ETH
+                )
```
itegulov: Nit: there is a more direct `anvil_set_balance` available as well but not a big deal

integration-tests/src/upgrade/tester.rs
```diff
@@ -0,0 +1,352 @@
...
+    pub async fn pause_bridgehub_migrations(&self) -> anyhow::Result<()> {
+        let pause_migration_tx = self
+            .bridgehub
+            .pauseMigration()
+            .into_transaction_request()
+            .with_from(self.bridgehub_owner);
+        let hash = self
+            .tester
+            .l1_provider
+            .anvil_send_impersonated_transaction(pause_migration_tx)
```
itegulov: What does this do exactly? Shouldn't regular send be okay here since you impersonated the account already?

lib/sequencer/src/execution/block_context_provider.rs
```diff
@@ -116,6 +118,28 @@ impl<Mempool: L2TransactionPool> BlockContextProvider<Mempool> {
 
                 let timestamp = (millis_since_epoch() / 1000) as u64;
 
+                // Check if we peeked an upgrade transaction info.
+                let mut force_preimages = Vec::new();
```
RomanBrodetski: nit - let's maybe make it immutable? `let force_preimages = if {} else {vec![]}` . This way it's immediately clear that `force_preimages`  is only populated for upgrades and not mutated further down for any other reason

integration-tests/src/upgrade/builder.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// and set it as the delegate, so this method is only needed if you want to
+    /// execute some specific logic during the upgrade.
+    pub fn delegate_to(mut self, delegate_address: Address, calldata: Bytes) -> Self {
+        self.delegate_to = (delegate_address, calldata);
+        self
+    }
+
+    /// Bump the minor protocol version by the specified amount.
+    pub fn bump_minor(mut self, by: u64) -> Self {
+        let semver_minor_version_multiplier = U256::from(4294967296u64); // 2^32
```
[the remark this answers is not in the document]
popzxc: Fully reworked this part; added a method to pack `ProtocolSemanticVersion` with proper assertions, so now we don't do bit shenanigans and instead rely on a high-level interface.

integration-tests/src/upgrade/tester.rs
popzxc: Encapsulated this logic and added a comment to explain purpose.

integration-tests/src/upgrade/builder.rs
```diff
@@ -0,0 +1,165 @@
...
+        let mut force_deployments: Vec<UniversalForceDeploymentInfo> = Vec::new();
+        let factory_deps = Vec::new();
+
+        for (address, bytecode) in self.force_deployments.unwrap_or_default() {
+            let mut account_properties = AccountProperties::default();
+            set_properties_code(&mut account_properties, &bytecode);
+
+            let deployed_bytecode_info = super::interfaces::ForceDeploymentBytecodeInfo {
+                bytecodeHash: B256::from_slice(account_properties.bytecode_hash.as_u8_ref()),
+                bytecodeSize: U256::from(account_properties.full_bytecode_len()),
```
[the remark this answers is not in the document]
popzxc: The bytecode hash is calculated based on the full length inside of zksync-os, so I thought that it's required.
If I run the test with `observable_len` instead of `full_bytecode_len()`, the test fails with `Forward run failed with: Internal error: Iterator length exceeds expected preimage length` inside of zksync OS.

I'm not 100% sure as I didn't write zksync OS, but by all the looks of it, we do need to provide full bytecode len here.

integration-tests/src/upgrade/tester.rs
```diff
@@ -0,0 +1,352 @@
...
+            .await?;
+        PendingTransactionBuilder::new(self.tester.l1_provider.root().clone(), hash)
+            .expect_successful_receipt()
+            .await?;
+
+        Ok(())
+    }
+
+    pub async fn generic_l2_upgrade_target(&self) -> anyhow::Result<(Address, Bytes)> {
+        // HACK: right now we need to call an account with bytecode to make the upgrade work.
```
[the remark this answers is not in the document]
popzxc: 1. Publishing bytecode doesn't work yet, so we will need to deploy it beforehand anyways :) 
2. This is a generic target, it's only there because contracts don't support upgrades without contract to invoke; if someone will want to write a "production-like" test with the flow you described, they certainly can (as soon as (1) is fixed) -- here we need to make sure that the "default" variant works (e.g. where we don't want any extra logic).

lib/sequencer/src/execution/block_executor.rs
```diff
@@ -209,12 +214,15 @@ pub async fn execute_block<R: ReadStateHistory + WriteState>(
         SealPolicy::UntilExhausted {
             allowed_to_finish_early,
         } => {
-            if !allowed_to_finish_early && seal_reason != SealReason::TxStreamExhausted {
+            if !allowed_to_finish_early
+                && seal_reason != SealReason::TxStreamExhausted
+                && seal_reason != SealReason::UpgradeTx
```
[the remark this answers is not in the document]
popzxc: Reworked the logic to match your proposal.

lib/mempool/src/stream.rs
```diff
@@ -98,11 +110,26 @@ impl BestTransactionsStream<'_> {
     /// Waits until there is a next transaction and returns a reference to it.
     /// Does not consume the transaction, it will be returned on the next poll.
     /// Returns `None` if the stream is closed.
-    pub async fn wait_peek(&mut self) -> Option<&ZkTransaction> {
+    /// Returns `Some(None)` if there is a transaction in the stream, but it's not an upgrade transaction.
+    /// Returns `Some(Some(upgrade_tx))` if the next transaction is an upgrade transaction.
+    // TODO: this interface leaks implementation details about the internal structure, and in general
```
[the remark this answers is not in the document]
popzxc: In (1) -- but we will still need to expose an interface to get upgrade tx metadata, no? Because upgrade tx might arrive when we wait for the first transaction of the block, and it will already be in the stream. The only thing I see being improved here is lack of `txs_already_provided` at cost of more logic in other place.
In (2) -- not sure what you mean here; my plan was to just check that peeked transaction is an upgrade tx, and fetch protocol version from the persistent source. E.g. everything remains the same as it is now (transactions go to the stream), but we don't need to expose the metadata from the stream itself.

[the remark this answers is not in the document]
RomanBrodetski: > In (1) -- but we will still need to expose an interface to get upgrade tx metadata, no? Because upgrade tx might arrive when we wait for the first transaction of the block, and it will already be in the stream. The only thing I see being improved here is lack of txs_already_provided at cost of more logic in other place.

Yes, the tx metadata is (2). And yes, in (1) the improvement is getting rid of `txs_already_provided` - with more logic elsewhere, but IMO it's mildly cleaner. With (b) you'll have individual stream combinators and write a higher level logic to combine them - more code lines in total, but you don't need to read implementation of combinator - it's clear what it does from name/comment (take elem from first stream, if it's empty - take from another and don't return to the first)) whereas with (a) you only add a little of sealing logic to block_execute (ExcludeAndSeal on an Upgrade transaction). But it's just a suggestion, can leave as is

> In (2) -- not sure what you mean here; my plan was to just check that peeked transaction is an upgrade tx, and fetch protocol version from the persistent source. E.g. everything remains the same as it is now (transactions go to the stream), but we don't need to expose the metadata from the stream itself.

Ah yes yes, I was wrong in my message earlier.. Still it'd be nicer to get everything you need from the stream itself - as opposed to doing a lookup (remember we discussed that we only need persistence because of complexity of reading it from L1) - and I think it's possible to achieve that with aforementioned enum and an `Into<ZkTransaction>`, but I'm not even sure it's worth it

lib/l1_watcher/src/upgrade_tx_watcher.rs
```diff
@@ -0,0 +1,311 @@
...
+
+// TODO: disabled until bytecode supplier integration is ready
+// use zksync_os_contract_interface::IBytecodeSupplier::BytecodePublished;
+// use zk_os_api::helpers::set_properties_code;
+// use zk_os_basic_system::system_implementation::flat_storage_model::AccountProperties;
+
+/// Don't try to process that many block linearly
+/// The constant value is higher than for other watchers, since we're looking for rare/specific events
+/// and we don't expect a lot of results.
+const MAX_L1_BLOCKS_LOOKBEHIND: u64 = 2_500_000;
```
itegulov: Top part of the comment doesn't make sense with the current implementation. All other L1 watchers run binsearch to find the starting L1 block (in your case that would be where current protocol version got activated I guess). I know you already expressed your dislike of running binsearch but I feel like the current approach with linear search on last 2.5M blocks on every restart is even more brittle, no?

lib/sequencer/src/execution/block_executor.rs
```diff
@@ -214,7 +227,7 @@ pub async fn execute_block<R: ReadStateHistory + WriteState>(
                     ctx,
                     txs: all_processed_txs.clone(),
                     error: format!(
-                        "block was expected to be sealed due to stream exhaustion, but sealed due to {:?} instead, block {}",
+                        "block was expected to be sealed due to either stream exhaustion, but sealed due to {:?} instead, block {}",
```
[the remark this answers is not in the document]
popzxc: Leftover from [this](https://github.com/matter-labs/zksync-os-server/pull/582#discussion_r2514166414)

integration-tests/src/upgrade/builder.rs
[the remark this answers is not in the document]
StanislavBreadless: Okay, if both pass the tests, lgtm I guess

lib/mempool/src/stream.rs
```diff
@@ -98,11 +110,26 @@ impl BestTransactionsStream<'_> {
...
+    // this information is only needed for the `BlockContextProvider` which already has access to the stream.
+    // This was introduced only because upgrade transaction can appear after we started waiting for the
+    // first tx, and we need protocol upgrade info to initialize block context.
+    // Consider refactoring this later.
+    pub async fn wait_peek(&mut self) -> Option<Option<UpgradeTransaction>> {
         if self.peeked_tx.is_none() {
             self.peeked_tx = self.next().await;
+            self.txs_already_provided = true; // TODO: implicit expectation that this method is _guaranteed_ to be called before using the stream.
```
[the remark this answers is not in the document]
itegulov: Thinking about it a bit more there might be some opportunity there to refactor shenanigans with enforcing upgrade tx always being first in the stream. Because you'll have invariant that there is at least one transaction by the time `BestTransactionsStream` is initialized. But not a strong opinion.

lib/sequencer/src/execution/mod.rs
```diff
@@ -103,6 +103,15 @@ where
...
 
+            self.state
+                .force_add_preimages(
+                    prepared_command
+                        .force_preimages
+                        .iter()
+                        .map(|(k, v)| (*k, v)),
+                )
+                .context("force_add_preimages")?;
+
```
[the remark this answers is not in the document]
popzxc: That's a good point, did it this way and was able to remove quite some code.
Thanks!

lib/l1_watcher/src/upgrade_tx_watcher.rs
popzxc: Good point, already working on changing this part.

lib/l1_watcher/src/upgrade_tx_watcher.rs
```diff
@@ -0,0 +1,307 @@
...
+use zksync_os_contract_interface::ZkChain;
+use zksync_os_types::{L1UpgradeEnvelope, ProtocolSemanticVersion, UpgradeTransaction};
+
+// TODO: disabled until bytecode supplier integration is ready
+// use zksync_os_contract_interface::IBytecodeSupplier::BytecodePublished;
+// use zk_os_api::helpers::set_properties_code;
+// use zk_os_basic_system::system_implementation::flat_storage_model::AccountProperties;
+
+/// Don't try to process that many block linearly
+const MAX_L1_BLOCKS_LOOKBEHIND: u64 = 100_000;
```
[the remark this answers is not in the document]
popzxc: Responded in another thread.

lib/l1_watcher/src/upgrade_tx_watcher.rs
```diff
@@ -307,5 +336,18 @@ pub enum UpgradeTxWatcherError {
...
+
+async fn find_l1_block_by_protocol_version(
+    zk_chain: ZkChain<DynProvider>,
+    protocol_version: ProtocolSemanticVersion,
+) -> anyhow::Result<BlockNumber> {
+    let protocol_version = protocol_version.packed()?;
+
+    util::find_l1_block_by_predicate(Arc::new(zk_chain), move |zk, block| async move {
+        let res = zk.get_raw_protocol_version(block.into()).await?;
+        Ok(res == protocol_version)
```
itegulov: You need to do `res >= protocol_version` here to find the first block where `protocol_version` got activated. Or `res > protocol_version` to find first block where next protocol version got activated (or it will return `latest + 1` if there is none). I think latter makes more sense for you

Apologies that the method doesn't have any documentation, I can see how easy it would be to confuse that predicate here does not need to be monotonic.

popzxc: Fixed, thanks. I do think that the first block where `protocol_version` got activated makes more sense, since we can have `setUpgradeTimestamp` at any point after this block for the next (not yet processed) upgrade.

PR #585, opened 2025-11-05
chore: add doc comment for block rebuild config

perekopskiy:
add doc comment for block rebuild config

### DIFF
--- a/node/bin/src/config.rs
+++ b/node/bin/src/config.rs
@@ -126,9 +126,9 @@ pub struct StatusServerConfig {
 #[derive(Clone, Debug, DescribeConfig, DeserializeConfig)]
 pub struct RebuildBlocksConfig {
     /// Number of the block to start rebuilding from.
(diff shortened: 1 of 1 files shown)


node/bin/src/config.rs
```diff
@@ -125,7 +125,9 @@ pub struct StatusServerConfig {
 
 #[derive(Clone, Debug, DescribeConfig, DeserializeConfig)]
 pub struct RebuildBlocksConfig {
+    /// Number of the block to start rebuilding from.
```
RomanBrodetski:
```suggestion
    /// Number of the block to start rebuilding from.
    /// All blocks starting from this number will be replayed - but unlike normal replay, 
    /// we'll not assert that the result will match the original ReplayRecord (block).
    /// That is, a block may close earlier (with less transactions), 
    /// have different hash, have some transactions rejected etc
```

RomanBrodetski (approved):

PR #586, opened 2025-11-05
fix: handle exclude and seal for l1 txs

perekopskiy:
Previously, node crashed on exclude and seal for l1 txs, this PR fixes it.
L1 txs channel is wrapped into a struct with a buffer, txs provider populates buffer which is then cleared in `on_canonical_state_change`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for invalid L1 and L2 transactions with enhanced error messaging.
  * Enhanced L1 envelope validation and processing logic for more robust transaction handling.

* **Refactor**
  * Streamlined L1 transaction buffering and draining mechanism for improved efficiency.

### DIFF
--- a/lib/mempool/src/l1_txs_receiver.rs
+++ b/lib/mempool/src/l1_txs_receiver.rs
@@ -48,15 +48,18 @@ impl L1TxsChannel {
     }
 
     /// Drains `count` envelopes from the buffer, awaiting new ones from the inner receiver if necessary.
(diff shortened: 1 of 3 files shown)


itegulov (commented):
Didn't review the PR fully but have you considered peeking txs from BestTransactionsStream instead? We already have that method there and AFAIU using it should help

perekopskiy:
Decided to move on with adding L1 txs to mempool

PR #594, opened 2025-11-10
feat!: v30 zksync os protocol upgrade support

AntonD3:
This PR introduces v30 zksync os protocol upgrade support.

Key enhancements include:
- New zksync-os/execution version
- Support for using blobs as a Data Availability (DA) option
- Updated L1 commit interface to include block range
- Backward-compatible implementation, allowing the server to function seamlessly with both existing and new version

## Summary by CodeRabbit

* **New Features**
  * Adds Execution/Proving V5 support and protocol mappings for minors 30/31.
  * End-to-end pubdata mode (Blobs/Calldata/Validium) propagation across batcher, prover, sender, verifier.
  * Blob-transaction support with blob sidecars and configurable per-blob gas cap.

* **Chores**
  * Genesis input accepts both pretty and raw storage formats (backward-compatible).
  * Workspace/runtime references updated to newer releases; CI tests run in release mode.

## Breaking Changes

New execution version V5 added that maps to protocol version V30.
`L1SenderConfig::rollup_pubdata_mode` renamed to `L1SenderConfig::pubdata_mode`.
Drops support for sealing/proving V3 batches.

## Rollout Instructions

Put correct `l1_sender_pubdata_mode` in config before updating to this version. If you're going to switch to blobs as a part of V30 upgrade then you should put `Blobs` right away, it will be ignored and V28 batches calldata will be used 
For doing protocol upgrade please follow the usual protocol upgrade process.
Make sure chain is at V4 execution version before updating the node.
EN should be updated before main node as old node cannot read new batch envelope format in S3.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7190,8 +7190,8 @@ dependencies = [
 
 [[package]]
 name = "protocol_version"
-version = "0.5.1"
-source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?branch=chore%2Fprotocol-protocol-v1.3%2Fzksync-airbender-prover#07c8fa9007b6b75ec749ade87f505052b0f1552a"
+version = "0.6.0"
+source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?tag=v0.6.0#7dee1169d83790f29bbbb140da893582259b08ca"
 
 [[package]]
 name = "prover"
@@ -11811,8 +11811,8 @@ checksum = "12685073683de64e01a87b0f0a8c1e11df2ea3d6dc3846f1320153ef2fe499ed"
 
 [[package]]
 name = "zksync_os_fri_prover"
-version = "0.5.1"
-source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?branch=chore%2Fprotocol-protocol-v1.3%2Fzksync-airbender-prover#07c8fa9007b6b75ec749ade87f505052b0f1552a"
+version = "0.6.0"
+source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?tag=v0.6.0#7dee1169d83790f29bbbb140da893582259b08ca"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
@@ -12136,8 +12136,8 @@ dependencies = [
 
 [[package]]
 name = "zksync_os_prover_service"
-version = "0.5.1"
-source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?branch=chore%2Fprotocol-protocol-v1.3%2Fzksync-airbender-prover#07c8fa9007b6b75ec749ade87f505052b0f1552a"
+version = "0.6.0"
+source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?tag=v0.6.0#7dee1169d83790f29bbbb140da893582259b08ca"
 dependencies = [
  "anyhow",
  "base64 0.22.1",
@@ -12389,8 +12389,8 @@ dependencies = [
 
 [[package]]
 name = "zksync_os_snark_prover"
-version = "0.5.1"
-source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?branch=chore%2Fprotocol-protocol-v1.3%2Fzksync-airbender-prover#07c8fa9007b6b75ec749ade87f505052b0f1552a"
+version = "0.6.0"
+source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?tag=v0.6.0#7dee1169d83790f29bbbb140da893582259b08ca"
 dependencies = [
  "anyhow",
  "clap",
@@ -12548,8 +12548,8 @@ dependencies = [
 
 [[package]]
 name = "zksync_sequencer_proof_client"
-version = "0.5.1"
-source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?branch=chore%2Fprotocol-protocol-v1.3%2Fzksync-airbender-prover#07c8fa9007b6b75ec749ade87f505052b0f1552a"
+version = "0.6.0"
+source = "git+https://github.com/matter-labs/zksync-airbender-prover.git?tag=v0.6.0#7dee1169d83790f29bbbb140da893582259b08ca"
 dependencies = [
  "anyhow",
  "async-trait",


lib/contract_interface/src/models.rs
```diff
@@ -87,7 +132,7 @@ pub struct CommitBatchInfo {
     pub priority_operations_hash: B256,
     pub dependency_roots_rolling_hash: B256,
     pub l2_to_l1_logs_root_hash: B256,
-    pub l2_da_validator: Address,
+    pub l2_da_commitment_scheme: DACommitmentScheme,
```
perekopskiy: If we don't get rid of batch persistence before this PR (cc @RomanBrodetski, iiuc it won't be done this week or so), then it should be done in a backward-compatible way -- l2_da_commitment_scheme should be optional

lib/l1_sender/src/commitment.rs
```diff
@@ -22,6 +22,9 @@ pub struct BatchInfo {
     /// L1 protocol upgrade transaction that was finalized in this batch. Missing for the vast
     /// majority of batches.
     pub upgrade_tx_hash: Option<B256>,
+    pub da_commitment_scheme: zk_ee::common_structs::DACommitmentScheme,
```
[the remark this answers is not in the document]
perekopskiy: I think you should get rid of `da_commitment_scheme` here. Instead use scheme from `commit_info` and implement converter to zksync-os enum

lib/batch_verification/src/client/mod.rs
```diff
@@ -195,6 +195,7 @@ impl<Finality: ReadFinality> BatchVerificationClient<Finality> {
             self.chain_id,
             self.diamond_proxy,
             request.batch_number,
+            unimplemented!(),
```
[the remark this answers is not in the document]
AntonD3: In fact, we need pubdata mode here, because it can affect operator_da_input, blobs_sidecar, etc. I mean, right now we can rely on the commitment scheme, as we have 1 to 1 mapping of the commitment scheme to pubdata mode, but this invariant is not always guaranteed to be true.

node/bin/src/batcher/mod.rs
```diff
@@ -307,6 +310,7 @@ impl Batcher {
             batch_number,
             self.chain_id,
             self.chain_address,
+            unimplemented!(),
```
AntonD3: Is it ok to use pubdata mode from config(saved in batcher) here?

lib/contract_interface/src/models.rs
AntonD3: Added default value, PTAL

lib/l1_sender/src/commands/execute.rs
```diff
@@ -41,6 +42,10 @@ impl SendToL1 for ExecuteCommand {
             self.to_calldata_suffix().into(),
         ))
     }
+
+    fn blob_sidecar(&self) -> Option<BlobTransactionSidecar> {
```
RomanBrodetski: nit: a default implementation would be handy here (fn blob_sidecar  {`None`})

lib/batch_verification/src/wire_format/v1.rs
```diff
@@ -13,6 +13,7 @@ pub struct BatchVerificationRequestWireFormatV1 {
     pub batch_number: u64,
     pub first_block_number: u64,
     pub last_block_number: u64,
+    pub pubdata_mode: u8,
```
popzxc: Is it safe to edit this file? I guess if 2FA was not rolled out in any envs then yes, but better to make sure with the platform team. There is a chance that you need to introduce a new version of wire format.

lib/types/src/protocol/proving_version.rs
```diff
@@ -83,8 +89,12 @@ impl ProvingVersion {
         forward_run_execution_version: ExecutionVersion,
     ) -> Self {
         match forward_run_execution_version {
-            ExecutionVersion::V1 | ExecutionVersion::V2 | ExecutionVersion::V3 => Self::V3,
+            // TODO: panic?
+            ExecutionVersion::V1 | ExecutionVersion::V2 | ExecutionVersion::V3 => {
+                panic!("Unsupported execution version")
+            }
```
popzxc: Soft -1 on changing pre-existing logic as a part of bigger PR; makes reviewing harder.
I don't have an opinion here, but I also have no confidence if it can potentially break something in production.

lib/types/src/protocol/proving_version.rs
```diff
@@ -109,7 +119,7 @@ mod tests {
         let test_vector = [
             ((0, 29, 0), ProvingVersion::V4),
             ((0, 29, 1), ProvingVersion::V4),
-            ((0, 30, 0), ProvingVersion::V4), // To be updated to V5 once 30 is ready.
+            ((0, 30, 0), ProvingVersion::V5),
```
popzxc: Same

lib/contract_interface/Cargo.toml
```diff
@@ -17,3 +17,4 @@ serde.workspace = true
 tracing.workspace = true
 vise.workspace = true
 structdiff.workspace = true
+zk_ee.workspace = true
```
popzxc: ⚠️⚠️⚠️ This is bad ⚠️⚠️⚠️

This crate is meant to be lightweight and freely usable, but adding `zk_ee` there make it very heavyweight.
This is added for a single conversion, so it's better to implement this conversion at the place where it's actually used and remove this dependency.

lib/types/src/protocol/proving_version.rs
AntonD3: Just didn't want to keep more than 2 supported versions of proving, AFAIU, the previous version of proving was needed only for the previous upgrade, now with the new version, we don't need it. cc @perekopskiy

lib/l1_sender/src/commands/prove.rs
```diff
@@ -128,18 +128,26 @@ impl ProofCommand {
...
+                    .batch
+                    .batch_info
+                    .clone()
+                    .into_stored(&batch.batch.protocol_version)
+            })
             .collect();
         // todo: awful and temporary
         let verifier_version = match self.proof.proving_execution_version() {
             // Use default verifier for fake proofs.
+            // TODO: old removed after upgrade?
```
perekopskiy: reasonable suggestion, we can support the same set of version that we do for prover input generation, i.e. latest and pre-latest

lib/l1_sender/src/lib.rs
```diff
@@ -132,6 +132,21 @@ pub async fn run_l1_sender<Input: SendToL1>(
...
+                        if fee_per_blob_gas > config.max_fee_per_blob_gas_gwei as u128 {
+                            tracing::warn!(
+                                max_fee_per_blob_gas = config.max_fee_per_blob_gas_gwei as u128,
+                                fee_per_blob_gas = fee_per_blob_gas,
+                                "L1 sender's configured maxPriorityFeePerGas is lower than the one estimated from network"
+                            );
+                        }
+                        tx_request.set_max_fee_per_blob_gas(config.max_fee_per_blob_gas_gwei as u128);
+                        tx_request.set_blob_sidecar(blob_sidecar);
+                    };
```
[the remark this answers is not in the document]
perekopskiy: @AntonD3 please make sure it's tackled

node/bin/src/batcher/batch_builder.rs
```diff
@@ -42,17 +43,46 @@ pub(crate) fn seal_batch(
...
-        std::iter::once(u32::try_from(blocks.len()).expect("too many blocks"))
-            .chain(
-                blocks
-                    .iter()
-                    .flat_map(|(_, _, _, prover_input)| prover_input.iter().copied()),
-            )
-            .collect();
+        match ProvingVersion::from_forward_run_execution_version(forward_run_execution_version) {
+            ProvingVersion::V1 | ProvingVersion::V2 | ProvingVersion::V3 => {
+                unreachable!("proving_run_execution_version does not return 1, 2 or 3")
```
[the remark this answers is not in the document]
perekopskiy: please replace with panic and update message

PR #596, opened 2025-11-11
feat: add config for l2 signer blacklist

perekopskiy:
Adds config for l2 signer blacklist

### DIFF
--- a/node/bin/src/config.rs
+++ b/node/bin/src/config.rs
@@ -2,12 +2,14 @@ use crate::command_source::RebuildOptions;
 use alloy::consensus::constants::GWEI_TO_WEI;
 use alloy::primitives::Address;
 use serde::{Deserialize, Serialize};
+use smart_config::de::{Qualified, WellKnown};
 use smart_config::metadata::TimeUnit;
 use smart_config::value::SecretString;
 use smart_config::{
     DescribeConfig, DeserializeConfig, Serde,
     de::{Delimited, Optional},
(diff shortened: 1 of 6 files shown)


node/bin/src/config.rs
```diff
@@ -216,6 +218,16 @@ impl SequencerConfig {
...
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct ConfigAddress(pub Address);
+
+const HASH_DE: Qualified<Serde![str]> =
+    Qualified::new(Serde![str], "hex string with optional 0x prefix");
+impl WellKnown for ConfigAddress {
+    type Deserializer = Qualified<Serde![str]>;
+    const DE: Self::Deserializer = HASH_DE;
+}
```
itegulov: I think you can just bump `smart-config` to `0.4.0` and enable `alloy` feautre. Then you'll just have this "for free"

itegulov (commented):
Do I understand correctly that this set will become dynamic in the future? Maybe it makes sense to introduce some abstraction that hold `Arc<DashMap>` inside?

itegulov:
> There are no plans to do it dymanic for now. Also it's easy to change in the future if needed

I thought we want to blacklist addresses that sent "malicious" (according to 2FA) transactions, no? My assumption was that this PR was the first step for it. If not what's the reasoning for this?

PR #603, opened 2025-11-12
fix(config): add config attributes to fee overrides

dimazhornyk:
Fee overrides were not configurable, the values from the ENV couldn't be converted to the config. 
This PR also changes the type of the configs to U256 to align with the usage of the config vars in the code.

### DIFF
--- a/node/bin/src/config.rs
+++ b/node/bin/src/config.rs
@@ -1,6 +1,6 @@
 use crate::command_source::RebuildOptions;
 use alloy::consensus::constants::GWEI_TO_WEI;
-use alloy::primitives::{Address, U256};
+use alloy::primitives::{Address, U128};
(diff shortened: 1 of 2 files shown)


node/bin/src/config.rs
```diff
@@ -178,13 +178,16 @@ pub struct SequencerConfig {
     pub fee_collector_address: Address,
 
     /// Override for base fee (in wei). If set, base fee will be constant and equal to this value.
-    pub base_fee_override: Option<u128>,
+    #[config(default_t = None, with = Optional(Serde![str]))]
+    pub base_fee_override: Option<U256>,
```
perekopskiy: let's use U128, alloy types expect fees to be in u128 range so large U256 shouldn't be allowed

PR #612, opened 2025-11-16
feat: Read force deploys from a file

StanislavBreadless:
Since we do not yet support reading bytecodes using `BytecodeSupplier` for zksync os, to temporarily enable upgrades, we will do the following:
- Each time there is a protocol upgrade tx, we just read the hardcoded JSON file and include all the bytecodes from there to get the list of bytecode hashes and their preimages.
- The correctness of the file is ensured by the person that prepares the upgrade (most likely me). Hopefully this hack will not last for long.
- The file is included using `include_str!` so it is included into the binary and so the binary can be shipped as-is just as before

PS: the file *will* change in the subsequent PRs/commits until the code of contracts is frozen.

### DIFF
--- a/lib/l1_watcher/src/factory_deps/mod.rs
+++ b/lib/l1_watcher/src/factory_deps/mod.rs
@@ -4,9 +4,12 @@
 //! This file is to be removed once the sequencer can load factory deps
 //! from L1 directly.
 
-use alloy::{hex::FromHex, primitives::{B256, hex}};
-use std::collections::HashMap;
+use alloy::{
+    hex::FromHex,
+    primitives::{B256, hex},
+};
 use serde::Deserialize;
+use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


lib/l1_watcher/src/factory_deps/mod.rs
```diff
@@ -0,0 +1,26 @@
+use alloy::{hex::FromHex, primitives::{B256, hex}};
```
popzxc:
```suggestion
//! TODO: This is a temporary solution to process factory deps for
//! upgrade transactions until we have a working bytecodes supplier
//! contract.
//! This file is to be removed once the sequencer can load factory deps
//! from L1 directly.

use alloy::{hex::FromHex, primitives::{B256, hex}};
```

popzxc (commented):
A few nits; otherwise LGTM.
Please make sure that CI passes tho (PR title, lints, etc).

PR #621, opened 2025-11-17
feat(api): implement EIP-7966 eth_sendRawTransactionSync

coffeexcoin:
Implement `eth_sendRawTransactionSync` API method to add [EIP-7966](https://eips.ethereum.org/EIPS/eip-7966) support to zksync os API.

`sendRawTransactionSync` waits for the transaction receipt up to a specified timeout (default & maximum configured in node config) 

## Summary by CodeRabbit

* **New Features**
  * Added sendRawTransactionSync RPC to submit a raw transaction and wait synchronously for a receipt; configurable timeout (default 2s).

* **Behavior / Public API Changes**
  * sendRawTransactionSync returns a receipt or a timeout error aligned with EIP‑7966.
  * Eth namespace and transaction handler now accept config and expose the new sync send path.
  * Block notifications and repository population now use a transactions lookup map keyed by tx hash; receipt repository insertion accepts iterator-based input.

* **Tests**
  * Integration tests for success and timeout paths; tester timeout increased for CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2171,6 +2171,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "basic_bootloader"
+version = "0.1.0"
+source = "git+https://github.com/matter-labs/zksync-os?tag=v0.2.2#45adae4b9cd5cea0bc39a05c0af55d5ad5dded54"
+dependencies = [
+ "arrayvec 0.7.6",
+ "basic_system 0.1.0 (git+https://github.com/matter-labs/zksync-os?tag=v0.2.2)",
+ "cfg-if",
+ "crypto 0.1.0 (git+https://github.com/matter-labs/zksync-os?tag=v0.2.2)",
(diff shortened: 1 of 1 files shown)


lib/rpc/src/tx_handler.rs
```diff
@@ -67,6 +73,60 @@ impl<Mempool: L2TransactionPool> TxHandler<Mempool> {
...
+                        tx_hash,
+                        stored_tx.receipt.clone(),
+                        &stored_tx.tx,
+                        &stored_tx.meta,
+                    ));
+                }
+            }
+        })
+        .await
+        .map_err(|_| EthSendRawTransactionSyncError::Timeout(timeout_duration))?
```
itegulov: I refactored your code a bit - added proper error types, moved logic to `TxHandler` and made so that block notifications are reused for receipts. Also added tests. LMK if you disagree with anything. If not we can merge I think.

PR #631, opened 2025-11-18
feat: support multiple SNARKers; enhance proving observability

RomanBrodetski:
## Highlights
This PR reworks Prover job management - FRI and SNARK:
* **Multiple SNARKifiers running in parallel are supported now**
* tested with ~200 real provers (>5000TPS)
* tested with 500 faulty fake provers (~1000 jobs in queue)
* extensive observability features for all operations
* got rid of string leaking in metrics
* split fri job timeout and snark job timeout

## Implementation details
`ProverJobMap` is reworked - and it now supports both FRI and SNARK. To avoid complexity tracking multi-batch jobs (as required for SNARK), **we don't explicitely retain batch->SNARK grouping information**. Instead, from `ProverJobMap` prospective we just mark all jobs in range as in_progress. This allows for uniform FRI/SNARK treatment with very little overhead.

`ProverJobMap` is now backed by a `Mutex<BTreeMap<batch_number, prover_input>>`.  I considered alternatives with parallel data structures (`DashMap` + additional cahces/atomics) but it was never worth the complexity. So, we only process queue requests sequentially. The total number of parallel jobs is bound (both naturally and with `max_assigned_batch_range` config), so the contention will not escalate.
I performed extensive testing with real and fake provers to ensure that. Even under extreme conditions (30% of all provers are timing out - resulting in some jobs taking 10+ attempts and >1000 jobs in queue with >500 parallel provers) there wasn't any observable contention. There are metrics and logs for this.
If this ever becomes a proble, we can augment the solution with a pointer to the oldest assigned and unassigned jobs - to further improve performance.

Now that we support multiple SNARKers, it's possible SNARK jobs are submitted out of order. So I had to add `GaplessL1ProofSender` - similar to `GaplessCommitter`. 

## Follow ups
With this approach to queuing we can implement job querying by VK (like @EmilLuta proposed initially). It's not done in this PR.

## Summary by CodeRabbit

* **New Features**
  * Separate configurable timeouts for FRI and SNARK prover jobs; simulated prover timeout frequency for testing
  * Gapless L1 proof sender to ensure ordered, gap-free proof batching
  * Generalized concurrent job map enabling multi‑prover scheduling and per‑job lifecycle APIs

* **Improvements**
  * Richer job-queue observability and new metrics (batch ranges, lock timing, job statistics)
  * Stronger backpressure controls and reduced buffering to better align queueing with downstream capacity
  * Added batch-range helpers on L1 send commands for easier introspection

### DIFF
--- a/node/bin/src/prover_api/metrics.rs
+++ b/node/bin/src/prover_api/metrics.rs
@@ -11,9 +11,16 @@ pub struct ProverMetrics {
     #[metrics(labels = ["stage"])]
     pub prover_job_map_max_batch_number: LabeledFamily<ProverStage, Gauge>,
     #[metrics(labels = ["stage"])]
+    /// Total number of batches in ProverMap.
+    /// There may be gaps - so prover_job_map_max_batch_number - prover_job_map_min_batch_number
+    /// doesn't always equal to .batch_count()
     pub batch_count: LabeledFamily<ProverStage, Gauge>,
+    /// The time passed between when a job was picked and reported back
     #[metrics(unit = Unit::Seconds, labels = ["stage", "type", "id"], buckets = Buckets::LATENCIES)]
     pub prove_time: LabeledFamily<(ProverStage, ProverType, String), Histogram<Duration>, 3>,
+    /// The time passed between when a job was picked and reported back
+    /// divided by the number of transactions in job.
+    /// That is, for SNARKs it's divided by the total number of txs in batch range.
     #[metrics(unit = Unit::Seconds, labels = ["stage", "type", "id"], buckets = Buckets::LATENCIES)]
     pub prove_time_per_tx: LabeledFamily<(ProverStage, ProverType, String), Histogram<Duration>, 3>,
     #[metrics(labels = ["stage", "type"], buckets = Buckets::values(&[1.0, 2.0, 3.0, 4.0, 5.0, 10.0, 20.0, 50.0]))]


lib/pipeline/src/peekable_receiver.rs
```diff
@@ -107,87 +107,6 @@ impl<T> PeekableReceiver<T> {
...
-                    }
-                }
-                Err(TryRecvError::Empty) => break,
-                Err(TryRecvError::Disconnected) => break,
-            }
-        }
-
-        out
-    }
-
```
RomanBrodetski: no more need for complex peeking!

node/bin/src/prover_api/prover_job_map/map.rs
```diff
@@ -0,0 +1,444 @@
...
+///
+/// This works both for FRI and SNARK jobs by allowing to pick multiple jobs atomically.
+/// We don't maintain the SNARK job grouping - so that on timeout, a wider range can be assigned.
+///
+#[derive(Debug)]
+pub struct ProverJobMap<T> {
+    // == state ==
+    jobs: Mutex<BTreeMap<u64, JobEntry<T>>>,
+    // Notification for waiting when batch range limit is hit (`max_assigned_batch_range`)
+    space_available: Notify,
```
RomanBrodetski: While it works, I don't exactly love this - suggestions are welcome!

node/bin/src/config.rs
```diff
@@ -463,6 +467,13 @@ pub struct FakeFriProversConfig {
     /// This gives real provers a head start when picking jobs
     #[config(default_t = Duration::from_millis(3000))]
     pub min_age: Duration,
+
+    /// Probability (0.0 to 1.0) that a job will timeout/be dropped instead of submitting a proof.
+    /// 0.0 means never timeout (default behavior).
+    /// For example, 0.1 means 10% of jobs will be dropped.
+    /// Used to test queuing behavior on timeout.
+    #[config(default_t = 0.0)]
+    pub timeout_frequency: f64,
```
EmilLuta: This looks like "testing" functionality. Would it make sense to just bake it inside of code, instead of exposing it via configs?

EmilLuta: To play devil's advocate, the entire fake proving setup is testing only, so if we're willing to commit to that, we may as well commit to this.

lib/pipeline/src/peekable_receiver.rs
EmilLuta: This is gorgeous. Thanks!

node/bin/src/prover_api/gapless_l1_prove_sender.rs
```diff
@@ -0,0 +1,66 @@
+use async_trait::async_trait;
```
EmilLuta: Metacomment: Does this component really belong to prover_api? I'd assume it might be more suitable for L1 interactions (maybe sender?)

node/bin/src/prover_api/metrics.rs
```diff
@@ -1,14 +1,28 @@
 use std::time::Duration;
-use vise::{Buckets, EncodeLabelValue, Histogram, LabeledFamily, Metrics, Unit};
+use vise::{Buckets, EncodeLabelValue, Gauge, Histogram, LabeledFamily, Metrics, Unit};
 
 #[derive(Debug, Metrics)]
 #[metrics(prefix = "prover")]
 pub struct ProverMetrics {
+    #[metrics(labels = ["stage"])]
```
EmilLuta: Adding docstrings to all metrics would help with following up how they're used. (i.e. what's `prover_job_min_batch_number`? Minimum number it's waiting for? Minimum number it has so far?)

node/bin/src/prover_api/fri_proving_pipeline_step.rs
```diff
@@ -68,16 +63,29 @@ impl PipelineComponent for FriProvingPipelineStep {
...
-                    let _ = self.batches_for_prove_sender.send(batch).await;
+                    tracing::info!(
+                        "Received batch for FRI proving: {:?}",
+                        batch.batch_number()
+                    );
+                    // Add job directly to FriJobManager - this will await if queue is full
+                    self.fri_job_manager.add_job(batch).await
                 }
-            } => anyhow::bail!("FRI proving input stream ended unexpectedly"),
+                Ok::<(), anyhow::Error>(())
```
EmilLuta: Interesting, isn't `Ok(())` enough here?

node/bin/src/prover_api/fri_job_manager.rs
```diff
@@ -75,168 +61,85 @@ pub struct FriJob {
...
     pub fri_job: FriJob,
-    pub assigned_seconds_ago: u64,
-}
-
-// TODO: remove, once legacy is deprecated
-#[derive(Debug, Serialize)]
-pub struct JobStateLegacy {
-    pub batch_number: u64,
-    pub assigned_seconds_ago: u64,
-}
```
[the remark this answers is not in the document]
EmilLuta: Yes, but it breaks backwards compatibility with old API. I'd suggest we add it back and remove it when we remove legacy altogether (as you suggested, can be a follow-up PR after this; provers should be updated everywhere by now).

node/bin/src/prover_api/fri_job_manager.rs
```diff
@@ -273,15 +187,70 @@ impl FriJobManager {
...
+
+        // Remove the job from the assigned map.
+        let Some(removed_job) = self
+            .jobs
+            .complete_job(batch_number, ProverType::Real, prover_id)
+            .await
+        else {
+            // If already removed due to a race
+            // (another submit won), we still return success to keep the API idempotent.
+            tracing::warn!(
```
EmilLuta: Nit: This will cause a sentry "alert". What exactly would we do in such scenario?

node/bin/src/prover_api/snark_job_manager.rs
```diff
@@ -10,48 +12,26 @@ use zksync_os_l1_sender::commands::prove::ProofCommand;
...
 };
-use zksync_os_pipeline::PeekableReceiver;
 use zksync_os_types::ProvingVersion;
 
-use crate::prover_api::fri_job_manager::FriJob;
-
 /// Job manager for SNARK proving.
 ///
-/// Doesn't support multiple provers yet (they'd get the same job)
+/// Support orchestrating multiple SNARK provers
```
EmilLuta:
```suggestion
/// Supports multiple SNARK provers
```

node/bin/src/prover_api/snark_job_manager.rs
```diff
@@ -60,73 +40,56 @@ pub struct SnarkJobManager {
...
+    pub async fn pick_real_job(
+        &self,
+        prover_id: String,
+    ) -> anyhow::Result<Option<Vec<(FriJob, FriProof)>>> {
+        // consume/remove all fake jobs that may be in the front of the queue
+        self.process_pending_fake_fri_proofs().await?;
+
+        let batches_with_real_proofs = self
+            .jobs
+            .pick_jobs_while(self.max_fris_per_snark, &prover_id, |job| {
```
EmilLuta: We need to make sure that these batches have the same VK/proving_execution_version or otherwise we cannot seal such batch. I believe this is already done at batcher level, but I'd really love to make sure we double check here (i.e. provers can't merge/SNARKify 2 proofs with different VKs).

node/bin/src/prover_api/snark_job_manager.rs
```diff
@@ -243,43 +164,57 @@ impl SnarkJobManager {
         Ok(())
     }
 
-    /// Consumes fake FRI proves from HEAD and turns them into fake SNARKs
-    /// Additionally, if `consume_by_timeout` is Some,
-    ///    also consumes real FRI proves that are older than `consume_by_timeout`
-    async fn consume_fake_proves_from_head(
+    /// Consumes fake FRI proofs from the head of the queue and turns them into fake SNARKs.
+    async fn process_pending_fake_fri_proofs(&self) -> anyhow::Result<()> {
```
EmilLuta: Will this work as intended? We call it in front of the `pick_real_jobs` which means it will consume all the fake proofs (as intended) but also timed out proofs (those would be the ones we sent and didn't receive an answer for, right?).

node/bin/src/prover_api/gapless_l1_prove_sender.rs
```diff
@@ -0,0 +1,66 @@
...
+                    // Flush ready commands
+                    while let Some(next_command) = buffer.remove(&next_expected_batch_number) {
+                        next_expected_batch_number += next_command.batch_count() as u64;
+                        latency_tracker.enter_state(GenericComponentState::WaitingSend);
+                        output.send(next_command).await?;
+                        latency_tracker.enter_state(GenericComponentState::Processing);
+                    }
+                }
+                None => {
+                    anyhow::bail!("GaplessCommitterPassthrough input stream ended unexpectedly");
```
EmilLuta: Typo? I assume this should be `GaplessL1ProofSender`. (also, filename & struct seem to be different)

node/bin/src/config.rs
EmilLuta: On a separate note, should we validate that this is between 0.0 and 1.0 anywhere?

node/bin/src/config.rs
```diff
@@ -423,9 +423,13 @@ pub struct ProverApiConfig {
     /// however, we won't turn real FRI proofs into fake ones - even on timeout.
     pub fake_snark_provers: FakeSnarkProversConfig,
 
-    /// Timeout after which a prover job is assigned to another Fri Prover Worker.
+    /// Timeout after which a FRI prover job is assigned to another Fri Prover Worker.
+    #[config(alias = "job_timeout", default_t = Duration::from_secs(300))]
+    pub fri_job_timeout: Duration,
```
EmilLuta: Should we make this number more sensible? I assume 60s is about enough, WDYT?

node/bin/src/prover_api/fri_proving_pipeline_step.rs
```diff
@@ -68,16 +63,29 @@ impl PipelineComponent for FriProvingPipelineStep {
...
+            } => {
+                result?;
+                anyhow::bail!("FRI proving input stream ended unexpectedly")
+            },
             _ = async {
```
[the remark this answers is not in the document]
RomanBrodetski: it's just to forward results from JobManager downstream. Due to Pipeline Framework design, it's cumbersome to do directly.

node/bin/src/prover_api/fri_proving_pipeline_step.rs
EmilLuta: Indeed. I reckon it's `tokio::select!` that doesn't know the type.

node/bin/src/prover_api/snark_job_manager.rs
EmilLuta: No, the comment is enough. The order of reading messed me up. My miss, don't mind me.

node/bin/src/prover_api/prover_job_map/map.rs
```diff
@@ -0,0 +1,444 @@
...
+        let mut jobs = self.lock_with_tracking(JobMapMethod::PickJobsWhile).await;
+
+        let mut selected_jobs = Vec::new();
+        for (_, entry) in jobs.iter_mut() {
+            if !self.is_job_eligible(&selected_jobs, entry, now, limit, &mut predicate) {
+                if selected_jobs.is_empty() {
+                    // We didn't find any jobs yet - continue looking for the first eligible one
+                    continue;
+                } else {
+                    // We already have some jobs - cannot add more jobs, otherwise we'd have a gap
```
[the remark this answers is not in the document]
RomanBrodetski: > non-fake

prover_job_map knows nothing about fake/non-fake.

> (but this also assumes we'll keep using a BTreeMap or ordered DS going forward).

Sure, it must be ordered - otherwise this approach doesn't work

node/bin/src/config.rs
RomanBrodetski: similar time

node/bin/src/prover_api/gapless_l1_proof_sender.rs
```diff
@@ -0,0 +1,66 @@
...
+        self,
+        mut input: PeekableReceiver<Self::Input>,
+        output: mpsc::Sender<Self::Output>,
+    ) -> anyhow::Result<()> {
+        let latency_tracker = ComponentStateReporter::global().handle_for(
+            "gapless_l1_proof_sender",
+            GenericComponentState::WaitingRecv,
+        );
+
+        let mut buffer: BTreeMap<u64, L1SenderCommand<ProofCommand>> = BTreeMap::new();
```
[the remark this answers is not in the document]
RomanBrodetski: thing is, we have a similar situation in `gapless_committer`. We need a way - not based on backpressure - to limit block/batch production when there are delays.  Here is a [ticket](https://github.com/orgs/matter-labs/projects/23/views/3?pane=issue&itemId=141615962&issue=matter-labs%7Czksync-os-server%7C670).

We may want to introduce some limiting - but I think in a uniform way for gapless_committer and gapless_l1_proof_semder - but not in this PR (and it doesn't introduce new issues - gapless_committer is already unbounded).


Also, to fill the whole RAM it will take days if not years at moderate load (and for very high load we don't have GPU anyway)

EmilLuta (approved):
There're still a few nits that I left. Feel free to resolve & merge the PR if there's nothing else you want to fix as part of this PR. LGTM, from my PoV.

node/bin/src/prover_api/prover_job_map/map.rs
```diff
@@ -0,0 +1,744 @@
...
+    use alloy::primitives::{Address, B256};
+    use std::time::Duration;
+    use zksync_os_contract_interface::models::{
+        CommitBatchInfo, DACommitmentScheme, StoredBatchInfo,
+    };
+    use zksync_os_l1_sender::batcher_model::{BatchForSigning, BatchMetadata};
+    use zksync_os_l1_sender::commitment::BatchInfo;
+    use zksync_os_types::{ProtocolSemanticVersion, PubdataMode};
+
+    fn create_test_batch_envelope(batch_number: u64) -> SignedBatchEnvelope<Vec<u8>> {
```
EmilLuta: Nit: Maybe I'd implement this as a method on the struct?

EmilLuta (approved):

PR #657, opened 2025-11-25
feat!: allow EN to sync with overriden records

perekopskiy:
## Summary

In case blocks are rebuilt main node overwrites replay record but keeps old records under some db key that are logged (see https://github.com/matter-labs/zksync-os-server/pull/620).
Usually, such blocks have to be debugged later. This PR adds config that allows EN to sync the uncanonical chain of blocks.
Config example:
```
sequencer_en_sync_up_to_block=2 \
sequencer_en_replay_record_overrides__JSON="[[2, \"0x00000000000000020000000069259e82\"]]" \
```

## Breaking Changes

Breaking change for EN<->main node communication.

## Rollout Instructions

Update main nodes and EN at the same time.

## Summary by CodeRabbit

* **New Features**
  * External nodes can halt synchronization at a configured block.
  * Per-block replay record overrides can be provided via configuration and HTTP query parameters.

* **Configuration**
  * Added en_sync_up_to_block and en_replay_record_overrides options for external node setup.

* **Behavior**
  * Replay streaming and replay retrieval now honor per-block override values during processing.

### DIFF
--- a/node/bin/src/command_source.rs
+++ b/node/bin/src/command_source.rs
@@ -5,7 +5,7 @@ use futures::StreamExt;
 use futures::stream::BoxStream;
 use std::collections::HashSet;
 use std::time::Duration;
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, watch};
 use zksync_os_pipeline::{PeekableReceiver, PipelineComponent};
 use zksync_os_sequencer::model::blocks::{BlockCommand, ProduceCommand, RebuildCommand};
 use zksync_os_storage_api::{ReadReplay, ReadReplayExt};
@@ -33,6 +33,7 @@ pub struct ExternalNodeCommandSource {
     pub record_overrides: Vec<(u64, Bytes)>,
     pub up_to_block: Option<u64>,
     pub replay_download_address: String,
+    pub stop_receiver: watch::Receiver<bool>,
 }
 
 #[async_trait]
@@ -78,7 +79,7 @@ impl PipelineComponent for ExternalNodeCommandSource {
     const OUTPUT_BUFFER_SIZE: usize = 5;
 
     async fn run(
-        self,
+        mut self,
         _input: PeekableReceiver<()>,
         output: mpsc::Sender<BlockCommand>,
     ) -> anyhow::Result<()> {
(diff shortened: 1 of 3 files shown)


node/bin/src/replay_transport.rs
```diff
@@ -27,9 +28,29 @@ pub async fn replay_server(
...
-            skip_http_headers(&mut reader)
+            let skipped_bytes = skip_http_headers(&mut reader)
                 .await
                 .expect("failed to skip HTTP headers");
+            let block_replay_query = if let Ok(skipped_text) = String::from_utf8(skipped_bytes) {
+                if let Some(url_str) = skipped_text
+                    .split_whitespace()
+                    .find(|s| s.starts_with("/block_replays"))
+                {
+                    if let Ok(url) = Url::parse(&format!("http://dummy{url_str}")) {
```
itegulov: We are slowly starting to implement an HTTP protocol parser at this point... I know this is not directly related to your PR but any idea where we are on being able to do raw TCP from the infra standpoint?

node/bin/src/replay_transport.rs
```diff
@@ -27,9 +28,29 @@ pub async fn replay_server(
...
+                        let pairs: Vec<(String, String)> = url.query_pairs().into_owned().collect();
+                        BlockReplayQuery::parse(pairs)
+                    } else {
+                        tracing::info!("Could not parse URL from HTTP headers");
+                        BlockReplayQuery::default()
```
itegulov: Same here

node/bin/src/command_source.rs
```diff
@@ -80,15 +83,31 @@ impl PipelineComponent for ExternalNodeCommandSource {
...
+
+            if let Some(up_to_block) = self.up_to_block
+                && command.block_number() > up_to_block
+            {
+                tracing::info!(
+                    up_to_block,
+                    "Reached up_to_block, halting external command source"
+                );
+                // Sleeping indefinitely.
+                futures::future::pending::<()>().await;
```
itegulov: Technically this should wait for `stop_receiver` but we have this problem in a lot of places. If it's difficult to get it from here feel free to ignore (but I'd at least leave a comment here)

itegulov (commented):
I'd heavily prefer not relying on HTTP query params but conceptually logic LGTM

lib/socket/src/lib.rs
```diff
@@ -70,10 +71,11 @@ pub async fn skip_http_headers<R: AsyncBufRead + Unpin>(
         }
 
         for (i, &byte) in buf.iter().enumerate() {
+            skipped.push(byte);
             if byte == b'\n' {
                 if empty_line {
                     reader.consume(i + 1);
-                    return Ok(());
+                    return Ok(skipped);
```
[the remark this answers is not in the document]
perekopskiy: Yes, but I'd still keep them

PR #659, opened 2025-11-25
fix: Sequencer requesting proofs for commited batches

Artemka374:
## Summary

Sometimes when node is restarted, sequencer requests FRI proofs for batches that were already commited on L1. This is not the behaviour we want to have, since it produces problems during the protocol upgrades(when we request proofs for batches with old version from provers with new version). This PR changes the behaviour to verify which batches were already commited(against L1) and don't send them down the FRI proving pipeline.

## Summary by CodeRabbit

* **Bug Fixes**
  * The proving pipeline now detects and skips batches already proved at startup, preventing redundant reprocessing and unnecessary work; such batches are forwarded downstream as already-proved.

* **Chores**
  * Startup now passes the last-proved batch into the proving pipeline so the gating behavior is applied on launch.

### DIFF
--- a/node/bin/src/prover_api/fri_proving_pipeline_step.rs
+++ b/node/bin/src/prover_api/fri_proving_pipeline_step.rs
@@ -80,7 +80,7 @@ impl PipelineComponent for FriProvingPipelineStep {
                         self.fri_job_manager.add_job(batch).await
                     } else {
                         // Already proven - send with fake proof to pass through the pipeline
-                        let batch_with_fake_proof = batch.with_data(FriProof::Fake);
+                        let batch_with_fake_proof = batch.with_data(FriProof::AlreadySubmittedToL1);
                         let _ = output.send(batch_with_fake_proof).await;
                     }
                 }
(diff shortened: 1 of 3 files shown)


node/bin/src/prover_api/fri_proving_pipeline_step.rs
```diff
@@ -73,7 +76,12 @@ impl PipelineComponent for FriProvingPipelineStep {
         tokio::select! {
             _ = async {
                 while let Some(batch) = input.recv().await {
-                    let _ = self.batches_for_prove_sender.send(batch).await;
+                    if batch.batch_number() > self.last_commited_batch_number {
+                        let _ = self.batches_for_prove_sender.send(batch).await;
+                     }
+                     // else {
+                    //     let _ = output.send(L1SenderCommand::Passthrough(Box::new(batch))).await;
+                    // }
```
RomanBrodetski: is this in its final state?

RomanBrodetski (changes requested):

PR #662, opened 2025-11-25
fix: 2FA followup

cytadela8:
## Summary

- fix attempting to sign commited batches
- metrics
- comments 

## TODO

- [ ] Finish integration test
- [ ] Integration test with restart
- [ ] Write more in docs? 

## Release Notes

* **New Features**
  * Expanded batch verification monitoring with new metrics including cache range tracking, request success/failure tracking, per-signer performance latency analysis, and retry attempt counters
  * Optimized verification pipeline to skip unnecessary re-processing of already-committed batches, improving overall system efficiency
  * Enhanced observability and debugging with address-aware client logging for verification operations

### DIFF
--- a/lib/batch_verification/src/client/block_cache.rs
+++ b/lib/batch_verification/src/client/block_cache.rs
@@ -13,6 +13,7 @@ use super::metrics::BATCH_VERIFICATION_CLIENT_METRICS;
 /// This may be optimized by using a ring buffer for data storage instead.
 pub(super) struct BlockCache<Finality> {
     data: HashMap<u64, (BlockOutput, ReplayRecord, BlockMerkleTreeData)>,
+    /// Range of cached data. Range is inclusive of both bounds.
     range: Option<(u64, u64)>,
     finality: Finality,
 }
@@ -67,7 +68,7 @@ impl<Finality: ReadFinality> BlockCache<Finality> {
             for num in low..block_number {
                 self.data.remove(&num);
             }
-            let new_range = (block_number + 1, high);
+            let new_range = (block_number, high);
 
             if new_range.0 > new_range.1 {
                 self.range = None;


lib/batch_verification/src/client/metrics.rs
```diff
@@ -1,11 +1,48 @@
...
+use vise::{Counter, Gauge, Metrics};
 
 #[derive(Debug, Metrics)]
 #[metrics(prefix = "batch_verification_client")]
 pub struct BatchVerificationClientMetrics {
-    pub block_cache_size: Gauge<usize>,
+    /// Number of blocks currently cached
+    block_cache_size: Gauge<usize>,
+    /// Lowest block height in the cache
+    block_cache_from_height: Gauge<u64>,
```
RomanBrodetski: nit: while valid, we never use term"block_height" we always say "block_number". Please change for consistency

lib/batch_verification/src/sequencer/component.rs
```diff
@@ -201,13 +220,34 @@ impl BatchVerifier {
...
+            // We skip signing batches that were already committed. This happens on startup
+            if batch_envelope.batch_number() <= self.last_committed_batch_number {
+                tracing::info!(
+                    "Skipping signing of already committed batch {}",
+                    batch_envelope.batch_number()
+                );
+                singed_batcher_sender
+                    .send(
+                        batch_envelope
+                            .with_signatures(BatchSignatureData::NotNeeded)
```
RomanBrodetski: Hmmm not sure it's right to use `NotNeeded` here. The comment there says that it's for backwards compatibility - but now we use it for something else. I guess a quick fix is to add one more enum variant

RomanBrodetski (approved):
looks very good - please note my comments but feel free to do so in a follow up

lib/batch_verification/src/client/metrics.rs
cytadela8: Will do in followup.

PR #664, opened 2025-11-26
feat(l1-sender): send EIP-7594 blobs when Fusaka is activated

### LINKED ISSUE #663: Use Fusaka blob transaction format
We should use the new 4844 blob format starting from Fusaka upgrade. Sepolia is already upgraded, mainnet is scheduled on 3rd of December.

itegulov:
## Summary

Fixes #663 

Adds conditional EIP-4844 -> EIP-7594 blob conversion logic that gets run when `l1_sender_fusaka_upgrade_timestamp` is passed on L1. Effectively disabled by default (set to `2^64-1`).

Unfortunately I had to propagate `FillProvider` down the chain as alloy does not give a nice way to use 7594 blobs otherwise. Even the official [example](https://alloy.rs/examples/transactions/send_eip7594_transaction/) depends on provider to be `FillProvider`.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable upgrade timestamp to control protocol upgrade behavior.

* **Refactor**
  * Reworked transaction sending pipeline to use a layered provider and transaction filler.
  * Prepared transactions are now encoded and sent based on pending-block conditions, enabling adaptive encoding for different protocol eras.

### DIFF
--- a/lib/l1_sender/src/lib.rs
+++ b/lib/l1_sender/src/lib.rs
@@ -160,6 +160,9 @@ pub async fn run_l1_sender<Input: SendToL1>(
                     let envelope = provider.fill(tx_request).await?.try_into_envelope()?.try_into_pooled()?;
 
                     let pending_block = provider.get_block(BlockId::pending()).await?.expect("no pending block");
+                    // todo: make conversion unconditional (and remove respective config) once both:
+                    //       1) Fusaka upgrade is executed on mainnet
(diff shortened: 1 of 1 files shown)


perekopskiy (dismissed):
Looks good! Maybe add a todo + issue for removing config var and eip4844 logic after both mainnet and anvil are updated?

Issue #510, opened 2025-10-13
get rid of proof storage


RomanBrodetski:
* don't store FRI
* use L1 for block->batch mapping

PR #672, opened 2025-11-26
feat(batcher): re-create batches using L1 watcher's data

itegulov:
## Summary

Related to #510 

Follow-up to #661 that makes batcher rely on data reported by `BatchRangesWatcher` (to be renamed) for batch recreation. New invariant is that `Batcher` always starts batching from `last_executed_batch + 1` and can skip some blocks at the start that have already been executed. Note that this PR almost entirely removes usage of S3 during state recovery.

## Summary by CodeRabbit

* **New Features**
  * Exposed on-chain upgrade info (upgrade transaction hash and upgrade batch number) for read-only queries.
  * Added a committed-batch stream so the batcher can reconstruct committed batches and support genesis startup.

* **Improvements**
  * More reliable node startup using genesis-based starting-block logic.
  * Watcher and batcher processing hardened: inclusive batch handling, sequential validation, improved logging, and cleaner shutdown/flow control.

### DIFF
--- a/node/bin/src/batcher/mod.rs
+++ b/node/bin/src/batcher/mod.rs
@@ -269,7 +269,6 @@ impl Batcher {
                             // arm the timer after we process the block number that's more or equal
                             // than last persisted one - we don't want to seal on timeout if we know that there are still pending blocks in the inbound channel
                             if deadline.is_none() {
-                                // todo: this logic is not needed anymore?
                                 if block_number >= self.startup_config.last_persisted_block {
                                     deadline = Some(Box::pin(tokio::time::sleep(self.batcher_config.batch_timeout)));
                                 } else {


lib/l1_watcher/src/batch_range_watcher.rs
```diff
@@ -1,25 +1,29 @@
...
 
-/// Discovers block ranges for batches `[last_executed_batch + 1; last_committed_batch]`. This is
+/// Discovers commitment data for batches `[last_executed_batch; last_committed_batch]`. This is
 /// needed to rebuild batches correctly in Batcher during replay.
 pub struct BatchRangeWatcher {
-    contract_address: Address,
-    provider: DynProvider,
+    zk_chain: ZkChain<DynProvider>,
     next_batch_number: u64,
     last_batch_number: u64,
```
RomanBrodetski: maybe `last_committed_batch_on_startup`?  Please add detailed comment for both next/last_batch_number - now it's not obvious what they are and how/if/when they change.

node/bin/src/batcher/mod.rs
```diff
@@ -78,7 +79,56 @@ impl PipelineComponent for Batcher {
...
+
+                (
+                    last_executed_batch_info.into_stored(&last_executed_batch.protocol_version),
+                    last_executed_block_number + 1,
+                )
+            }
+        };
+
+        // We might receive some blocks that belong to already executed batches. We can skip these
+        // as there is no need to perform any L1 operations on them.
```
RomanBrodetski: hmm this makes it impossible to re-prove an already executed batch - but maybe indeed it's not such a important operations tool..

PR #718, opened 2025-12-11
fix(tracer): Fix call tracer behavior for 'empty' transactions

0xVolosnikov:
## Summary

Currently our call tracer can panic in some edge cases when tx fails _before_ any EVM frame is created. We have to populate the trace somehow, otherwise it may break data consumers assumptions


0xVolosnikov:
Ok, seems like we can't do a quick fix here and have to make it properly right away.

PTAL

PR #726, opened 2025-12-16
feat: 2FA L1 integration

cytadela8:
## Summary

Use signatures collected through 2FA Batch Verification for submission through `commitBatchesMultisig`. Validator set and threshold are also received from L1.

Ignoring RUSTSEC-2025-0137 as we do not use the vulnerable function and there is no fixed version available.

### TODO

- [x] Pr description
- [x] Rebase
- [x] Fix test
- [x] Use const where indicated by todo comment
- [ ] consider separating signatures from config and L1 signatures (so we might have an extra L2 mechanism ?)

### DIFF
--- a/lib/batch_verification/src/wire_format/conversion.rs
+++ b/lib/batch_verification/src/wire_format/conversion.rs
@@ -55,7 +55,7 @@ impl From<BatchVerificationRequest> for BatchVerificationRequestWireFormatV1 {
         } = value;
         let commit_data_alloy = CommitBatchInfoZKsyncOS::from(commit_data);
         let encoded_commit_data = commit_data_alloy.abi_encode();
-        // last_commit_timestamp is used in transport, but it is fine, because it is unused in L1
+        // StoredBatchInfo conversion is not lossless last_commit_timestamp is zeroed. It is fine, because it is a legacy field unused in L1.
         let prev_commit_data_alloy = IExecutor::StoredBatchInfo::from(&prev_commit_data);
         let encoded_prev_commit_data = prev_commit_data_alloy.abi_encode();
         Self {


itegulov (commented):
Please add PR description

lib/batch_types/src/batch_signature.rs
```diff
@@ -72,9 +112,58 @@ impl BatchSignature {
...
+    let batch_data_hash = keccak256(batch_data);
+
+    let encoded = (
+        typehash,
+        batch_info.chain_address,
+        U256::from(batch_info.batch_number), // processBatchFrom
+        U256::from(batch_info.batch_number), // processBatchTo
+        batch_data_hash,
+    )
+        .abi_encode_params();
```
itegulov: I'd use `alloy::sol!` to generate a struct for `CommitBatchesMultisig` and then construct it properly. Having to care about type hash and constructing a tuple manually feels too low-level

lib/batch_types/src/batch_signature.rs
```diff
@@ -72,9 +112,58 @@ impl BatchSignature {
...
+        name: "MultisigCommitter",
+        version: "1",
+        chain_id: l1_chain_id,
+        verifying_contract: multisig_committer,
+    };
+    let domain_separator = domain.separator();
+
+    keccak256(
+        [
+            &[0x19, 0x01],
```
itegulov: What is this magic value?

lib/batch_verification/src/client/mod.rs
```diff
@@ -61,19 +63,31 @@ type VerificationInput = (
...
     ) -> Self {
+        let signer = PrivateKeySigner::from_str(private_key.expose_secret())
+            .expect("Invalid batch verification private key");
+        if let BatchVerificationL1::Enabled(l1_config) = l1_state.batch_verification.clone()
+            && !l1_config.validators.contains(&signer.address())
+        {
+            tracing::warn!(
+                "Your address {} is not authorized to verify batches on L1",
+                signer.address()
+            );
```
itegulov: Nit: please use structured logging

lib/batch_verification/src/sequencer/component.rs
```diff
@@ -30,20 +30,53 @@ fn report_exit<T, E: std::fmt::Debug>(name: &'static str) -> impl Fn(Result<T, E
...
+        last_committed_batch_number: u64,
+    ) -> Self {
+        let config_validators = config
+            .accepted_signers
+            .clone()
+            .into_iter()
+            .map(|s| s.parse().unwrap())
+            .collect();
+        // If on L1 batch verifiers re configured, we use that configuration instead
+        let (threshold, validators) = match l1_state.batch_verification.clone() {
```
itegulov: Nit: why clone here?

lib/batch_verification/src/wire_format/v1.rs
```diff
@@ -16,6 +16,7 @@ pub struct BatchVerificationRequestWireFormatV1 {
     pub pubdata_mode: u8,
     pub request_id: u64,
     pub commit_data: Vec<u8>,
+    pub prev_commit_data: Vec<u8>,
```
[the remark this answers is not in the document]
cytadela8: I think its fine to break the transport protocol version compatibility at this point. It will only be a problem when we cannot update all ENs at the same time.

lib/batch_verification/src/wire_format/conversion.rs
```diff
@@ -42,16 +51,20 @@ impl From<BatchVerificationRequest> for BatchVerificationRequestWireFormatV1 {
             pubdata_mode,
             request_id,
             commit_data,
+            prev_commit_data,
         } = value;
         let commit_data_alloy = CommitBatchInfoZKsyncOS::from(commit_data);
         let encoded_commit_data = commit_data_alloy.abi_encode();
+        let prev_commit_data_alloy = IExecutor::StoredBatchInfo::from(&prev_commit_data);
+        let encoded_prev_commit_data = prev_commit_data_alloy.abi_encode();
```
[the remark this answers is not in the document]
itegulov: I think what's confusing me is that I don't see a field named `last_commit_timestamp` anywhere. Do you mean just regular `timestamp`? If so then there is a comment inside `from` conversion already and it is expected to be `0` indeed.

PR #730, opened 2025-12-17
feat: add sequencer sandbox mode

vasyl-ivanchuk:
## Summary

Added a `--sandbox` flag to the CLI to use a temporary directory for RocksDB files. The directory is automatically dropped on shutdown. In sandbox mode, the `general_rocks_db_path` setting is ignored and a warning is logged if it has been set.

Command examples:
```
./zksync-os-server --sandbox
```

Logs examples:
```
WARN zksync_os_server: general_rocks_db_path parameter is ignored in sandbox mode _original_path_=./db/test123
```
```
INFO zksync_os_server: Sandbox mode enabled. Using temporary RocksDB directory _path_=/var/folders/1m/djcrv53173g3gcc9s6zs92180000gn/T/.tmp8TNBXr
```

### DIFF
--- a/node/bin/src/main.rs
+++ b/node/bin/src/main.rs
@@ -318,9 +318,7 @@ fn build_external_config(repo: ConfigRepository<'_>) -> Config {
 
 fn enable_sandbox_mode(config: &mut Config) -> Option<TempDir> {
     let original_path = config.general_config.rocks_db_path.clone();
-    if original_path != Path::new(DEFAULT_ROCKS_DB_PATH)
-        || std::env::var_os("GENERAL_ROCKS_DB_PATH").is_some()
-    {
+    if original_path != Path::new(DEFAULT_ROCKS_DB_PATH) {
         tracing::warn!(
             original_path = %original_path.display(),
             "general_rocks_db_path parameter is ignored in sandbox mode"


node/bin/src/config/mod.rs
```diff
@@ -44,6 +44,8 @@ pub struct Config {
     pub batch_verification_config: BatchVerificationConfig,
 }
 
+pub const DEFAULT_ROCKS_DB_PATH: &str = "./db/node1";
```
popzxc: There exists the [config_constants](https://github.com/matter-labs/zksync-os-server/blob/main/node/bin/src/config_constants.rs) module; it probably makes sense to put this constant there. Even though it was introduced for constants that can change from time to time, I guess it still makes sense to keep things consistent.

popzxc (commented):
Generally LGTM

popzxc (dismissed):
LGTM, but I'd wait for @itegulov review as well

node/bin/src/main.rs
```diff
@@ -303,6 +316,27 @@ fn build_external_config(repo: ConfigRepository<'_>) -> Config {
     }
 }
 
+fn enable_sandbox_mode(config: &mut Config) -> Option<TempDir> {
+    let original_path = config.general_config.rocks_db_path.clone();
+    if original_path != Path::new(DEFAULT_ROCKS_DB_PATH)
+        || std::env::var_os("GENERAL_ROCKS_DB_PATH").is_some()
```
itegulov: Nit: no need to check the envvar directly, `config.general_config.rocks_db_path` gets populated from it anyway (that's how `smart-config` works)

itegulov (approved):
Just one nit

node/bin/src/main.rs
vasyl-ivanchuk: Makes sense! Removed env var check.

PR #758, opened 2026-01-06
fix: construct pending block context in `eth_call`-like methods

itegulov:
## Summary

Right now we presume that pending block is the same as latest block. This shouldn't be the case, we need to construct fresh block context based on current system state instead. This PR is limited to choosing new timestamp and updated last 256 block hashes. Fee-related fields are set to the latest block's values for now.


lib/rpc/src/eth_call_handler.rs
```diff
@@ -197,6 +198,45 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
         Ok(Recovered::new_unchecked(tx, from).into())
     }
 
+    fn pending_block_ctx(&self) -> BlockContext {
```
perekopskiy: I'd like to unify this approach and the one used in estimate_gas. estimate_gas uses `EthCallHandler::pending_block_context` which is a watch that gets populated in context provider each time context is created, basically it refers either to "pending" or "latest" (if block is already fully processed). I think we can look at `EthCallHandler::pending_block_context`, if it refers to "pending" then just use it, otherwise (it's "latest" or none) use your approach that takes latest context and changes timestamp and block hashes

itegulov: Ah, I didn't even notice that logic was added at some point. I'll refactor, makes total sense

perekopskiy (approved):

Issue #708, opened 2025-12-08
Evaluate devp2p and feasibility of migration


itegulov:
We want to refactor our networking stack to be robust, secure and extendable. Most promising option (and the industry standard) is devp2p that we want to evaluate and potentially reuse in our system. Links:
* https://github.com/ethereum/devp2p
* https://github.com/ethereum/devp2p/blob/master/rlpx.md#capability-messaging - optional capabilites that can drive sidecar features such as 2FA

PR #773, opened 2026-01-08
feat(network): add runnable `NetworkService` (disabled by default)

itegulov:
## Summary

Second part of #708. Adds runnable `NetworkService` that is disabled by default. NAT discovery is disabled for now, at least one bootstrap node is required for now.


lib/network/src/config.rs
```diff
@@ -0,0 +1,15 @@
+use reth_network::config::SecretKey;
+use reth_network_peers::NodeRecord;
+
+pub struct NetworkConfig {
```
popzxc: `#[derive(Debug)]`?

lib/network/src/service.rs
```diff
@@ -0,0 +1,111 @@
...
+        config: NetworkConfig,
+        is_main_node: bool,
+        replay: impl ReadReplay + Clone,
+        zk_provider_factory: ZkProviderFactory<
+            impl ReadStateHistory + Clone,
+            impl ReadRepository + Clone,
+        >,
+        replay_sender: mpsc::UnboundedSender<ReplayRecord>,
+    ) -> Result<Self, NetworkError> {
+        // todo: do not use localhost
```
popzxc: Is it expected to be changed immediately after this PR? Does it make sense to create/reference an issue for that?

lib/network/src/service.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    /// Consume the service by registering it as an endless task that drives the network state.
+    pub fn run(mut self, tasks: &mut JoinSet<()>, stop_receiver: watch::Receiver<bool>) {
+        tasks.spawn(self.network_manager);
+        tasks.spawn(async move {
+            while !*stop_receiver.borrow() {
+                let Some(event) = self.protocol_rx.recv().await else {
+                    break;
+                };
+                tracing::trace!(?event, "received zks protocol event");
```
popzxc: Maybe comment on why we don't need to do anything with events and/or TODO on the plans to process them?

node/bin/src/config/mod.rs
```diff
@@ -204,6 +210,33 @@ pub struct GeneralConfig {
...
+#[derive(Clone, Debug, DescribeConfig, DeserializeConfig)]
+#[config(derive(Default))]
+pub struct NetworkConfig {
+    /// Whether devp2p-based networking should be enabled.
+    #[config(default_t = false)]
+    pub enabled: bool,
+    /// The node's secret key, from which the node's identity is derived. Used during initial RLPx
+    /// handshake.
+    #[config(default_t = SecretString::from("21b0ee131240821c39627c39d0fdde5edbda968c5877f5b63c5c542f267b5349"))]
+    pub secret_key: SecretString,
```
slowli: Is having a default value for this field temporary? I don't think it's a particularly good idea. If we need to have a zero-configuration runnable node, maybe it would make sense to make the field optional and create a secret key randomly if it's not set?

node/bin/src/config/mod.rs
```diff
@@ -204,6 +210,33 @@ pub struct GeneralConfig {
...
+    pub port: u16,
+    /// All boot nodes to start network discovery with. Expected format is
+    /// `enode://<node ID>@<IP address>:<port>`.
+    #[config(
+        default_t = vec![
+            "enode://dbd18888f17bad7df7fa958b57f4993f47312ba5364508fd0d9027e62ea17a037ca6985d6b0969c4341f1d4f8763a802785961989d07b1fb5373ced9d43969f6@127.0.0.1:3060".into()
+        ],
+        with = Delimited(",")
+    )]
+    pub boot_nodes: Vec<String>,
```
slowli: Nit: It would be more idiomatic to have `Vec<NodeRecord>` here (via a custom deserializer), so that there's meaningful error handling etc.

node/bin/src/config/mod.rs
itegulov: I would agree but it is needed for `boot_nodes` below. Public key there is derived from `secret_key`, I added a clarifying comment.

node/bin/src/config/de.rs
```diff
@@ -0,0 +1,27 @@
...
+impl DeserializeParam<Vec<NodeRecord>> for NodeRecordVec {
+    const EXPECTING: BasicTypes = BasicTypes::STRING;
+
+    fn deserialize_param(
+        &self,
+        ctx: DeserializeContext<'_>,
+        param: &'static ParamMetadata,
+    ) -> Result<Vec<NodeRecord>, ErrorWithOrigin> {
+        let de = ctx.current_value_deserializer(param.name)?;
+        let records = Vec::<NodeRecord>::deserialize(de)?;
```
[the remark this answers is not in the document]
slowli: Here's the `smart-config` PR: https://github.com/matter-labs/smart-config/pull/47

lib/network/src/service.rs
```diff
@@ -0,0 +1,111 @@
...
+            .required_block_hashes(vec![])
+            // Set network id to ZKsync OS chain's id, otherwise we might connect to unrelated peers
+            .network_id(Some(zk_provider_factory.chain_spec().chain.id()))
+            // Add latest version of `zks` subprotocol. In the future this can be extended so that
+            // several versions are registered here.
+            .add_rlpx_sub_protocol(ZksProtocolHandler::<ZksProtocolV1, _> {
+                replay,
+                // we only want to request blocks if this is not the main node
+                to_request_blocks: !is_main_node,
+                state: ProtocolState::new(protocol_tx, 10),
```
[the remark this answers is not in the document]
itegulov: Intentional as we try not to add too many configuration parameters for now. `10` is likely to be a good deafult for the foreseeable future. Moved it to a constant though.

lib/reth_compat/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+
+zk_os_api.workspace = true
+
+reth-chainspec = { workspace = true }
+reth-storage-api = { workspace = true }
+reth-trie-common = { workspace = true }
+reth-primitives-traits = { workspace = true }
+reth-revm = { workspace = true }
+
+alloy = { workspace = true, default-features = false }
```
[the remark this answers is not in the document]
itegulov: The workspace manifest already has `default-features = false` for alloy. Or do you mean that specifying it again on the crate level is redundant? Makes sense, I specify it purely out of habit and so it's immediately obvious just by looking at crate's `Cargo.toml`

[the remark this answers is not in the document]
slowli: Yeah, meant it was redundant.

PR #777, opened 2026-01-08
feat: add bash script to run local chains

vasyl-ivanchuk:
## Summary

Adds `run_local.sh` script that automates starting Anvil and chain node(s):

```bash
# Run a single chain (auto-detects latest version)
./run_local.sh

# Run a single chain (explicit path)
./run_local.sh ./local-chains/v31.0/default

# Run multiple chains
./run_local.sh ./local-chains/v31.0/multi_chain

# Run with logging to files
./run_local.sh ./local-chains/v31.0/multi_chain --logs-dir ./logs
```

Note that this is intended to be a temporary solution while we work on a better user-facing orchestrator. Therefore, I don't think we should invest much time in adding more features to the script.

#### How the Script Works

1. **Validates configuration directory** - Checks that the directory exists and contains `zkos-l1-state.json`
2. **Builds ZKsync OS**
3. **Starts Anvil** - Loads the L1 state snapshot on port 8545
4. **Waits for Anvil readiness** - Polls the JSON-RPC endpoint until Anvil responds (up to 30 seconds)
5. **Detects chain mode**:
   - If `config.json` exists → Starts single chain
   - Otherwise → Starts all `chain_*.json` files found (e.g., `chain_6565.json`, `chain_6566.json`)
6. **Database cleanup prompt** (single chain mode only) - If the `db/` folder contains existing data, prompts whether to clean it up before starting
7. **Monitors processes** - If any process fails, all services are stopped
8. **Graceful shutdown** - Press `Ctrl+C` to stop all services

#### Script Output

- **Anvil logs**: Suppressed (or written to `anvil-<timestamp>.log` if `--logs-dir` is specified)
- **Chain logs**: Displayed in terminal (or written to `<config-name>-<timestamp>.log` if `--logs-dir` is specified)
- **Script messages**: Color-coded status updates

### DIFF
--- a/local-chains/README.md
+++ b/local-chains/README.md
@@ -59,12 +59,12 @@ anvil --load-state ./local-chains/v30.2/default/zkos-l1-state.json --port 8545
 Node configuration file used to override the default values defined in the [config module](../node/bin/src/config).
 Commonly modified values include:
 
-- `genesis.chain_id` - Chain ID of the chain node operates on
-- `genesis.bridgehub_address` - Address of the Bridgehub contract on L1
-- `genesis.bytecode_supplier_address` - Address of the bytecode supplier contract
-- `l1_sender.operator_commit_pk` - Private key for committing batches
-- `l1_sender.operator_prove_pk` - Private key for proving batches
-- `l1_sender.operator_execute_pk` - Private key for executing batches
+- `genesis.chain_id` — Chain ID of the chain node operates on
+- `genesis.bridgehub_address` — Address of the Bridgehub contract on L1
+- `genesis.bytecode_supplier_address` — Address of the bytecode supplier contract
+- `l1_sender.operator_commit_sk` — Private key for committing batches
+- `l1_sender.operator_prove_sk` — Private key for proving batches
+- `l1_sender.operator_execute_sk` — Private key for executing batches
 
 ### `genesis.json`
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1620,6 +1620,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "155a5a185e42c6b77ac7b88a15143d930a9e9727a5b7b77eed417404ab15c247"
 
 [[package]]
+name = "assert-json-diff"
+version = "2.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "47e4f2b81832e72834d7518d8487a0396a28cc408186a2e8854c0f98011faf12"
+dependencies = [
+ "serde",
+ "serde_json",
+]
+
+[[package]]
 name = "assert_matches"
 version = "1.5.0"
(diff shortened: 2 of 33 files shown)


docs/src/setup/local_run.md
```diff
@@ -1,6 +1,18 @@
 ## Run
 
-### Local
+### Using the `run_local.sh` Script
```
popzxc: Can we add a warning mentioning that this script is a temporary solution until a proper orchestrator is implemented? Something like this:

```
⚠️ This script is a temporary solution that will later be replaced by a dedicated orchestrator binary. Do not depend on it in production
```

run_local.sh
```diff
@@ -0,0 +1,145 @@
+#!/bin/bash
```
popzxc: Same here -- I think it makes sense to echo the warning as well.

popzxc (approved):
LGTM, but I'd recommend getting one more review

run_local.sh
```diff
@@ -0,0 +1,149 @@
...
+    
+    for config_file in "${CHAIN_CONFIGS[@]}"; do
+        echo -e "${GREEN}Starting chain with config: $config_file${NC}"
+        cargo run --release -- --config "$config_file" &
+        CHAIN_PID=$!
+        PIDS+=($CHAIN_PID)
+        echo -e "${GREEN}Chain started with PID $CHAIN_PID${NC}"
+        
+        # Small delay between starting chains to avoid port conflicts
+        sleep 2
```
Deniallugo: sleep 2 is not enough, because we have a build part and it shows that everything started even though it's in a build process

run_local.sh
```diff
@@ -0,0 +1,160 @@
...
+done
+
+# Determine which chain configs to use
+SINGLE_CONFIG="$CONFIG_DIR/config.json"
+
+if [ -f "$SINGLE_CONFIG" ]; then
+    # Single chain mode
+    
+    # Prompt to clean up db folder (only for single chain mode)
+    if [ -d "db" ] && [ "$(ls -A db 2>/dev/null)" ]; then
```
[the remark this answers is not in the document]
vasyl-ivanchuk: Makes sense! Switched to absolute paths.

local-chains/README.md
```diff
@@ -60,7 +60,41 @@ If you are changing source code of any of the `initial_contracts` you should als
...
-### Running a Single Chain
+### Using the `run_local.sh` Script
+
+⚠️ This script is a temporary solution that will later be replaced by a dedicated orchestrator binary. Do not depend on it in production
+
+The `run_local.sh` script automates starting Anvil and chain node(s):
+
+```bash
+# Run a single chain
+./run_local.sh ./local-chains/v30
```
itegulov: AFAIU the directory format changed a bit (`default` contains single chain setup now). Also the directory is called `v30.2`.

itegulov (approved):

PR #779, opened 2026-01-09
feat!: token price updater component

perekopskiy:
## Summary

Adds a token price updater component responsible for polling relevant token prices, and setting base token price ratio on L1. This PR does not apply these prices to fee params yet, it will be done in a separate PR: `base_token_price_usd` will affect native price (and thus base fee as well) and `base_token_to_sl_base_token_ratio` will affect pubdata price.

Some code was moved from zksync-era. Main differences are:
- Prices are no longer persisted, they are only kept in memory (and reported to grafana).
- Price clients return price in USD, not ETH like in zksync-era. We'll need USD price to charge for native used properly, native translates to proving cost which is tied to USD.
- Price clients return price for base token unit which is `token_price / 10^token_decimals`. For Era it worked correctly for tokens with decimals=18, if decimals are different Era's code incorrectly uses ratios.

## Breaking Changes

Added new required config field `external_price_api_client_source`.

## Rollout Instructions

You should set config for `external_price_api_client` and `base_token_price_updater`. Docs: https://github.com/matter-labs/zksync-os-server/blob/078368a4e80ea3ac3f2d0418a3a99bc901cc5f00/docs/src/setup/base_token_price_updater.md.

Firstly, chain operator should decide on price source, options are `CoinGecko`, `CoinMarketCap` (3rd party APIs), or `Forced` which instantiates a client that returns prices that are configured, by default price randomly fluctuate a little to simulate real world scenario (fluctuation can be disabled).
For mainnets with base tokens that are not stablecoins tied to USD it's recommended to use one of 3rd party sources and provide an API key. In case you run a testnet and want fees to behave as on mainnet, you can still use 3rd party source and set config vars: `base_token_addr_override` should contain your mainnet token address that source can provide price for and `base_token_decimals_override` with token decimals (since token is on mainnet but node connects to testnet it cannot get the decimals from L1 and it has to be set manually).

Also, for chains with base token different from ETH it's recommended to configure `token_multiplier_setter_sk`, then new component will also update "ETH:token" price ratio on L1 as per updater's config. Component and node will still work without it but there will be a warning in logs and ratio on L1 won't change meaning that price for L1->L2 txs will eventually get outdated.

Other config fields have reasonable default values but it's still recommended to check and set if you have better values for the chain.

Also, it's recommended to rename `operator_<type>_pk` config vars to `operator_<type>_sk`. But "_pk" suffix is still supported

### DIFF
--- a/lib/base_token_adjuster/src/lib.rs
+++ b/lib/base_token_adjuster/src/lib.rs
@@ -7,13 +7,12 @@ use alloy::providers::fillers::{FillProvider, TxFiller};
 use alloy::providers::{Provider, WalletProvider};
 use alloy::rpc::types::TransactionReceipt;
 use alloy::rpc::types::trace::geth::{CallConfig, GethDebugTracingOptions};
+use alloy::signers::k256::ecdsa::SigningKey;
 use alloy::signers::local::PrivateKeySigner;
 use anyhow::Context;
 use num::rational::Ratio;
 use num::{BigUint, ToPrimitive};
-use secrecy::{ExposeSecret, SecretString};
 use std::collections::{HashMap, HashSet};
-use std::str::FromStr;
 use std::time::{Duration, Instant};
 use tokio::sync::watch;
 use zksync_os_contract_interface::{
@@ -48,7 +47,7 @@ pub struct BaseTokenPriceUpdaterConfig {
     /// Signing key to update base token price on L1.
     /// Must be consistent with the key set on the chain admin contract.
     /// It's not used for chains with ETH as base token and it's expected to be set for all other chains.
-    pub token_multiplier_setter_sk: Option<SecretString>,
+    pub token_multiplier_setter_sk: Option<SigningKey>,
     /// Max fee per gas we are willing to spend (in wei).
     pub max_fee_per_gas_wei: u128,
     /// Max priority fee per gas we are willing to spend (in wei).
@@ -72,10 +71,9 @@ pub struct BaseTokenPriceUpdater<
 
 async fn register_operator<P: Provider + WalletProvider<Wallet = EthereumWallet>>(
     provider: &mut P,
-    private_key: &SecretString,
+    signing_key: SigningKey,
 ) -> anyhow::Result<Address> {
-    let signer = PrivateKeySigner::from_str(private_key.expose_secret())
-        .context("failed to parse operator private key")?;
+    let signer = PrivateKeySigner::from_signing_key(signing_key);
     let address = signer.address();
     provider.wallet_mut().register_signer(signer);
 
@@ -103,12 +101,14 @@ impl<F: TxFiller<Ethereum> + WalletProvider<Wallet = EthereumWallet>, P: Provide
         base_token_adjuster_config: BaseTokenPriceUpdaterConfig,
         external_price_api_client_config: ExternalPriceApiClientConfig,
     ) -> anyhow::Result<Self> {
-        let token_multiplier_setter_address =
-            if let Some(pk) = &base_token_adjuster_config.token_multiplier_setter_sk {
-                Some(register_operator(&mut l1_provider, pk).await?)
-            } else {
-                None
-            };
+        let token_multiplier_setter_address = if let Some(sk) = base_token_adjuster_config
+            .token_multiplier_setter_sk
+            .clone()
+        {
+            Some(register_operator(&mut l1_provider, sk).await?)
+        } else {
+            None
+        };
 
         let base_token_address = base_token_adjuster_config
             .base_token_addr_override
--- a/lib/l1_sender/src/config.rs
+++ b/lib/l1_sender/src/config.rs
@@ -1,13 +1,13 @@
-use secrecy::SecretString;
+use alloy::signers::k256::ecdsa::SigningKey;
 use std::marker::PhantomData;
 use std::time::Duration;
 
 /// Configuration of L1 sender.
 #[derive(Clone, Debug)]
 pub struct L1SenderConfig<Input> {
-    /// Private key to operate from.
+    /// Signing key to operate from.
     /// Depending on the mode, this can be a commit/prove/execute operator.
-    pub operator_sk: SecretString,
+    pub operator_sk: SigningKey,
 
     /// Max fee per gas we are willing to spend (in wei).
     pub max_fee_per_gas_wei: u128,
--- a/lib/l1_sender/src/lib.rs
+++ b/lib/l1_sender/src/lib.rs
@@ -21,12 +21,11 @@ use alloy::providers::fillers::{FillProvider, TxFiller};
 use alloy::providers::{PendingTransactionError, Provider, WalletProvider};
 use alloy::rpc::types::trace::geth::{CallConfig, GethDebugTracingOptions};
 use alloy::rpc::types::{TransactionReceipt, TransactionRequest};
+use alloy::signers::k256::ecdsa::SigningKey;
 use alloy::signers::local::PrivateKeySigner;
 use anyhow::Context;
 use futures::future::BoxFuture;
 use futures::{FutureExt, StreamExt, TryStreamExt};
-use secrecy::{ExposeSecret, SecretString};
-use std::str::FromStr;
 use std::time::Duration;
 use tokio::sync::mpsc::Sender;
 use zksync_os_observability::{ComponentStateHandle, ComponentStateReporter};
@@ -78,8 +77,7 @@ pub async fn run_l1_sender<Input: SendToL1>(
         ComponentStateReporter::global().handle_for(Input::NAME, L1SenderState::WaitingRecv);
     let command_name = Input::NAME;
 
-    let operator_address =
-        register_operator::<_, Input>(&mut provider, &config.operator_sk).await?;
+    let operator_address = register_operator::<_, Input>(&mut provider, config.operator_sk).await?;
     let mut cmd_buffer = Vec::with_capacity(config.command_limit);
 
     // Process all potential passthrough commands first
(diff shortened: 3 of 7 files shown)


perekopskiy (commented):
It's possible to make PR non-breaking and set forced client with a pre-configured ETH price as a default. However, I'm afraid that it can be missed for some chains resulting in incorrect fee calculation. IMO it's better to keep it new required config. Feel free to comment if you disagree

lib/base_token_adjuster/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// Maximum number of attempts to fetch quote from a remote API before failing over.
+    pub price_fetching_max_attempts: u32,
+    /// Override for address of the base token address.
+    pub base_token_addr_override: Option<Address>,
+    /// Override for decimals of the base token.
+    pub base_token_decimals_override: Option<u8>,
+    /// Override for address of the gateway base token address used to calculate ETH<->GatewayBaseToken ratio on gateway using chains.
+    pub gateway_base_token_addr_override: Option<Address>,
+    /// Private key to update base token price on L1.
+    /// Must be consistent with the key set on the chain admin contract.
```
slowli: Nit / future work (?): Is there a way to check this on startup? E.g., is there a relevant getter in the chain admin contract?

lib/base_token_adjuster/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+                    }]
+                        .observe(start_time.elapsed());
+                    tokio::time::sleep(Duration::from_secs(5)).await;
+                }
+            }
+        }
+        let error_message = "Failed to fetch base token ratio after multiple attempts";
+        Err(last_error
+            .map(|x| x.context(error_message))
+            .unwrap_or_else(|| anyhow::anyhow!(error_message)))
```
slowli: Nit: This looks overly defensive; AFAIU, `last_error` is always set here.

lib/base_token_adjuster/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+            // No setter address configured, nothing to do.
+            return Ok(());
+        };
+        let old_ratio = self.last_l1_ratio.clone();
+
+        // Instead of converting to signed bigint it's easier to do an if-else here.
+        let diff = if new_ratio > old_ratio {
+            new_ratio.clone() - old_ratio.clone()
+        } else {
+            old_ratio.clone() - new_ratio.clone()
```
slowli: Nit: Arithmetic operations are defined on `Ratio` references, too, so it's possible to use expressions like `&new_ratio - &old_ratio` here and in similar places.

lib/base_token_adjuster/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+        };
+        let deviation = (diff / old_ratio.clone()) * BigUint::from(100u32);
+        if deviation
+            >= Ratio::from_integer(BigUint::from(self.config.l1_update_deviation_percentage))
+        {
```
slowli: Nit: Could make sense to exit early on small `deviation` in order to decrease code indent / improve readability.

lib/external_price_api/src/cmc_api.rs
```diff
@@ -0,0 +1,390 @@
...
+                HeaderValue::from_str(api_key).expect("Failed to create header value"),
+            )]);
+
+            reqwest::Client::builder().default_headers(default_headers)
+        } else {
+            reqwest::Client::builder()
+        }
+        .timeout(config.client_timeout)
+        .build()
+        .expect("Failed to build reqwest client");
```
slowli: Bikeshedding: Propagating errors looks like a better long-term strategy.

lib/external_price_api/src/cmc_api.rs
```diff
@@ -0,0 +1,390 @@
...
+}
+
+impl CmcPriceApiClient {
+    pub fn new(config: ExternalPriceApiClientConfig) -> Self {
+        let client = if let Some(api_key) = &config.api_key {
+            use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
+
+            let default_headers = HeaderMap::from_iter([(
+                HeaderName::from_static(AUTH_HEADER),
+                HeaderValue::from_str(api_key).expect("Failed to create header value"),
```
slowli: Please call `.set_sensitive(true)` so that the value isn't logged etc. Ditto for the other usage of `api_key`.

lib/base_token_adjuster/src/lib.rs
perekopskiy: Added check on startup

lib/base_token_adjuster/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+
+    async fn loop_iteration(&mut self) -> anyhow::Result<()> {
+        let tokens_to_watch = HashSet::from([self.base_token, self.sl_token, APIToken::ETH]);
+
+        let mut token_prices = HashMap::new();
+        for token in tokens_to_watch {
+            if token_prices.contains_key(&token) {
+                continue;
+            }
+            let ratio = self.retry_fetch_ratio(token).await?;
```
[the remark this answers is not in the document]
perekopskiy: At the moment it brings a little benefit since we have at most 3 tokens to fetch prices for. I wouldn't do requests concurrently to keep things simpler for now.

lib/base_token_adjuster/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+            let balance = format_ether(
+                provider
+                    .get_balance(token_multiplier_setter_address)
+                    .await?,
+            );
+            let nonce = provider
+                .get_transaction_count(token_multiplier_setter_address)
+                .await?;
+            METRICS.l1_updater_balance.set(balance.parse()?);
+            METRICS.l1_updater_nonce.set(nonce);
```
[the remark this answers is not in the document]
perekopskiy: Given that the component is not on a critical path (e.g. like sequencer) I think it's ok to keep this code in the main tokio task. Component already does quite some l1 provider calls that are non-metric related and these 2 calls do not add much latency or risk of failure

lib/base_token_adjuster/src/lib.rs
```diff
@@ -0,0 +1,456 @@
...
+    config: BaseTokenPriceUpdaterConfig,
+    last_l1_ratio: Ratio<BigUint>,
+    chain_admin_contract: IChainAdminOwnableInstance<FillProvider<F, P>, Ethereum>,
+    token_multiplier_setter_address: Option<Address>,
+    zk_chain_address: Address,
+}
+
+async fn register_operator<P: Provider + WalletProvider<Wallet = EthereumWallet>>(
+    provider: &mut P,
+    private_key: SecretString,
```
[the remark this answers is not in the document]
perekopskiy: Agree, should be possible, but I think we should implement `WellKnown` for the `SigningKey` type, right?

lib/types/src/base_token_price.rs
```diff
@@ -0,0 +1,41 @@
+use chrono::{DateTime, Utc};
+use num::rational::Ratio;
+use num::{BigInt, BigUint};
+
+/// Struct to represent API response containing denominator, numerator, and timestamp.
+#[derive(Debug, Clone, PartialEq)]
+pub struct BaseTokenApiRatio {
+    pub ratio: Ratio<BigUint>,
```
itegulov: Maybe make it clear here that `ratio` is USD value per one base unit?

lib/external_price_api/src/forced_price_client.rs
```diff
@@ -0,0 +1,114 @@
...
+/// (price is always a configured "forced" price with a configured fluctuation).
+#[derive(Debug)]
+pub struct ForcedPriceClient {
+    // Configured base ratios for tokens. Note, these are ratios for 1 **token**, not base unit.
+    eth_base_ratio: Option<Ratio<BigUint>>,
+    zk_base_ratio: Option<Ratio<BigUint>>,
+    erc20_base_ratios: HashMap<Address, Ratio<BigUint>>,
+
+    // Previous ratios to apply fluctuation against.
+    previous_ratios: Mutex<HashMap<APIToken, Ratio<BigUint>>>,
```
itegulov: Nit: no need for async Mutex here

lib/external_price_api/src/forced_price_client.rs
```diff
@@ -0,0 +1,114 @@
...
+        };
+        let decimals = token.decimals();
+        let Some(base_ratio) = base_ratio else {
+            anyhow::bail!("No forced price configured for token: {:?}", token);
+        };
+        let mut previous_ratios = self.previous_ratios.lock().await;
+        let previous_ratio = previous_ratios
+            .get(&token)
+            .cloned()
+            .unwrap_or(base_ratio.clone());
```
itegulov: Mega-nit: there is likely an idiomatic way to get a mutable entry here instead and avoid cloning

lib/base_token_adjuster/src/metrics.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub external_price_api_latency: Family<OperationResultLabels, Histogram<Duration>>,
+    /// Token multiplier setter's balance.
+    pub l1_updater_balance: Gauge<f64>,
+    /// Used to report token multiplier setter's address. Gauge is always set to one.
+    #[metrics(labels = ["address"])]
+    pub l1_updater_address: LabeledFamily<&'static str, Gauge, 1>,
+    /// Ratio of ETH:base token, that is set on L1.
+    pub ratio_l1: Gauge<f64>,
+    /// L1 Transaction fee in Ether (i.e. total cost of `setTokenMultiplier`).
+    #[metrics(buckets = Buckets::exponential(0.0001..=100.0, 3.0))]
```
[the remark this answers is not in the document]
slowli: JFYI, I've created a PR in `vise` that would make configuring buckets like this easier: https://github.com/matter-labs/vise/pull/42

PR #785, opened 2026-01-13
feat: add support for YAML config files

vasyl-ivanchuk:
## Summary

Add support for YAML config files.


popzxc (approved):
Generally looks good, but I'm a bit worried about supporting multiple config formats. Probably it makes sense to slowly deprecate JSON support (after this release), so that later we can remove it?

No action required -- it's just something to be discussed with the team.

slowli:
@popzxc IMO, supporting multiple formats is good from the UX perspective. OTOH, IMO it does make sense to transition to YAML *for the configs in the repo*, since it's more concise / readable.

PR #787, opened 2026-01-14
feat!: use token prices in fee model

perekopskiy:
## Summary

Token prices are now used for params calculation. Also refactors fee params calculation code. Adds a separate fee config.

## Breaking Changes

`base_fee_override`, `native_price_override`, `pubdata_price_override` are moved from sequencer config to fee config.

## Rollout Instructions

In case you use any of previosuly mentoned config variables, you should change their name from `sequencer_<name>` to `fee_<name>`, e.g. `sequencer_base_fee_override` should be renamed to `fee_base_fee_override` in your config.

### DIFF
--- a/node/bin/src/lib.rs
+++ b/node/bin/src/lib.rs
@@ -547,16 +547,23 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
     };
 
     let (token_price_sender, token_price_receiver) = watch::channel(None);
-    let previous_block_fee_params = first_replay_record
-        .as_ref()
-        .and_then(|record| {
-            block_replay_storage.get_replay_record(record.block_context.block_number - 1)
+    let previous_block_fee_params = if starting_block == 1 {
+        None
+    } else {
+        let prev_record = block_replay_storage
+            .get_replay_record(starting_block - 1)
+            .unwrap_or_else(|| {
+                panic!(
+                    "Missing replay record for block `starting_block - 1` = {}",
+                    starting_block - 1
+                )
+            });
+        Some(FeeParams {
+            eip1559_basefee: prev_record.block_context.eip1559_basefee,
+            native_price: prev_record.block_context.native_price,
+            pubdata_price: prev_record.block_context.pubdata_price,
         })
-        .map(|record| FeeParams {
-            eip1559_basefee: record.block_context.eip1559_basefee,
-            native_price: record.block_context.native_price,
-            pubdata_price: record.block_context.pubdata_price,
-        });
+    };
 
     // todo: `BlockContextProvider` initialization and its dependencies
     // should be moved to `sequencer`


lib/sequencer/src/execution/fee_provider.rs
```diff
@@ -0,0 +1,291 @@
...
+use tokio::sync::watch;
+use zksync_os_storage_api::ReplayRecord;
+use zksync_os_types::{PubdataMode, TokenPricesForFees};
+
+/// Fee-related configuration.
+#[derive(Debug, Clone)]
+pub struct FeeConfig {
+    /// Price for one unit of native resource in USD.
+    /// Default is set based on the current estimate of proving price.
+    pub native_price_usd: f64,
```
itegulov: Nit: I'd store `Ration<BigUint>` here directly to avoid calculating it on every block

lib/sequencer/src/execution/fee_provider.rs
```diff
@@ -0,0 +1,291 @@
...
+                r *= BigUint::from(*fill_ratio.denom());
+                r /= BigUint::from(*fill_ratio.numer());
+                r.to_integer()
+            };
+
+            tracing::debug!(
+                desired_pubdata_price = ?pubdata_price,
+                ?base_pubdata_price,
+                ?native_overhead,
+                ?fill_ratio,
```
itegulov: Nit: I believe `Ratio`'s `Debug` impl is pretty ugly, I'd use `Display` instead:
```suggestion
                %fill_ratio,
```
Probably makes sense to use `%` everywhere. Shouldn't matter for `BigUint` but it's easy to miss a sneaky `Ratio` among them.

lib/sequencer/src/execution/fee_provider.rs
```diff
@@ -0,0 +1,291 @@
...
+                const NATIVE_PRICE_MAX_CHANGE_DENOMINATOR: u32 = 8;
+
+                let previous_native_price =
+                    BigUint::from_bytes_le(previous_native_price.as_le_slice());
+
+                let min_native_price = &previous_native_price
+                    - (&previous_native_price / BigUint::from(NATIVE_PRICE_MAX_CHANGE_DENOMINATOR));
+                let max_native_price = &previous_native_price
+                    + (&previous_native_price / BigUint::from(NATIVE_PRICE_MAX_CHANGE_DENOMINATOR))
+                        .max(BigUint::from(1u32));
```
itegulov: Why `max(1)` here?

docs/src/setup/base_token_price_updater.md
```diff
@@ -0,0 +1,84 @@
...
+- `base_token_decimals_override` - token decimals (since token is on mainnet but node connects to testnet it cannot get the decimals from L1);
+    in case base token is ETH or ZK it can be omitted.
+Similarly, you can set `gateway_base_token_addr_override` to ZK mainnet address in case settlement layer is Gateway.
+
+Example configuration for a chain that settles to Gateway, and chain's base token is USDC:
+```yaml
+base_token_price_updater:
+  base_token_addr_override: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48" # USDC
+  base_token_decimals_override: 6 # USDC decimals
+  gateway_base_token_addr_override: "0x66a5cfb2e9c529f14fe6364ad1075df3a649c0a5" # ZK
```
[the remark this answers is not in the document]
perekopskiy: I didn't add it because I don't see why someone would set override for gateway token to anything other than ZK. Also it's unused currently, and I didn't want to add another config variable. When it's used I think it will be an explicit error on start-up. Can be reconsidered if there is a use-case for non-zk gateway token override.

PR #788, opened 2026-01-14
chore: move local chain configs from json to yaml

antonbaliasnikov:
## Summary

* [x] Move local chain configs from `json` to `yaml`
* [x] Introduce minimal interface to load configs for integration tests 

## Why?

To improve user experience working with configs and add ability to add comments into configs if required.

### DIFF
--- a/run_local.sh
+++ b/run_local.sh
@@ -185,7 +185,7 @@ for i in {1..30}; do
 done
 
 # Determine which chain configs to use
-SINGLE_CONFIG="$CONFIG_DIR/config.json"
+SINGLE_CONFIG="$CONFIG_DIR/config.yaml"
 
 if [ -f "$SINGLE_CONFIG" ]; then
     # Single chain mode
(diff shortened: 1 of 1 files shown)


vasyl-ivanchuk:
@antonbaliasnikov any reason why you do not migrate genesis files to yaml? Apart from that, could you please also update `run_local.sh` to work with yaml instead of json?

antonbaliasnikov:
> update `run_local.sh` to work with yaml instead of json

Done, good catch. Wondering why my global search did not find it earlier...

integration-tests/src/config.rs
```diff
@@ -1,110 +1,130 @@
...
-    let config_path = format!("{workspace_dir}/local-chains/{version}/default/config.json");
-    let config_schema = Config::schema();
-    let mut config_sources = ConfigSources::default();
-    let config_contents =
-        std::fs::read_to_string(&config_path).expect("Failed to read config file");
+impl<'a> ChainLayout<'a> {
+    fn chain_id(self) -> Option<u64> {
+        match self {
+            ChainLayout::Default { .. } => None,
+            ChainLayout::MultiChain { chain_index, .. } => Some(6565u64 + chain_index as u64),
```
perekopskiy: nit: 6565u64 can be moved to constant var

PR #800, opened 2026-01-16
feat: return zeroes in `reward` in `eth_feeHistory`

perekopskiy:
## Summary

Return zeroes in `reward` in `eth_feeHistory`. ZKsync OS chains are not fully EIP-1559 compliant and using 0 as a priority fee should always work, so we return zeroes to keep code simpler.


lib/rpc/src/eth_impl.rs
```diff
@@ -358,10 +358,31 @@ impl<RpcStorage: ReadRpcStorage, Mempool: L2TransactionPool> EthNamespace<RpcSto
...
         let end_block_plus = end_block + 1;
         // Ensure that we would not be querying outside of genesis
-        let block_count = end_block_plus.min(block_count.try_into().unwrap());
+        let mut block_count = end_block_plus.min(block_count.try_into().unwrap());
+
+        const MAX_FEE_HISTORY_BLOCKS: u64 = 1024;
+        const MAX_REWARD_PERCENTILE_COUNT: usize = 100;
+        if block_count > MAX_FEE_HISTORY_BLOCKS {
+            block_count = MAX_FEE_HISTORY_BLOCKS;
+        }
```
itegulov: Nit: maybe use `[end_block_plust, block_count, MAX_FEE_HISTORY_BLOCKS].min()` above to keep all requirements in one place and avoid `block_count` from being mutable?

lib/rpc/src/eth_impl.rs
```diff
@@ -372,6 +393,34 @@ impl<RpcStorage: ReadRpcStorage, Mempool: L2TransactionPool> EthNamespace<RpcSto
...
+                                BlockNumberOrTag::Number(block),
+                            )))?,
+                    );
+                }
+                let r = Self::calculate_reward_percentiles_for_block(
+                    percentiles,
+                    repository_block.header.gas_used,
+                    base_fee.saturating_to(),
+                    &stored_txs,
+                );
```
itegulov: Open question: do we even want to calculate real EIP-1559 rewards? zksync-os does not really support EIP-1559 (in a sense that fees do not fluctuate according to EIP-1559 spec) and priority fees are not collected/relevant for us. I suppose the only reason why priority fees can matter right now is our mempool implementation that sorts transactions based on that (default behaviour inherited from reth but can be configured). Which would only be relevant in extreme load scenarios.

This is why I did not implement reward logic originally and decided to postpone at the time. Can we get away with just returning `[0, 0, ..., 0]` every time?

lib/rpc/src/eth_impl.rs
```diff
@@ -356,9 +356,32 @@ impl<RpcStorage: ReadRpcStorage, Mempool: L2TransactionPool> EthNamespace<RpcSto
...
+        const MAX_REWARD_PERCENTILE_COUNT: usize = 100;
+        if reward_percentiles.as_ref().map(|perc| perc.len()) > Some(MAX_REWARD_PERCENTILE_COUNT) {
+            return Err(EthError::InvalidRewardPercentiles);
+        }
+        // If reward percentiles were specified, we
+        // need to validate that they are monotonically
+        // increasing and 0 <= p <= 100
+        if let Some(percentiles) = &reward_percentiles
+            && (percentiles.windows(2).any(|w| w[0] > w[1] || w[0] > 100.)
+                || percentiles.first().is_some_and(|w| *w > 100.0 || *w < 0.0))
```
itegulov: Looks like you still don't check that `w[i] < 0.0` for `i > 0`. Also you need special handling for `last` instead of `first`.

I think it's easier and less error prone to write two separate conditions for `windows(2).any(w[0] > w[1])` and `iter().any(x < 0 || x > 100)`

lib/rpc/src/eth_impl.rs
```diff
@@ -396,8 +419,7 @@ impl<RpcStorage: ReadRpcStorage, Mempool: L2TransactionPool> EthNamespace<RpcSto
             gas_used_ratio: vec![0.5; block_count as usize],
             base_fee_per_blob_gas: vec![],
             blob_gas_used_ratio: vec![],
-            // TODO: fill reward
-            reward: None,
+            reward: reward_percentiles.map(|p| vec![vec![0; p.len()]; block_count as usize]),
```
itegulov: Nit: please leave a comment here why we decided to return `[0; N]`

PR #802, opened 2026-01-17
fix: prevent "subtract with overflow" error on EN startup 

RomanBrodetski:
This happens sometimes when all batches are committed and executed. Adding `1` before subtracting shoudl fix it.

### DIFF
--- a/lib/l1_watcher/src/util.rs
+++ b/lib/l1_watcher/src/util.rs
@@ -82,7 +82,9 @@ async fn find_latest_l1_revert(
     // Early return if latest block is behind start block. This can happen if we hit different
     // L1 nodes between calls where the second node is behind the first.
     if latest_block < start_block_number {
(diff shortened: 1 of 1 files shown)


lib/l1_watcher/src/util.rs
```diff
@@ -79,7 +79,7 @@ async fn find_latest_l1_revert(
         "checking for revert events"
     );
 
-    let blocks_to_scan = latest_block - start_block_number + 1;
+    let blocks_to_scan = latest_block + 1 - start_block_number;
```
itegulov: I think it makes sense to add an early return clause instead (`if latest_bn < start_bn { return Ok(None); }`). Two reasons:
1. It will give us an opportunity to leave an explicit comment
2. I believe there is a possibility for diff to be `2` if we happen to hit different L1 nodes in-between two calls (i.e. second L1 node is 1 block behind)

itegulov (approved):

itegulov (approved):

PR #803, opened 2026-01-19
feat!: Execution of service interop transactions

Artemka374:
### IMPORTANT NOTE
While this PR introduces changes that affect execution of system transactions, zksync-os version was not updated as it is not yet expected that such transactions will be provided for execution

### Changelog

Implementation of execution of service interop transaction & additional changes to their interface

* Removed `SystemTransaction` and related types. Left API as `InteropRootsTx` and `InteropRootsEnvelope`. (Implementation became too specific to interop, so the generic type didn't feel right)
* Added new field to replay record:
  - `starting_interop_event_index` - basically the same as `starting_l1_priority_id`, but for interop roots transaction(and stores the index of event - block number and index in block)
* A few additional changes to block execution API:
  - `BlockOutputExt` struct that contains `BlockOutput` and `last_interop_log_index` which is calculated dynamically during execution and being used for advancing `next_interop_log_index` in `BlockContextProvider`
  - `ZkPoolTransaction` struct that contains `ZkTransaction` and additional metadata(for now only interop log index if the transaction is interop one) that is being used for calculating `last_interop_log_index`
* Implemented sealing logic for interop transactions:
  - If the first transaction in stream was interop - `best_transaction_stream` provides only interop transactions
  - Block can contain up to 1000 interop roots
  - Batch can contain up to 1000 interop roots

## Breaking changes
* Changed replay record schema - 1 field added

## Rollout instructions
* ENs must be upgraded to this version before the main node does


lib/storage_api/src/replay_wire_format/v6.rs
```diff
@@ -0,0 +1,85 @@
...
+//! (This is enforced by CI)
+
+use bincode::{Decode, Encode};
+
+// It is somewhat safe to assume that these will not change
+use alloy::primitives::{Address, B256, U256};
+use zksync_os_types::{InteropRootsLogIndex, ProtocolSemanticVersion};
+
+// Differences from v5:
+// - added `interop_root_log_start_index` and `interop_root_indexes` to `ReplayWireFormatV6`
```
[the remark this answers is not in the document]
Artemka374: yeah sorry just an artifact from playing around

lib/network/src/wire/replays/mod.rs
```diff
@@ -5,6 +5,7 @@
 
 pub mod v0;
 pub mod v1;
+pub mod v2;
```
itegulov: Feel free to break `v1`. It's not being used yet.

lib/storage/src/db/replay.rs
```diff
@@ -294,6 +309,22 @@ impl ReadReplay for BlockReplayStorage {
...
+            let stored: InteropRootsLogIndex = bincode::serde::decode_from_slice(
+                &starting_interop_event_index,
+                bincode::config::standard(),
+            )
+            .expect("Failed to deserialize starting interop event index")
+            .0;
+            stored
+        } else {
+            InteropRootsLogIndex::default()
+        };
```
itegulov: The implementation looks backwards-compatible to me. If it is, please remove `Breaking Changes` and `Rollout Instructions` from PR description.

lib/types/src/transaction/interop/mod.rs
```diff
@@ -0,0 +1,327 @@
...
+    pub hash: B256,
+    /// Log index of the first event from which the transaction was created.
+    /// In case we use demo-version, it is the index of the only one event in transaction.
+    /// Stored in an envelope to be able to easier keep track of it, but it is not part of the transaction
+    #[serde(skip)]
+    pub first_log_index: InteropRootsLogIndex,
+    /// Log index of the last event from which the transaction was created.
+    /// Stored in an envelope to be able to easier keep track of it, but it is not part of the transaction
+    #[serde(skip)]
+    pub last_log_index: InteropRootsLogIndex,
```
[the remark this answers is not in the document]
itegulov: I am asking because if there is any source of non-determinism in interop roots L1 watcher, then state recovery logic might not work the way we expected it to when we were designing this

itegulov: > Stored in an envelope to be able to easier keep track of it, but it is not part of the transaction

So far reading the code it _does_ seem to me like these two fields should be a part of the transaction struct itself. You use them in RLP encoding/decoding. And they are not recoverable purely from `inner`.

itegulov: After reading more I think you might not even need to persist these fields in RocksDB. Looks like you only ever use them in contexts where you indexed these transactions from SL and hence have immediate access to `{first,last}_log_index` if you adapt the logic a little bit (happy to elaborate further but I think it might be easier to show interactively)

perekopskiy (dismissed):

PR #806, opened 2026-01-19
feat: Interop sequencer implementation

Artemka374:
## Summary

Implement sealing logic for system interop transactions.
Added new sealing criterias:
* Max 1000 interop roots per block and per batch (can be revisited)

### DIFF
--- a/lib/mempool/src/stream.rs
+++ b/lib/mempool/src/stream.rs
@@ -88,8 +88,11 @@ impl Stream for BestTransactionsStream<'_> {
                     Poll::Ready(Some(tx)) => {
                         return Poll::Ready(Some(ZkTransaction::from(tx)));
                     }
-                    Poll::Pending => return Poll::Pending,
+                    Poll::Pending if this.first_tx_is_interop => {
+                        return Poll::Pending;
+                    }
                     Poll::Ready(None) => todo!("channel closed"),
+                    _ => {}
                 }
             }
 


lib/sequencer/src/execution/block_executor.rs
```diff
@@ -87,7 +90,7 @@ pub async fn execute_block<R: ReadStateHistory + WriteState>(
                 latency_tracker.enter_state(SequencerState::Execution);
                 match maybe_tx {
                     /* ----- got a transaction with gas limit within the block gas limit left --- */
-                    Some(tx) if cumulative_gas_used + tx.inner.gas_limit() <= ctx.gas_limit => {
+                    Some(tx) if should_exclude_and_seal(&ctx, cumulative_gas_used, interop_roots_count, &tx).is_none() => {
```
perekopskiy: nit: let's refactor a bit to make code more readable, I suggest doing
```
let Some(tx) = maybe_tx else {
       tracing::debug!(
                            block_number = ctx.block_number,
                            txs = executed_txs.len(),
                            "stream exhausted → sealing"
                        );
                        break SealReason::TxStreamExhausted;
}
```
then you have unwrapped tx and so you can match the result of `should_exclude_and_seal` without doing an `if` inside the match

lib/mempool/src/stream.rs
```diff
@@ -80,11 +83,17 @@ impl Stream for BestTransactionsStream<'_> {
...
+            if !this.txs_already_provided || this.first_tx_is_interop {
+                match this.interop_transactions.poll_recv(cx) {
+                    Poll::Ready(Some(tx)) => {
+                        return Poll::Ready(Some(ZkTransaction::from(tx)));
+                    }
+                    Poll::Pending if this.first_tx_is_interop => {
+                        return Poll::Pending;
+                    }
+                    Poll::Ready(None) => todo!("channel closed"),
+                    _ => {}
```
itegulov: This matches only `Poll::Pending` but if first tx is not interop right? I'd handle explicitly and add a comment

lib/mempool/src/stream.rs
```diff
@@ -123,18 +132,25 @@ impl TxStream for BestTransactionsStream<'_> {
     }
 }
 
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug, Clone)]
+pub struct PeekedInfo {
+    pub upgrade_info: Option<UpgradeTransaction>,
+    pub is_peeked_tx_interop: bool,
+}
```
itegulov: I think it will be a bit cleaner to have `enum PeekedBlockType { Upgrade(UpgradeTransaction), Interop(InteropTransaction), Regular }` but can be done in a follow-up PR

PR #807, opened 2026-01-20
fix(rpc): make `eth_estimateGas` work when sender has no balance

itegulov:
## Summary

This rollbacks [#549](https://github.com/matter-labs/zksync-os-server/pull/549)'s fee calculation changes during simulation. And effectively disables base fee checks during estimation by setting `basefee=0`.

### DIFF
--- a/lib/rpc/src/call_fees.rs
+++ b/lib/rpc/src/call_fees.rs
@@ -19,8 +19,6 @@ impl CallFees {
         call_max_fee_per_gas: Option<u128>,
         call_max_priority_fee_per_gas: Option<u128>,
         block_base_fee: u128,
-        // todo: remove once confirmed that this is not needed
-        _for_estimate_gas: bool,
     ) -> Result<Self, CallFeesError> {
         match (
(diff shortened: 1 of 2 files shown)


lib/rpc/src/call_fees.rs
```diff
@@ -19,27 +19,18 @@ impl CallFees {
...
-                    (false, _) => {
-                        // either legacy transaction or no fee fields are specified
-                        // when no fields are specified, set gas price to zero
-                        gas_price.unwrap_or_default()
-                    }
-                    // only enforce the fee cap if provided input is not zero
-                    (_, None | Some(0)) => block_base_fee,
-                    (_, Some(gas_price)) if gas_price < block_base_fee => {
-                        return Err(CallFeesError::FeeCapTooLow);
-                    }
```
perekopskiy: IMO it's a bit weird that we check `max_fee_per_gas < block_base_fee` a few lines below but allow `gas_price < block_base_fee`.

PR #827, opened 2026-01-27
feat: do not require S3 for RPC

itegulov:
## Summary

Based on top of #825 (~requires to be merged first~ **merged**)

Introduces `WriteBatch` trait (on top of existing `ReadBatch`) implemented by a new RocksDB-backed implementation. It is populated in the background by a new watcher `L1PersistBatchWatcher` that only persists **executed** (i.e. immutable) batches.

This makes ENs completely S3-agnostic. The only remaining usage of S3 is to save successful/failed proofs and serve them via prover API which might still be useful for debugging. CC @EmilLuta


itegulov:
> What about SL reorgs?

I left a comment with your thoughts + added some of mine. I think we can address this by tracking finalized finality (terrible combination of terms I know), then there is no chance executed batch can be reorged.

PR #835, opened 2026-01-28
feat!: Use log id from event for interop root indexing

Artemka374:
## Summary

Rework interop watcher persistence to work with log id instead of event indexes in block.
All the places where InteropRootLogIndex was used was made to work with explicit id of event instead of event index.

## Breaking changes
Replay record format was changed - remove old value for root indexes and added new value for root id.
Therefore, latest replay record version was changed

## Rollout instructions
ENs should be upgraded before main node

### DIFF
--- a/lib/l1_watcher/src/interop_watcher.rs
+++ b/lib/l1_watcher/src/interop_watcher.rs
@@ -78,7 +78,7 @@ impl ProcessRawEvents for InteropWatcher {
             indexes.insert(sol_event.logId, log);
         }
 
-        return indexes.values().cloned().collect();
+        indexes.into_values().collect()
     }
 
     async fn process_raw_event(&mut self, log: Log) -> Result<(), L1WatcherError> {


lib/l1_watcher/src/interop_watcher.rs
```diff
@@ -1,90 +1,82 @@
...
             );
             return Ok(());
         }
-
         let interop_root = InteropRoot {
             chainId: tx.chainId,
-            blockOrBatchNumber: tx.batchNumber,
-            sides: vec![tx.chainRoot],
+            blockOrBatchNumber: tx.blockNumber,
+            sides: tx.sides.clone(),
```
itegulov: Nit: why clone here?

lib/sequencer/src/execution/block_executor.rs
```diff
@@ -358,6 +358,7 @@ fn should_exclude_and_seal(
     }
     if let ZkEnvelope::InteropRoots(interop_roots_tx) = tx.inner.inner()
         && interop_roots_count + interop_roots_tx.interop_roots_count() > interop_roots_per_block
+        && interop_roots_count + interop_roots_tx.interop_roots_count() > interop_roots_per_block
```
itegulov: Second check to make absolutely sure this holds? :) jkjk

lib/storage/src/db/replay.rs
```diff
@@ -58,7 +58,7 @@ impl NamedColumnFamily for BlockReplayColumnFamily {
         BlockReplayColumnFamily::ProtocolVersion,
         BlockReplayColumnFamily::BlockOutputHash,
         BlockReplayColumnFamily::ForcePreimages,
-        BlockReplayColumnFamily::StartingInteropEventIndex,
+        BlockReplayColumnFamily::StartingInteropRootId,
```
itegulov: So side effect of this is we'll have garbage orphan data under old "starting_interop_event_index" CF, correct? That seems acceptable if so

lib/storage_api/src/replay_wire_format/v7.rs
```diff
@@ -0,0 +1,84 @@
...
+//! (This is enforced by CI)
+
+use bincode::{Decode, Encode};
+
+// It is somewhat safe to assume that these will not change
+use alloy::primitives::{Address, B256, U256};
+use zksync_os_types::ProtocolSemanticVersion;
+
+// Differences from v5:
+// - added `starting_interop_event_index` to `ReplayWireFormatV6`
```
itegulov: Update this please

lib/storage_api/src/model.rs
```diff
@@ -47,9 +46,7 @@ pub struct ReplayRecord {
     pub block_output_hash: B256,
     /// Forced preimages to be included before the block execution.
     pub force_preimages: Vec<(B256, Vec<u8>)>,
-    /// Event index(block number and index in block) of the interop root tx executed first in the block
-    /// If there is no interop root tx in the block, equals to the previous block's value
-    pub starting_interop_event_index: InteropRootsLogIndex,
+    pub starting_interop_root_id: u64,
```
itegulov: Nit: add doc

lib/storage/src/db/replay.rs
Artemka374: yeah, kinda like that

PR #843, opened 2026-01-29
docs: add fees model documentation

perekopskiy:
## Summary

Add fees model documentation


cytadela8 (dismissed):
I ❤️ docs

docs/src/setup/fees.md
itegulov: `setup` is the wrong directory for this IMHO, I'd put into `design`. Also same for `base_token_price_updater.md`, I missed that it was put into `setup`.

docs/src/setup/fees.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Fee configuration and calculation
+
+### Native price
+
+Since native reflects ZK proving costs, it should be calculated based on 2 things:
+- prover machine cost
+- prover performance (how many native units it processes per second)
+
+Native price is configured with `FeeConfig::native_price_usd` (`fee_native_price_usd`).
```
itegulov: I remember you had some calculations you followed to derive default value that we currently use. Maybe provide an example of that calculation here, so it's clear how this should be changed under different market conditions?

docs/src/setup/fees.md
```diff
@@ -0,0 +1,77 @@
...
+- if calldata is used, then L1 gas price is used for calculation
+If rollup settles to Gateway, then gateway pubdate price is used.
+
+Pricing for blobs case is special because calculation of blob commitments is proven so it results in additional proving costs,
+thus pubdata price also depends on native price.
+Also, if chain settles frequently and posts blob that is not full then operator still needs to pay for the full blob.
+Node calculates a statistic for a fill ratio of submitted blobs and uses it to adjust pubdata price accordingly.
+
+Blob price is not stable on Ethereum testnets, and can grow a lot sometimes. 
+At some point it can be the case that pubdata price is high enough so that gas costs for small transactions are higher than block gas limit.
```
itegulov: Random thought: is there an easily calculable ratio between block gas limit and current pubdata price that makes sequencing stall? If yes, can we return appropriate error in `eth_estimateGas` that communicates to the user that L1 conditions are so severe that  transactions are not executable right now?

PR #846, opened 2026-01-30
feat: generate compressed states with anvil-1.6.0-5bcdddc0

itegulov:
## Summary

Alternative to #837 that uses custom anvil (temporarily)

L1 state was generated using https://github.com/matter-labs/zksync-os-scripts/pull/8, which:
* makes single-chain setup reuse the same L1 state as multichain
* enables preserved historical states


itegulov:
@popzxc it would certainly be better _if_ I had confidence the PR will get reviewed and merged soon. I am hopeful it will be at some point (and in some form), but this is blocking a few things including gateway integration. I don't want to put a rush on Paradigm to look at the PR -- well, it's not like we have leverage anyway. IMHO it's fine to use this random commit for now, but I could improve this PR slightly by moving this branch from my own fork to matterlabs/foundry (I would've done so from the start but I don't have write rights in that repo for some reason).

The other option is to keep this PR as a draft for now and keep using uncompressed state for gateway integration. It will permanently impact history size but maybe that's still acceptable.

PR #848, opened 2026-01-30
feat: Accumulated interop txs

Artemka374:
## Summary

This PR introduces changes to crafting system transactions from interop roots.
Introduced `InteropTxPool` and `InteropTransactions` stream. 
`InteropTxPool` behaves as a storage for all received interop roots and it can provide a stream `InteropTransactions`(returning values from which we can delay if needed) which returns interop transactions when polled. `on_canonical_state_change` will also operate on storage part - removing imported roots from itself.

BlockOutputExt was removed, since when on_canonical_state_change is called it effectively returns the index of last executed interop root(since roots are coming from watcher, indexes always exist), so there is no big difference between replay/produce/rebuild. It also consumes executed interop roots for Produce, and the ones from watcher during Replay/Rebuild. Also removed ZkPoolTransaction wrapper, since it's also not needed in this case.

Added config values for max_interop_roots_per_tx and max_interop_roots_per_batch instead of constants.

### DIFF
--- a/lib/sequencer/src/execution/block_context_provider.rs
+++ b/lib/sequencer/src/execution/block_context_provider.rs
@@ -48,6 +48,7 @@ pub struct BlockContextProvider<Mempool> {
     pubdata_limit: u64,
     interop_roots_per_block: u64,
     interop_roots_per_tx: usize,
+    service_block_delay: Duration,
     next_interop_tx_allowed_after: Instant,
     /// Protocol version to be used for the next produced block.
     /// Can change in runtime in case of upgrades.
@@ -73,6 +74,7 @@ impl<Mempool: L2TransactionPool> BlockContextProvider<Mempool> {
         pubdata_limit: u64,
         interop_roots_per_block: u64,
         interop_roots_per_tx: usize,
+        service_block_delay: Duration,
         protocol_version: ProtocolSemanticVersion,
         fee_collector_address: Address,
         last_constructed_block_ctx_sender: watch::Sender<Option<BlockContext>>,
@@ -92,6 +94,7 @@ impl<Mempool: L2TransactionPool> BlockContextProvider<Mempool> {
             pubdata_limit,
             interop_roots_per_block,
             interop_roots_per_tx,
+            service_block_delay,
             next_interop_tx_allowed_after: Instant::now(),
             protocol_version,
             fee_collector_address,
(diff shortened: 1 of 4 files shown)


lib/mempool/src/interop_tx_stream.rs
```diff
@@ -0,0 +1,198 @@
...
+            let roots_to_consume = self
+                .pending_roots
+                .drain(starting_index..)
+                .rev() // reversing iterator as last element is the one received earliest
+                .collect::<Vec<_>>();
+
+            let tx = IndexedInteropRootsEnvelope {
+                log_index: roots_to_consume.last().unwrap().log_index.clone(),
+                envelope: InteropRootsEnvelope::from_interop_roots(
+                    roots_to_consume.iter().map(|r| r.root.clone()).collect(),
```
itegulov: Nit: no need to clone here if you do `into_iter()`

Also, fairly sure you don't have to collect twice in this method. But it might be awkward to get last element for `log_index` above. You'll have to use `.peekable()` before reversing the iterator or something along these lines. Up to you to decide if it makes sense to take into account.

lib/mempool/src/interop_tx_stream.rs
```diff
@@ -0,0 +1,198 @@
...
+            match self.receiver.try_recv() {
+                Ok(root) => {
+                    self.pending_roots.push_front(root);
+                    continue;
+                }
+                Err(TryRecvError::Empty) | Err(TryRecvError::Lagged(_)) => {
+                    if let Some(envelope) = self.take_tx(true) {
+                        return Some(envelope);
+                    }
+                    return None;
```
itegulov: Nit:
```suggestion
                    return self.take_tx(true);
```

lib/mempool/src/interop_tx_stream.rs
```diff
@@ -0,0 +1,198 @@
...
+            if let Some(envelope) = self.take_tx(false) {
+                return Some(envelope);
+            }
+
+            match self.receiver.try_recv() {
+                Ok(root) => {
+                    self.pending_roots.push_front(root);
+                    continue;
+                }
+                Err(TryRecvError::Empty) | Err(TryRecvError::Lagged(_)) => {
```
itegulov: I think `Lagged` is a fatal error in our case, no? It means the channel got full and an older interop root was overwritten by a newer one. But we care about _all_ interop roots so we cannot ignore that some interop roots got skipped. Maybe it makes more sense for this channel to be unbounded?

PR #849, opened 2026-02-02
feat: Set SL chain id txs

Artemka374:
## Summary

Reworked `InteropTx` interface to be generic `SystemTx`. Added new subtype of system transaction for setting the SL chain id. Adjusted sequncer/batcher logic to include such transactions into their own blocks and assure that such block will be the first one in batch. Added the corresponding watcher to listen for migration events.


lib/rpc/src/eth_call_handler.rs
```diff
@@ -650,7 +650,7 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
 
 fn set_gas_limit(tx: &mut ZkTransaction, gas_limit: u64) {
     match tx.inner.inner_mut() {
-        ZkEnvelope::InteropRoots(_) => {
+        ZkEnvelope::System(_) => {
             unreachable!("interop roots transactions don't have explicit gas limit");
```
itegulov: Nit: "system transactions ..."

lib/sequencer/src/execution/block_executor.rs
```diff
@@ -134,12 +134,16 @@ pub async fn execute_block<R: ReadStateHistory + WriteState>(
...
                         );
 
-                        if let ZkEnvelope::InteropRoots(interop_roots_tx) = tx.inner.inner() {
-                            interop_roots_count += interop_roots_tx.interop_roots_count();
+                        if let ZkEnvelope::System(system_tx) = tx.inner.inner() {
+                            interop_roots_count += if let SystemTxType::ImportInteropRoots(roots_count) = system_tx.tx_type(){
+                                roots_count
+                            } else {
+                                0
+                            };
```
itegulov: Nit: something like this would look cleaner
```suggestion
                            if let SystemTxType::ImportInteropRoots(roots_count) = system_tx.tx_type() {
                                interop_roots_count += roots_count
                            }
```

lib/sequencer/src/execution/block_executor.rs
```diff
@@ -134,12 +134,16 @@ pub async fn execute_block<R: ReadStateHistory + WriteState>(
...
                         }
 
                         let tx_type = tx.tx_type();
-                        executed_txs.push(tx);
+                        executed_txs.push(tx.clone());
```
itegulov: Please avoid unnecessary cloning, I am pretty sure you can move this line further down the flow so that you won't need to clone

lib/sequencer/src/execution/block_executor.rs
itegulov: Nit: for debugging purposes I'd provide system tx (sub)type here as well

lib/types/src/transaction/system/mod.rs
```diff
@@ -1,39 +1,76 @@
...
 pub mod tx;
+pub mod utils;
 
-pub const BOOTLOADER_FORMAL_ADDRESS: Address =
-    address!("0x0000000000000000000000000000000000008001");
-pub const L2_INTEROP_ROOT_STORAGE_ZKSYNC_OS_ADDRESS: Address =
-    address!("0x0000000000000000000000000000000000010008");
-
-pub const INTEROP_ROOTS_TX_TYPE_ID: u8 = 125;
+pub use utils::{SYSTEM_TX_TYPE_ID, SystemTxType};
```
itegulov: Nit: remove empty line between `mod` and its re-exports

lib/types/src/transaction/system/mod.rs
```diff
@@ -1,39 +1,76 @@
...
+use alloy::primitives::{Address, B256, Bytes, TxKind, U256};
 use alloy::rpc::types::{AccessList, SignedAuthorization};
-use alloy::sol_types::SolCall;
 use alloy_rlp::{BufMut, Decodable, Encodable};
 use serde::{Deserialize, Serialize};
-use zksync_os_contract_interface::IMessageRoot::addInteropRootCall;
+use std::sync::OnceLock;
 use zksync_os_contract_interface::InteropRoot;
-
 pub mod tx;
```
itegulov: Nit: bring back the empty line between mod declarations and imports

lib/mempool/src/interop_tx_stream.rs
```diff
@@ -158,20 +158,25 @@ impl InteropTxPoolInner {
...
 
-            let envelope = InteropRootsEnvelope::from_interop_roots(
+            let envelope = SystemTxEnvelope::import_interop_roots(
                 roots.iter().map(|r| r.root.clone()).collect(),
             );
             log_index = roots.last().unwrap().log_index.clone();
 
-            assert_eq!(&envelope, &tx);
+            assert_eq!(&envelope.hash(), &tx.hash());
+            assert_eq!(&envelope.inner, &tx.inner);
```
itegulov: Presumably you made this change because `PartialEq` now takes `tx_type` into account? I think you should either:
1. derive it so that `tx_type` is ignored (`derive_more` crate can help you with that)
2. implement it manually so that `tx_type()` is called instead of plain field comparison (this is what alloy does for `hash: OnceLock<TxHash>` but I think in your case this is redundant)

lib/types/src/transaction/system/mod.rs
```diff
@@ -1,39 +1,76 @@
...
-#[derive(Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 #[serde(into = "tx_serde::TransactionSerdeHelper")]
-pub struct InteropRootsEnvelope {
+pub struct SystemTxEnvelope {
     /// Hash of the transaction
     /// Stored in an envelope and calculated separately from transaction as hash of transaction is not part of transaction itself.
     pub hash: B256,
-    pub inner: InteropRootsTx,
+    pub inner: SystemTx,
```
itegulov: Actually I would also make both `inner` and `hash` private. I think everything should be available through methods as is

lib/types/src/transaction/system/mod.rs
```diff
@@ -1,39 +1,76 @@
...
+    }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, Hash, Eq, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
 pub struct IndexedInteropRootsEnvelope {
     pub log_index: InteropRootsLogIndex,
-    pub envelope: InteropRootsEnvelope,
+    pub envelope: SystemTxEnvelope,
 }
```
itegulov: Is this type even used anywhere?

lib/types/src/transaction/system/mod.rs
```diff
@@ -302,36 +308,27 @@ impl Transaction for InteropRootsEnvelope {
...
 #[cfg(test)]
 mod tests {
-    use crate::InteropRootsEnvelope;
-    use crate::transaction::tx::InteropRootsTx;
+    use crate::SystemTxEnvelope;
 
     #[test]
     fn interop_roots_tx_serialization() {
         // Interop roots serialization should be consistent with Ethereum JSON-RPC spec
         // See https://ethereum.github.io/execution-apis/api-documentation/
```
itegulov: Please keep the original test and introduce a new one for `set_sl_chain_id`

lib/types/src/transaction/system/tx.rs
```diff
@@ -134,7 +134,7 @@ impl RlpEcdsaDecodableTx for InteropRootsTx {
 }
 
 // if something goes wrong with encoding, there's a chance that something is wrong here
```
itegulov: Question: is it safe to remove this comment at this point?

lib/types/src/transaction/system/mod.rs
```diff
@@ -1,39 +1,76 @@
...
+}
+
+impl SystemTxEnvelope {
+    pub fn import_interop_roots(roots: Vec<InteropRoot>) -> Self {
+        Self::create_from_input(SystemTxInput::ImportInteropRoots(roots))
+    }
+
+    pub fn set_sl_chain_id(chain_id: ChainId) -> Self {
+        Self::create_from_input(SystemTxInput::SetSLChainId(chain_id))
+    }
```
itegulov: I like how you represented these two helper methods, great job here. I'd just a single-liner doc to these methods or at least add a link to `SystemTxType` where these two subtypes are explained

lib/types/src/transaction/system/mod.rs
```diff
@@ -1,39 +1,76 @@
...
+
+    pub fn tx_type(&self) -> SystemTxType {
+        self.tx_type
+            .get_or_init(|| match SystemTxInput::abi_decode(self.inner.input()) {
+                SystemTxInput::ImportInteropRoots(roots) => {
+                    SystemTxType::ImportInteropRoots(roots.len() as u64)
+                }
+                SystemTxInput::SetSLChainId(_) => SystemTxType::SetSLChainId,
+            })
+            .clone()
```
itegulov: Remove `clone()` and make `tx_type()` return `&SystemTxType`. This would be conventional to "the Rust way" so that the caller can _choose_ to clone if need be but I expect we won't have to do it anywhere

lib/types/src/transaction/system/utils.rs
```diff
@@ -0,0 +1,65 @@
...
+use serde::{Deserialize, Serialize};
+use zksync_os_contract_interface::{
+    IMessageRoot::addInteropRootsInBatchCall, ISystemContext::setSettlementLayerChainIdCall,
+    InteropRoot,
+};
+
+pub const BOOTLOADER_FORMAL_ADDRESS: Address =
+    address!("0x0000000000000000000000000000000000008001");
+pub const L2_INTEROP_ROOT_STORAGE_ZKSYNC_OS_ADDRESS: Address =
+    address!("0x0000000000000000000000000000000000010008");
```
itegulov: Nit: the name is too verbose for what it represents. Server's code is ZKsync OS-specific so I'd rename to just `L2_INTEROP_ROOT_STORAGE_ADDRESS`.

Also, please replace existing `0x0000000000000000000000000000000000010008` declaration in integration-tests with this one

lib/types/src/transaction/mod.rs
```diff
@@ -198,7 +197,7 @@ impl From<L2Transaction> for ZkTransaction {
 impl fmt::Display for ZkTxType {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Self::InteropRoots => write!(f, "InteropRoots"),
+            Self::System => write!(f, "InteropRoots"),
```
itegulov: Nit: "System"

node/bin/src/batcher/seal_criteria.rs
```diff
@@ -60,13 +61,29 @@ impl BatchInfoAccumulator {
             .transactions
             .iter()
             .map(|tx| match tx.inner.inner() {
-                ZkEnvelope::InteropRoots(interop_roots_tx) => {
-                    interop_roots_tx.interop_roots_count()
+                ZkEnvelope::System(envelope) => {
+                    let SystemTxType::ImportInteropRoots(roots_count) = envelope.tx_type() else {
+                        return 0;
+                    };
+                    roots_count
```
itegulov: Nit: a bit cleaner like this IMHO
```suggestion
                    if let SystemTxType::ImportInteropRoots(roots_count) = envelope.tx_type() {
                        roots_count
                    } else {
                        0
                    }
```

itegulov: Off-topic: AFAIU if-let-guard is to be stabilized in rust 1.95 👀

node/bin/src/batcher/seal_criteria.rs
```diff
@@ -141,6 +158,15 @@ impl BatchInfoAccumulator {
             return true;
         }
 
+        // In case SL chain id update tx is present and there are other transactions in batch, we need it to go in the first block of the next batch
```
itegulov:
```suggestion
        // In case SL chain id update tx is present but not in the first block, we need to seal and
        // exclude. It will then go in the first block of the next batch.
```

itegulov (commented):
Apologies for a lot of comments. Nothing major though, mostly nits

PR #873, opened 2026-02-11
feat(network)!: fully migrate replay transport to p2p network

itegulov:
## Summary

This PR gets rid of legacy block replay transport protocol in favour of the new devp2p-driven implementation. Most of the implementation work was done some time ago before this PR, here I am mostly ironing out remaining issues and finalizing the migration.

I tested this manually on the new bootstrap node with static IP address. Hit me in DMs to give you an instruction on how to do it (don't want to expose the IP address here before we finalize everything).

## Breaking Changes

- Removed `sequencer_block_replay_download_address`
- Removed `sequencer_block_replay_server_enabled`
- Removed `sequencer_block_replay_server_address`
- `general_node_role` is a new config parameter that is required to be set on ENs
- Block replay transport is no longer hosted over HTTP on port 3053, instead [devp2p](https://github.com/ethereum/devp2p) protocol is used for p2p communication over TCP (port 3060 by default)

## Rollout Instructions

- Delete `sequencer_block_replay_*` config parameters from all nodes
- If your environment does not use external nodes (replay transport was unused before) - no further actions are required, otherwise continue down the list
- Make sure your cluster has at least one node with static IP address (it can be local as long as you do not want to connect external nodes from outside)
- Make sure all external nodes have `general_node_role` set to `external`
- Make sure networking is enabled and configured properly for all nodes (main and external), example:
```yaml
network:
  # Enables P2P networking, required for syncing ENs
  enabled: "true"
  # Secret key used for node identity in the network, can be randomly generated locally.
  # MUST be different for different nodes
  secret_key: "21b0ee131240821c39627c39d0fdde5edbda968c5877f5b63c5c542f267b5349"
  # Set to node's static IP address if there is one, leave unconfigured otherwise (defaults to 0.0.0.0)
  address: "14.28.133.107"
  # Comma-separated list of node records that have static IP addresses.
  # Can be left unconfigured if this is the node with static IP address. Otherwise this MUST contain at least one entry.
  # Format is `enode://<node ID>@<IP address>:<port>` where `node ID` is public key corresponding to that node's secret key (see above).
  # NOTE that node with static IP address does not necessarily need to be main node
  boot_nodes: "enode://dbd18888f17bad7df7fa958b57f4993f47312ba5364508fd0d9027e62ea17a037ca6985d6b0969c4341f1d4f8763a802785961989d07b1fb5373ced9d43969f6@14.28.133.115:3060"
```

### DIFF
--- a/node/bin/src/config/mod.rs
+++ b/node/bin/src/config/mod.rs
@@ -936,7 +936,7 @@ impl From<RpcConfig> for zksync_os_rpc::RpcConfig {
 impl From<&Config> for zksync_os_sequencer::config::SequencerConfig {
     fn from(c: &Config) -> Self {
         Self {
-            node_role: c.general_config.node_role.clone(),
+            node_role: c.general_config.node_role,
             block_time: c.sequencer_config.block_time,
             max_transactions_in_block: c.sequencer_config.max_transactions_in_block,
             block_dump_path: c.sequencer_config.block_dump_path.clone(),
--- a/node/bin/src/lib.rs
+++ b/node/bin/src/lib.rs
@@ -113,7 +113,7 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
 ) {
     let mut tasks: JoinSet<()> = JoinSet::new();
 
-    let node_role = config.general_config.node_role.clone();
+    let node_role = config.general_config.node_role;
     let role: &'static str = node_role.as_str();
 
     // Priority tree is required for main node
@@ -276,7 +276,7 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
         commit_proof_execute_block_numbers(&l1_state, &committed_batch_provider).await;
 
     let node_startup_state = NodeStateOnStartup {
-        node_role: node_role.clone(),
+        node_role,
         l1_state: l1_state.clone(),
         state_block_range_available: state.block_range_available(),
         block_replay_storage_last_block: block_replay_storage.latest_record(),
@@ -365,7 +365,7 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
 
         let network_service = NetworkService::new(
             config.network_config.clone().into(),
-            node_role.clone(),
+            node_role,
             block_replay_storage.clone(),
             starting_block,
             // This will be gone once we migrate away from record overrides
--- a/lib/network/src/protocol.rs
+++ b/lib/network/src/protocol.rs
@@ -46,7 +46,7 @@ impl<P: AnyZksProtocolVersion, Replay: Clone> ZksProtocolHandler<P, Replay> {
     ) -> ZksProtocolConnectionHandler<P, Replay> {
         ZksProtocolConnectionHandler {
             replay: self.replay.clone(),
-            node_role: self.node_role.clone(),
+            node_role: self.node_role,
             starting_block: self.starting_block,
             record_overrides: self.record_overrides.clone(),
             state: self.state.clone(),


node/bin/src/lib.rs
```diff
@@ -386,6 +362,45 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+    } else if node_role.is_main() {
+        tracing::info!(
+            "p2p networking is disabled; to enable set `network.enabled=true` and populate `network.secret_key`"
+        );
+    } else {
+        panic!(
+            "EN cannot run without p2p networking; to fix: \
+            set `network.enabled=true` to enable p2p networking, \
+            populate `network.secret_key` with a 256-bit ECDSA key (can be randomly generated locally), \
+            populate `network.boot_nodes` with at least one known node from the chain"
```
RomanBrodetski: nice! Would it make sense to add a link to this PR as well? 🤔

RomanBrodetski (dismissed):
Looks good!
Would be great if @popzxc could also review as he was following this whole initiative 🙏

node/bin/src/config/mod.rs
```diff
@@ -236,29 +235,23 @@ pub struct NetworkConfig {
...
+    /// The node's secret key (256-bit ECDSA), from which the node's identity is derived. Used during
+    /// initial RLPx handshake.
     #[config(secret)]
-    #[config(
-        default_t = SecretKey::from_str("21b0ee131240821c39627c39d0fdde5edbda968c5877f5b63c5c542f267b5349").unwrap(),
-        with = Serde![str]
-    )]
-    pub secret_key: SecretKey,
+    #[config(default, with = Serde![str])]
+    pub secret_key: Option<SecretKey>,
```
RomanBrodetski: could we use the same key for 2FA?

itegulov: Yes, we could reuse the same key as it's the same crypto scheme. But I am not sure whether it is safe from cryptography perspective to use the same key for handshake and also for signing batches. Will need to check with someone later, I'd keep them separate for now

docs/src/setup/external_node.md
```diff
@@ -1,16 +1,21 @@
 # External node
 
-Setting the `block_replay_download_address` environment variable puts the node in external node mode, which means it
+Setting the `general_node_role=ExternalNode` environment variable puts the node in external node mode, which means it
 receives block replays from another node instead of producing its own blocks. The node will get priority transactions
 from L1 and check that they match the ones in the replay but it won't change L1 state.
 
-To run the external node locally, you need to set its services' ports so they don't overlap with the main node.
+To run the external node locally, you need to enable networking on both main node and external node. Then set external node's services' ports so they don't overlap with the main node.
```
popzxc: Bikeshedding: line length is broken now. (Though not insist on fixing, as such stuff should be caught by linter/formatter which we don't have I guess).

docs/src/setup/external_node.md
```diff
@@ -1,16 +1,21 @@
 # External node
 
-Setting the `block_replay_download_address` environment variable puts the node in external node mode, which means it
+Setting the `general_node_role=ExternalNode` environment variable puts the node in external node mode, which means it
```
popzxc: Nit: maybe slap `rename_all(snake_case)` on the `NodeRole` so that this setting looks less weird?

node/bin/src/lib.rs
```diff
@@ -345,6 +318,9 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
 
     // `starting_block` - the block number to go through the pipeline.
     let starting_block = if node_startup_state.l1_state.last_committed_batch > 0 {
+        // todo: ideally this should be searched through p2p networking instead of RPC
```
popzxc: Not sure if you currently create issues for TODOs, but if you do -- don't forget to do it.

node/bin/src/replay_transport.rs
popzxc: Hell yes!

popzxc (approved):
Overall, great job! Left some nits, but don't think that they're worth blocking the merge; will be fine with merging as-is.

node/bin/src/main.rs
```diff
@@ -350,23 +349,24 @@ fn build_external_config(repo: ConfigRepository<'_>) -> Config {
...
-                &mut sequencer_config,
-                &mut rpc_config,
-                &mut l1_sender_config,
-                &mut genesis_config,
-                &mut prover_api_config,
-                &mut observability_config,
-            )
-            .unwrap_or_else(|_| panic!("Failed to load zkstack config from `{config_dir}`: "));
-    }
+    // todo: is this relevant anymore?
```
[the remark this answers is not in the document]
itegulov: Completely forgot to raise this. 99% chance it is irrelevant, will return back if someone complains

popzxc (approved):

PR #886, opened 2026-02-17
feat: some gateway features

perekopskiy:
## Summary

Features needed for server to settle on Gateway.
- Splits `l1_provider` into `l1_provider` and  `sl_provider`, those are now 2 different entities. All components are updated correspondingly to accept need provider(s).
- Adds `RelayedL2Calldata` pubdata mode.
- Updates RPC to server l2 to l1 logs proofs for chains that settle on gateway.
- Prover calculation for aggregated root starting from v31.

### DIFF
--- a/lib/batch_verification/src/tests/mod.rs
+++ b/lib/batch_verification/src/tests/mod.rs
@@ -81,6 +81,9 @@ pub fn dummy_batch_metadata(batch_number: u64, from: u64, to: u64) -> BatchMetad
         execution_version: 1,
         protocol_version: ProtocolSemanticVersion::legacy_genesis_version(),
         computational_native_used: None,
+        logs: vec![],
+        messages: vec![],
+        multichain_batch_root: Default::default(),
     }
 }
 
--- a/lib/contract_interface/src/lib.rs
+++ b/lib/contract_interface/src/lib.rs
@@ -290,27 +290,36 @@ alloy::sol! {
             bytes calldata _commitData
         ) external;
 
-       function proofPayload(StoredBatchInfo old, StoredBatchInfo[] newInfo, uint256[] proof);
+        function proofPayload(StoredBatchInfo old, StoredBatchInfo[] newInfo, uint256[] proof);
 
-       function proveBatchesSharedBridge(
+        function proveBatchesSharedBridge(
             address _chainAddress,
             uint256 _processBatchFrom,
             uint256 _processBatchTo,
             bytes calldata _proofData
-       );
+        );
 
-       struct PriorityOpsBatchInfo {
-           bytes32[] leftPath;
-           bytes32[] rightPath;
-           bytes32[] itemHashes;
+        struct PriorityOpsBatchInfo {
+            bytes32[] leftPath;
+            bytes32[] rightPath;
+            bytes32[] itemHashes;
+        }
+
+        struct L2Log {
+           uint8 l2ShardId;
+           bool isService;
+           uint16 txNumberInBatch;
+           address sender;
+           bytes32 key;
+           bytes32 value;
        }
 
-       function executeBatchesSharedBridge(
-           address _chainAddress,
-           uint256 _processFrom,
-           uint256 _processTo,
-           bytes calldata _executeData
-       );
+        function executeBatchesSharedBridge(
+            address _chainAddress,
+            uint256 _processFrom,
+            uint256 _processTo,
+            bytes calldata _executeData
+        );
     }
 
     // taken from v29 version of `IExecutor.sol`
--- a/lib/contract_interface/src/models.rs
+++ b/lib/contract_interface/src/models.rs
@@ -1,5 +1,5 @@
 use crate::{IExecutor, IExecutorV29, IExecutorV30};
-use alloy::primitives::{B256, Bytes, U256, keccak256};
+use alloy::primitives::{Address, B256, Bytes, U256, keccak256};
 use alloy::sol_types::SolValue;
 use serde::{Deserialize, Serialize};
 use std::fmt;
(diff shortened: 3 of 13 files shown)


lib/gas_adjuster/src/lib.rs
```diff
@@ -23,11 +27,12 @@ mod statistics;
 pub struct GasAdjuster {
     base_fee_statistics: GasStatistics<u128>,
     blob_base_fee_statistics: GasStatistics<u128>,
+    l2_pubdata_price_statistics: GasStatistics<U256>,
```
itegulov: Nit: consider naming it as `sl_pubdata_price_statistics` (or even `gateway_`?) as by `l2_` we usually refer to the chain that is running on this node

lib/rpc_api/src/types.rs
```diff
@@ -55,6 +55,15 @@ pub struct BlockMetadata {
     pub execution_version: u32,
 }
 
+/// Extended FeeHistory struct including L2 pubdata price history.
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct L2FeeHistory {
+    #[serde(flatten)]
+    pub base: FeeHistory,
+    pub l2_pubdata_price: Option<Vec<U256>>,
```
itegulov: Nit: similar thing here, maybe `pubdata_price_per_byte` to be consistent with `BlockMetadata` above?

lib/l1_watcher/src/tx_watcher.rs
```diff
@@ -77,22 +84,44 @@ impl ProcessL1Event for L1TxWatcher {
...
-            self.l1_subpool.insert(Arc::new(tx)).await;
+            let processable = if let Some(total_priority_ops) = self.cached_total_priority_ops_resp
+                && total_priority_ops > tx.priority_id()
+            {
+                true
+            } else {
+                let total_priority_ops = self
+                    .zk_chain_sl
+                    .get_total_priority_txs_at_block(BlockId::Number(BlockNumberOrTag::Latest))
+                    .await?;
```
itegulov: General idea here was that `process_event` can defer until processing is possible. So you'd sleep-retry here instead of returning `false` (or even better subscribe to blocks but we are far away from this). I think your approach is buggy because it can process the same event twice if there is a block with multiple events and one of them is not processable.

lib/rpc/src/zks_impl.rs
```diff
@@ -1,41 +1,48 @@
...
 const LOG_PROOF_SUPPORTED_METADATA_VERSION: u8 = 1;
 
 pub struct ZksNamespace<RpcStorage> {
     bridgehub_address: Address,
     bytecode_supplier_address: Address,
-    committed_batch_provider: CommittedBatchProvider,
     storage: RpcStorage,
     genesis_input_source: Arc<dyn GenesisInputSource>,
+    l2_chain_id: u64,
+    gateway_provider: Option<DynProvider>,
```
itegulov: Sometimes we use `sl_provider`, sometimes `gateway_provider`. Do you think it makes sense to make something like this:
```rust
struct MultiProvider {
    pub l1_provider: DynProvider<Ethereum>,
    // can be Ethereum for now but custom Network would allow us to convinently call our own methods instead of doing raw calls
    pub gateway_provider: Option<DynProvider<ZksyncOs>>,
}
```
And it would have convenience method `sl_provider()` that returns either `l1_provider` or `gateway_provider` depending on whether gateway provider is present or not.

lib/storage/src/db/batch.rs
```diff
@@ -131,6 +145,26 @@ impl ReadBatch for ExecutedBatchStorage {
         serde_json::from_slice(&bytes).context("failed to deserialize context")
     }
 
+    fn get_execute_sl_block_number_by_batch_number(
+        &self,
+        batch_number: u64,
+    ) -> anyhow::Result<Option<u64>> {
```
itegulov: Let's make it a part of `DiscoveredCommittedBatch`, no?

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -186,8 +179,64 @@ impl<BatchStorage: WriteBatch, Finality: WriteFinality> ProcessL1Event
...
+    for L1PersistBatchWatcher<BatchStorage, Finality>
+{
+    fn name(&self) -> &'static str {
+        "persist_batch"
+    }
+
+    fn event_signatures(&self) -> Topic {
+        Topic::default()
+            .extend(ReportCommittedBatchRangeZKsyncOS::SIGNATURE_HASH)
+            .extend(BlockExecution::SIGNATURE_HASH)
```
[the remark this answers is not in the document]
itegulov: Ah, I see. You want to track both events because you need `execute_sl_block_number`

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -186,8 +179,64 @@ impl<BatchStorage: WriteBatch, Finality: WriteFinality> ProcessL1Event
...
+                let execute = BlockExecution::decode_log(&log.inner)?.data;
+                let batch_number = execute.batchNumber.to::<u64>();
+                let batch_hash = execute.batchHash;
+                if let Some(committed_batch) = self.committed_batches.remove(&batch_number) {
+                    tracing::debug!(
+                        batch_number,
+                        ?batch_hash,
+                        "discovered executed batch, persisting"
+                    );
+                    self.batch_storage.write(
```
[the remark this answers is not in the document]
perekopskiy: I think we should keep waiting for finality, because the processing of `execute` events works under the assumption that the batch is 100% finalized and in-memory commit info is accurate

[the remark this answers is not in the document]
perekopskiy: Currently yes, but at some point we should support L1 finality and wait here for it

PR #904, opened 2026-02-24
feat(merkle-tree): Implement storage proofs for `zks_getProof`

slowli:
## Summary

Changes / extends the Merkle tree API in preparation for `zks_getProof` implementation. The main changes are as follows:

- The public tree API is extracted into a separate crate (to be used in the future by `storage_api`).
- A proof format compatible with `zks_getProof` is added. In this format, each proven slot has a separate Merkle path (or two paths for missing reads). Previously implemented batch proofs had joint intermediate tree hashes for all proven slots, which is space- and compute-efficient, but is harder to reason about.


antoniolocascio (dismissed):
Generally looks good to me. As discussed, we should close the one in zksync-os and just keep this one. If we do, you can remove the TODO in flat.rs too

PR #918, opened 2026-02-26
fix: Compare block hash during block replay

EmilLuta:
Current block replay struct comparison includes all fields (such as metadata, like `NodeVersion`. This means that any node sent over the wire on old node version will fail comparison (if it was not also committed) after Main Node reboot. This is exactly what happened on stage-rollup (that is currently crash looping).

```bash
# EN have:
node_version: Version { major: 0, minor: 14, patch: 2 }
# But are receiving from MN:
node_version: Version { major: 0, minor: 15, patch: 1 }
```

As such, comparison fails, even though the block is exactly the same.

This PR checks whether the block is the same or not by comparing `block_output_hash` alone.

### DIFF
--- a/lib/storage/src/db/replay.rs
+++ b/lib/storage/src/db/replay.rs
@@ -428,7 +428,7 @@ impl WriteReplay for BlockReplayStorage {
             let old_record = self
                 .get_replay_record(block_context.block_number)
                 .expect("Old record must exist");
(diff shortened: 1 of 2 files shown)


perekopskiy (approved):
Would be nice to enforce that `PartialEq` is updated when we add new field to `ReplayRecord` but I don't see an easy way to do it

PR #922, opened 2026-02-27
fix: execute works on gateway

AntonD3:


itegulov:
Included as a part of #886

PR #953, opened 2026-03-09
feat: consensus integration 1/5: Sequencer split in BlockExecutor and BlockApplier

RomanBrodetski:
This prepares the node for the consensus integration. 
There are no changes to node or network behavior.  
Main Node wiring is modified:
* Split Sequencer into `BlockExecutor` (executes, updates mempools) and `BlockApplier` (saves result to the DB)
* Introduce BlockCanonizer between them:
```
BlockExecutor ~> BlockCanonizer ~> BlockApplier
```
`BlockCanonizer` will wait for the proposed blocks to be canonized by consensus before advancing them in the pipeline - this behavior is stubbed for now. 
* Modify the `MainNodeCommandSource` so that it can replay blocks recieved from consensus. This behaviour is not currently used (as we don't have consensus yet - ENs are still using the `ENCommandSource`)
* Additionally introduces `OverlayBuffer` to let `BlockExecutor` execute ahead of persisted state by layering uncommitted block diffs on top of  the last persisted block.


lib/storage_api/src/overlay_buffer.rs
```diff
@@ -0,0 +1,133 @@
...
+
+#[derive(Debug, Clone)]
+pub struct BlockOverlay {
+    pub storage_writes: Vec<StorageWrite>,
+    pub preimages: Vec<(B256, Vec<u8>)>,
+}
+
+#[derive(Debug, Default)]
+pub struct OverlayBuffer {
+    overlays: BTreeMap<u64, BlockOverlay>,
```
itegulov: Nit: `u64` -> `BlockNumber` here and in other places in this file for (subjectively) better context

lib/storage_api/src/overlay_buffer.rs
```diff
@@ -0,0 +1,133 @@
...
+        );
+        self.purge_already_persisted_blocks(base_latest)?;
+        let first_overlay = self.overlays.keys().next().copied();
+        let last_overlay = self.overlays.keys().next_back().copied();
+        tracing::debug!(
+            "Synced overlay buffer with base (base_latest={base_latest}, overlays_len={}, overlays_range={first_overlay:?}..={last_overlay:?}). \
+            Preparing storage view to execute block {block_number_to_execute}.",
+            self.overlays.len(),
+        );
+        if base_latest >= block_number_to_execute - 1 {
```
itegulov: `block_number_to_execute` cannot be 0 here right?

lib/sequencer/src/execution/mod.rs
```diff
@@ -1,86 +1,88 @@
...
 
-/// Sequencer pipeline component
-/// Contains all the dependencies needed to run the sequencer
-pub struct Sequencer<Subpool, State, Replay, Repo>
+pub use block_applier::BlockApplier;
+pub use block_canonizer::{BlockCanonization, BlockCanonizer, NoopCanonization};
+/// Executes blocks, while only updating local in-memory state (mempool, block context).
+/// Does not persist anything to disk.
+/// Does not track the node role - reacts on the ordered inbound commands instead (`Produce` vs `Replay`)
+pub struct BlockExecutor<Subpool, State>
```
itegulov: Nit: does this belong in a separate file like `BlockApplier` and `BlockCanonizer`?

lib/sequencer/src/execution/mod.rs
```diff
@@ -1,86 +1,88 @@
...
     type Input = BlockCommand;
-    type Output = (BlockOutput, ReplayRecord);
+    /// Outputs executed blocks. Passes along information whether it's a replayed or new block -
+    ///  new blocks need to be canonized by network (enforced by `BlockCanonizer`)
+    type Output = (BlockOutput, ReplayRecord, BlockCommandType);
 
-    const NAME: &'static str = "sequencer";
-    const OUTPUT_BUFFER_SIZE: usize = 5;
+    const NAME: &'static str = "block_executor";
+    const OUTPUT_BUFFER_SIZE: usize = 1;
```
itegulov: Is this set to 1 intentionally? It only gives us room to execute one extra block on top, right?

lib/sequencer/src/execution/block_canonizer.rs
```diff
@@ -0,0 +1,162 @@
...
+
+#[async_trait]
+impl<Consensus> PipelineComponent for BlockCanonizer<Consensus>
+where
+    Consensus: BlockCanonization,
+{
+    // Input from BlockExecutor
+    type Input = (BlockOutput, ReplayRecord, BlockCommandType);
+    // Output to BlockApplier
+    type Output = (BlockOutput, ReplayRecord, BlockCommandType);
```
itegulov: Nit: I'd introduce structs as these tuples are getting a bit large already. Also, some type safety would not hurt anyway, so we can distinguish canonized blocks from non-canonized blocks, otherwise you could skip `BlockCanonizer` in the pipeline accidentally, right?

lib/sequencer/src/execution/block_canonizer.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    const NAME: &'static str = "block_canonizer";
+    const OUTPUT_BUFFER_SIZE: usize = 2;
```
itegulov: What was the thought process behind putting 2 here? Please leave some comments here in other similar places

lib/sequencer/src/execution/block_applier.rs
```diff
@@ -0,0 +1,81 @@
...
+                block_output.storage_writes.clone(),
+                block_output
+                    .published_preimages
+                    .iter()
+                    .map(|(k, v)| (*k, v)),
+                override_allowed,
+            )?;
+
+            self.repositories
+                .populate(block_output.clone(), executed_replay.transactions.clone())
```
itegulov: Returning to what we discussed during the meeting with "soft-writes" to repositories for RPC purposes: I think our current design is already very compatible with this idea. Repositories already hold some in-memory state, we just need to maintain one extra variable `latest_persistable_block` and disallow repositories to persist past it in the background. Then `populate` can be moved to `BlockExecutor` and here we'd just bump `latest_persistable_block`.

lib/sequencer/src/execution/block_canonizer.rs
RomanBrodetski: Coincidently we actually skip it on purpose in ENs... But indeed can be made more explicit with smt like `NoOpCanonizer: ExecutedBlock -> CanonizedBlock`.  

As for types/structs - I agree. I'll do a separate PR with this change (will even consider doing that against `main`)

node/bin/src/command_source.rs
```diff
@@ -37,31 +40,119 @@ impl<Replay: ReadReplay> PipelineComponent for MainNodeCommandSource<Replay> {
...
+                send_res = output.send(BlockCommand::Produce(ProduceCommand)) => {
+                    if send_res.is_err() {
+                        tracing::warn!("Command output channel closed, stopping source");
+                        break;
+                    }
+                }
+            }
+        }
+
+        anyhow::bail!("Execution loop in MainNodeCommandSource ended unexpectedly");
```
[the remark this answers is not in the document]
RomanBrodetski: this logic changed in one of the downstream PRs - will leave as is for now, and make sure it's not present downstream

romanbrodetski-ai:
## Block-Rebuild Maintainer Review

### 🔴 High severity — `NoopCanonization` deadlock when two produced/rebuild blocks are in flight simultaneously

**File:** `lib/sequencer/src/execution/block_canonizer.rs`

`NoopCanonization::new()` creates a channel of capacity **1**, but `MAX_PRODUCED_QUEUE_SIZE` inside `BlockCanonizer::run` is **2**. These two constants are inconsistent, and the inconsistency produces a silent async deadlock in the main production loop (and during startup rebuilds).

**How the deadlock occurs:**

Inside `BlockCanonizer::run`, when a `Produce` or `Rebuild` command arrives, the code does:

```rust
self.consensus.propose(proposed).await?;   // (1) send to NoopCanonization channel (cap 1)
produced_queue.push_back(...);             // (2)
```

After step (2), `produced_queue.len() == 1 < MAX_PRODUCED_QUEUE_SIZE (2)`, so the `maybe_executed` arm remains enabled in the `select!`. If `BlockExecutor` has a second block ready in its output channel and `tokio::select!` picks `maybe_executed` before `canonized`:

- `propose(block2).await` tries to send to the NoopCanonization channel — which is already full (capacity 1, block1 still there).
- The task suspends inside `propose`, waiting for capacity.
- The only way to drain the channel is `self.consensus.next_canonized()` in the `canonized` arm — but that arm is unreachable because the task is blocked inside `propose`.
- **Result: the task never makes progress again.**

This affects both startup rebuild blocks (processed quickly with no mempool delay) and produce blocks under any non-trivial mempool load. Replay blocks are unaffected (they bypass `propose`).

**Fix:** increase `NoopCanonization`'s channel capacity to at least `MAX_PRODUCED_QUEUE_SIZE`:

```rust
// block_canonizer.rs
impl NoopCanonization {
    pub fn new() -> Self {
        let (sender, receiver) = mpsc::channel(2); // must be >= MAX_PRODUCED_QUEUE_SIZE
        Self { sender, receiver }
    }
}
```

**Test confirming the bug** (added to `block-rebuild-maintainer/tests/tests/rebuild.rs` in the agents repo, commit `9ac9f38`):

```rust
#[tokio::test]
async fn noop_canonization_channel_supports_max_produced_queue_size_proposals() {
    let noop = NoopCanonization::new();
    // Propose 2 blocks without draining. If channel capacity < 2, propose(block2) deadlocks.
    let result = tokio::time::timeout(Duration::from_secs(5), async {
        noop.propose(replay_record(1, 0, vec![])).await.unwrap();
        noop.propose(replay_record(2, 0, vec![])).await.unwrap();
    }).await;
    result.expect("NoopCanonization channel capacity (1) < MAX_PRODUCED_QUEUE_SIZE (2): ...");
}
```

Fails (5 s timeout) on the PR code; passes after `mpsc::channel(1)` → `mpsc::channel(2)`.

---

*Also noted (non-blocking):*

- **Typo** in the new `Replay` ordering assertion (`block_context_provider.rs`): `"blocks received our of order"` → `"out of order"`.
- **No ordering check for `Rebuild` commands** in `block_context_provider.rs` — the `Replay` branch gained `anyhow::ensure!(self.next_block_number == record.block_context.block_number, ...)` but the `Rebuild` branch did not. Out-of-order rebuilds would go undetected. Knowledge invariant: *replay/rebuild ordering bugs are high severity*.

PR #956, opened 2026-03-09
feat: Adding operator signing with HSM

vladbochok:
## Summary

Currently, operator private keys are stored in the plaintext in the server. There is a better way to manage secrets; we can store keys in Google HSM. Google HSM generates the secret and never exposes it externally. All signing is done through API requests. 

This PR adds an option to load operator keys from Google controlled HSM and sign batches that way. 

Please note, I have tested that server works locally with Google KMS, but the code was written by Claude so be skeptical.

### DIFF
--- a/lib/operator_signer/src/gcp.rs
+++ b/lib/operator_signer/src/gcp.rs
@@ -25,14 +25,12 @@ pub(crate) async fn create_gcp_signer(resource_name: &str) -> anyhow::Result<Gcp
     .await
     .context("failed to create GCP KMS client")?;
 
-    GcpSigner::new(client, specifier, None)
-        .await
-        .map_err(|e| {
-            anyhow::anyhow!(
-                "failed to initialize GCP KMS signer for '{resource_name}': {e}. \
+    GcpSigner::new(client, specifier, None).await.map_err(|e| {
+        anyhow::anyhow!(
+            "failed to initialize GCP KMS signer for '{resource_name}': {e}. \
                  Ensure the key uses EC_SIGN_SECP256K1_SHA256 algorithm"
-            )
-        })
+        )
+    })
 }
 
 /// Parsed components of a KMS resource name.


lib/operator_signer/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+            Self::Local(_) => f.debug_tuple("Local").field(&"[REDACTED]").finish(),
+            Self::GcpKms { resource_name } => f
+                .debug_struct("GcpKms")
+                .field("resource_name", resource_name)
+                .finish(),
+        }
+    }
+}
+
+impl PartialEq for OperatorSignerConfig {
```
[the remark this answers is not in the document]
itegulov: I think because we check that commit/prove/execute operators are different. I think it makes more sense to check that by comparing `Signer::address` instead. Then we'll even support cross-variant comparison

node/bin/src/config/util.rs
```diff
@@ -6,30 +6,70 @@ use smart_config::ErrorWithOrigin;
...
+            )),
+        }
     }
 
-    fn serialize_param(&self, param: &SigningKey) -> Value {
-        let bytes = B256::from_slice(param.to_bytes().as_slice());
-        serde_json::to_value(bytes).expect("failed serializing to JSON")
+    fn serialize_param(&self, param: &OperatorSignerConfig) -> Value {
+        match param {
+            OperatorSignerConfig::Local(_) => Value::String("[REDACTED]".to_string()),
```
itegulov: I think you might not need to this if you mark all usages of `OperatorSignerConfig` with `#[config(secret)]`. But @slowli would know for sure

lib/operator_signer/src/gcp.rs
```diff
@@ -0,0 +1,103 @@
...
+    let keyring = GcpKeyRingRef::new(&parts.project_id, &parts.location, &parts.keyring_name);
+    let specifier = KeySpecifier::new(keyring, &parts.key_id, parts.version);
+
+    let client = GoogleApi::from_function(
+        KeyManagementServiceClient::new,
+        "https://cloudkms.googleapis.com",
+        None,
+    )
+    .await
+    .map_err(|e| anyhow::anyhow!("failed to create GCP KMS client: {e}"))?;
```
slowli: Nit: It would be more idiomatic to use `context("failed to create GCP KMS client")` here; this would structure the error as appropriate. Ditto below; `context()` could be used instead of `map_err()`.

lib/operator_signer/src/gcp.rs
```diff
@@ -0,0 +1,103 @@
...
+fn parse_kms_resource_name(resource_name: &str) -> anyhow::Result<KmsResourceParts> {
+    let parts: Vec<&str> = resource_name.split('/').collect();
+    if parts.len() != 10
+        || parts[0] != "projects"
+        || parts[2] != "locations"
+        || parts[4] != "keyRings"
+        || parts[6] != "cryptoKeys"
+        || parts[8] != "cryptoKeyVersions"
+    {
+        anyhow::bail!(
```
slowli: Nit: Can use `anyhow::ensure!(parts.len() == 10 && ..., "...")` here to be more idiomatic.

node/bin/src/config/util.rs
```diff
@@ -6,30 +6,70 @@ use smart_config::ErrorWithOrigin;
...
         ctx: DeserializeContext<'_>,
         param: &'static ParamMetadata,
-    ) -> Result<SigningKey, ErrorWithOrigin> {
+    ) -> Result<OperatorSignerConfig, ErrorWithOrigin> {
         let deserializer = ctx.current_value_deserializer(param.name)?;
+        let value = Value::deserialize(deserializer)?;
 
-        let b256 = B256::deserialize(deserializer)?;
-        SigningKey::from_slice(b256.as_slice()).map_err(ErrorWithOrigin::custom)
+        match value {
```
slowli: Wouldn't it be possible to replace this logic with an enum with `derive(Serialize, Deserialize)` + `#[serde(tag = "type")]` + `#[serde(untagged)]` [on the default enum variant](https://serde.rs/variant-attrs.html#untagged)? And then use `serde![str, object]` deserializer for `smart-config`.

node/bin/src/config/util.rs
slowli: In general `serialize_param` is used for canonicalizing configs etc., so it's expected to not be lossy (it should be possible to deserialize a param back from the serialized representation). It *is* also used to display config values, but the config params marked as `#[config(secret)]` are indeed redacted automatically based on this marker.

lib/operator_signer/src/gcp.rs
hatemosphere: can we keep .map_err() in second call? it includes resource_name and algorithm hint in the message. .context() only prepends static text.

lib/operator_signer/src/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+
+mod gcp;
+
+/// Configuration for how an operator signing key is provided.
+///
+/// For GCP KMS keys, the signer (and its underlying API client) is created lazily
+/// on first use and cached for subsequent calls. Cloned configs share the same cache
+/// via `Arc`, so multiple calls to [`address`](Self::address) and
+/// [`register_with_wallet`](Self::register_with_wallet) only create one GCP client.
+pub enum OperatorSignerConfig {
```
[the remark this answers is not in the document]
hatemosphere: i renamed

PR #957, opened 2026-03-09
feat: scale eth_gasPrice by configurable factor

perekopskiy-ai:
## Summary

Scale `eth_gasPrice` by a configurable RPC factor instead of returning the raw base fee.

This adds `rpc.gas_price_scale_factor` with a default value of `1.5`, applies it to the base fee returned by `eth_gasPrice`, and adds unit / integration coverage for the new behavior.

### DIFF
--- a/node/bin/src/config/mod.rs
+++ b/node/bin/src/config/mod.rs
@@ -427,6 +427,7 @@ pub struct RpcConfig {
     pub send_raw_transaction_sync_timeout: Duration,
 
     /// Factor applied to the pending block base fee returned by `eth_gasPrice`.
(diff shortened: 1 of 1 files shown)


node/bin/src/config/mod.rs
```diff
@@ -426,6 +426,10 @@ pub struct RpcConfig {
     #[config(default_t = 2 * TimeUnit::Seconds)]
     pub send_raw_transaction_sync_timeout: Duration,
 
+    /// Factor applied to the pending block base fee returned by `eth_gasPrice`.
```
itegulov: Let's add a comment that this is important for MetaMask to provide context for the future.

PR #958, opened 2026-03-09
feat: consensus integration 2/5: Consensus interface, raft dependency

RomanBrodetski:
## Description
  This PR extracts the first, minimal slice of the consensus work.

  It introduces a shared consensus runtime boundary in zksync_os_raft and updates node wiring to use it, but keeps behavior single-node:

  - explicit consensus runtime parts
  - shared ConsensusRole / LeadershipSignal
  - loopback consensus initialization
  - local canonization / always-leader behavior
  - it also makes the batcher subsystem optional - but it's of course running by default

  What is intentionally not included yet:

  - OpenRaft transport over the network
  - cluster bootstrap / multi-node behavior
  - broader config, status, and test rollout

  The goal is to make the node consensus-aware first, without mixing that review with distributed networking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14444,6 +14444,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "zksync_os_raft"
+version = "0.16.0"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "futures",
+ "tokio",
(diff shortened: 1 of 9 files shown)


itegulov (approved):
Hard to judge new abstractions yet but LGTM in general

PR #966, opened 2026-03-10
fix(tests): decompress L1 state in build.rs instead of in-process cache

tomg10-claude:
## Summary

- **Revert the in-process `LazyLock` cache** for decompressed L1 state — it was ineffective because each test runs in a separate process and doesn't share static state
- **Move decompression to `build.rs`** so it happens once at build time; a SHA-256 hash sidecar file (`l1-state.json.sha256`) ensures re-decompression only when the `.gz` file changes
- **Simplify `config.rs`**: `l1_state()` now just reads the pre-decompressed `.json` file, removing `flate2`/`HashMap`/`Mutex` from runtime dependencies
- **Add CI step** to decompress L1 state fixtures on the GPU runner before prover tests (the nextest archive only contains binaries, not build.rs outputs)
- **Add `.gitignore` entries** for the generated `l1-state.json` and `.sha256` files

## Task prompt

> Every test is run in a separate process so they do not share static state. Could you please revert this?
> Is there a different way to decompress only once?
> Put this step into build.rs. Only build if the .gz file changed. Put the hash of the file alongside the decompressed state.

## Test plan

- [x] `cargo fmt --all --check` passes
- [x] `cargo clippy -p zksync_os_integration_tests --all-targets -- -D warnings` passes
- [x] `cargo build -p zksync_os_integration_tests` produces decompressed files on first run
- [x] Subsequent builds skip decompression (hash matches)
- [ ] `cargo nextest run -p zksync_os_integration_tests` — all integration tests pass
- [ ] Prover tests pass in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -234,10 +234,10 @@ jobs:
 
       - name: Decompress L1 state fixtures
         run: |
-          for gz in local-chains/*/l1-state.json.gz; do
-            json="${gz%.gz}"
+          for gz_file in local-chains/*/l1-state.json.gz; do
+            json="${gz_file%.gz}"
             if [ ! -f "$json" ]; then
-              gunzip -k "$gz"
+              gunzip -k "$gz_file"
             fi
           done
 


integration-tests/build.rs
```diff
@@ -1,8 +1,75 @@
...
+    let Ok(entries) = std::fs::read_dir(&local_chains) else {
+        return;
+    };
+
+    for entry in entries.flatten() {
+        let gz_path = entry.path().join("l1-state.json.gz");
+        println!("cargo::rerun-if-changed={}", gz_path.display());
+
+        if !gz_path.is_file() {
+            continue;
```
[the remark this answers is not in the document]
itegulov: Ah, I think I understand. You are relying on this to distinguish between real entries (e.g. `v30.2/l1-state.json.gz`) vs others (`README.md/l1-state.json.gz`). I think it might be better to be defensive and filter for directories, then assert existence.

integration-tests/build.rs
```diff
@@ -1,8 +1,75 @@
...
+        let hash = Sha256::digest(&compressed);
+        let hex_hash = format!("{hash:x}");
+
+        // The decompressed JSON lives next to the .gz file.
+        let json_path = gz_path.with_extension(""); // strips .gz
+        let hash_path = json_path.with_extension("json.sha256");
+
+        // Skip decompression if the hash file matches.
+        if let Ok(existing_hash) = std::fs::read_to_string(&hash_path)
+            && existing_hash.trim() == hex_hash
```
itegulov: You can break current logic by accidentally deleting `l1-state.json` but leaving `l1-state.json.sha256`, no?

tomg10: fixed

PR #968, opened 2026-03-10
feat: add gateway interop fee updater

perekopskiy:
## Summary

- add a gateway-only interop fee updater crate and wiring
- add SetInteropFee system tx support and a dedicated mempool subpool
- adds ID number for new system tx type, so it can be used as salt, number is stored as a part of replay record

### DIFF
--- a/integration-tests/tests/interop.rs
+++ b/integration-tests/tests/interop.rs
@@ -319,7 +319,6 @@ async fn fund_wallet_via_l1_deposit(tester: &Tester, wallet: Address, amount: U2
         }))
         .await?;
     let max_fee_per_gas = base_l1_fees_data.max_fee_per_gas + max_priority_fee_per_gas;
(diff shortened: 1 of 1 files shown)


lib/interop_fee_updater/src/lib.rs
```diff
@@ -0,0 +1,230 @@
...
+use zksync_os_types::TokenPricesForFees;
+
+#[derive(Debug, Clone)]
+pub struct InteropFeeUpdaterConfig {
+    pub polling_interval: std::time::Duration,
+    pub update_deviation_percentage: u32,
+}
+
+const GW_ASSET_TRACKER_ADDRESS: Address = address!("0x0000000000000000000000000000000000010010");
+const L2_INTEROP_CENTER_ADDRESS: Address = address!("0x000000000000000000000000000000000001000d");
```
Artemka374: You define the same value in `types/src/transaction/system/utils` - I think you should just import and reuse it here

PR #1010, opened 2026-03-16
feat: Sync l1 state with draft-v31

Artemka374:
## Summary                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                   
  - **`ProvingVersion::V7`**: Maps protocol versions `v31.x` and `v32.x` to `ProvingVersion::V7`. The build system maps the `dev-20260311` git tag to V7 for binary resolution. The prover input generator gains a V7 
  execution path using the dev variants of `zk_ee` and `forward_system`.
                                                                                                                                                                                                                                                   
  - **L1 sender — `settlementFeePayer` in execute calldata**: For v31+ protocol versions with gateway enabled, the `executeBatchesSharedBridge` calldata now includes `settlementFeePayer`.
                                                                                                                                                                                                                                                   
  - **Gateway migration watcher refactor**: Removes the generic `MigrationProcessor` trait-based design. The watcher is now a concrete type handling both `MigrateToGateway` and `MigrateFromGateway` events in one place, using                   
  `l1_chain_id`/`gw_chain_id` split to assign the correct settlement chain ID per event direction. The `enable_gw_migration_watcher` config flag is removed — the watcher is always on when protocol version is v31 or highrt.
                                                                                                                                                                                                                                                   
  - **Config**: Add a `gateway_chain_id` field (default 506) used by the migration watcher. The correct value should be set before the migration to gateway.
  
  - **Interop integration tests**: Unignore and fix `interop_bundle_send` test.
                                                                                                                                                                                                                                                   
  - **Local chain state**: Regenerated v31 state from the `draft-v3`1 contracts.

### DIFF
--- a/node/bin/src/lib.rs
+++ b/node/bin/src/lib.rs
@@ -530,7 +530,7 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
         if config.general_config.gateway_rpc_url.is_some() {
             tasks.spawn(
                 InteropWatcher::create_watcher(
-                    node_startup_state.l1_state.bridgehub_sl.clone(), // TODO: what bridgehub to use here?
+                    node_startup_state.l1_state.bridgehub_sl.clone(),
                     config.l1_watcher_config.clone().into(),
                     next_interop_event_index.clone(),
                     interop_roots_subpool.clone(),


local-chains/v31.0/versions.yaml
```diff
@@ -17,4 +17,4 @@ zkstack-cli:
 zksync-os-scripts:
   # TODO: remove branch and put tag and sha when the release is done
 #  tag: "TBD"
-  sha: "e377376e1b45697d04889ab439e4ad13bd406eed"
+  sha: "544c8c31a6e5e76ffb307f6c921596ca920b7fc6"
```
itegulov: Just double-checking: these SHAs are up-to-date, correct? I updated them last time I generated state, not sure if you re-generated it after or not

node/bin/src/config/mod.rs
```diff
@@ -187,6 +187,11 @@ pub struct GeneralConfig {
     /// Currently, it's a marker of whether chain settles to Gateway or not.
     pub gateway_rpc_url: Option<String>,
 
+    /// Gateway chain ID. Used by the migration watcher to construct `SetSLChainId` system
+    /// transactions when a `MigrateToGateway` event fires. Defaults to 506 (ZKsync Gateway).
+    #[config(default_t = 506)]
+    pub gateway_chain_id: u64,
```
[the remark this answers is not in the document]
itegulov: > Currently, it's a marker of whether chain settles to Gateway or not.

Ok, _this_ definitely should not be the case long-term. We'll need gateway RPC to be present before migration is complete. So generally I agree with @perekopskiy

lib/l1_watcher/src/gateway_migration_watcher.rs
```diff
@@ -90,38 +67,87 @@ impl<T: MigrationProcessor> GatewayMigrationWatcher<T> {
...
+    async fn process_raw_event(&mut self, log: Log) -> Result<(), L1WatcherError> {
+        let Some(&topic0) = log.topic0() else {
+            return Ok(());
+        };
+
+        let (new_sl_chain_id, migration_number) = match topic0 {
+            MigrateToGateway::SIGNATURE_HASH => {
+                let event = MigrateToGateway::decode_log(&log.inner)?.data;
+                let migration_number: u64 = event.migrationNumber.try_into().unwrap();
+                (self.gw_chain_id, migration_number)
```
itegulov: Hmm, is there no way to get gw chain id from L1 somehow here? I'd expect it to be a part of the event but evidently not.

node/bin/src/lib.rs
```diff
@@ -509,56 +509,36 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
-                    config.l1_watcher_config.clone().into(),
-                    sl_chain_id_subpool.clone(),
-                )
-                .await
-                .expect("failed to start L1 chain id update watcher")
-                .run()
-                .map(report_exit("L1 chain id update watcher")),
-            );
-        } else {
+        if config.general_config.gateway_rpc_url.is_some() {
```
itegulov: Related to the point above: I think we need a better marker of whether we are settling on gateway or not

itegulov (approved):

local-chains/v31.0/versions.yaml
Artemka374: Right, I didn't regenerate state

PR #1034, opened 2026-03-18
feat(readctor `ReplayRecord`): extract `BlockStartCursors` struct from flat cursor fields (eg `l1_priority_id`)

romanbrodetski-ai:
## Summary
- Introduces a new `BlockStartCursors` struct in `lib/types` that groups the four L1 watcher cursor fields (`l1_priority_id`, `interop_event_index`, `migration_number`, `interop_fee_number`) that were previously duplicated as flat fields across multiple types.
- Replaces flat cursor fields in `ReplayRecord`, `PreparedBlockCommand`, `BlockContextProvider`, and the node entrypoint with a single `BlockStartCursors` field, reducing parameter counts and duplication.
- Preserves serde backwards-compatibility via `#[serde(flatten)]` + `#[serde(rename)]`. Wire format files (v1.rs, v2.rs) are unchanged. RocksDB column families still store cursors individually with TODO comments marking the migration path.

## Test plan
- [ ] CI passes (cargo check, fmt, clippy, unit tests, integration tests)
- [ ] Verify serde round-trip compatibility (the `#[serde(flatten)]` + `#[serde(rename)]` produces identical JSON to the old flat layout)

No new tests added — this is a pure refactoring that preserves existing behavior. Existing integration tests cover the affected code paths.

### DIFF
--- a/lib/network/src/wire/replays/v2.rs
+++ b/lib/network/src/wire/replays/v2.rs
@@ -2,9 +2,6 @@
 //! stable unique ID for a type, so instead we define it in this separate file.
 //!
 //! Do not change this file under any circumstances. Copy it instead. May be deleted when obsolete.
-//! `BlockStartCursors` was introduced in the internal model/storage layer to group the replay
-//! cursor fields, but the wire representation did not change. Because this version still encodes
-//! the same payload shape, that refactor did not require introducing `v3`.
 
 // Difference from v1:
 // - Added `starting_migration_number` field to `ReplayRecord`.


lib/network/src/wire/replays/v2.rs
```diff
@@ -2,6 +2,9 @@
 //! stable unique ID for a type, so instead we define it in this separate file.
 //!
 //! Do not change this file under any circumstances. Copy it instead. May be deleted when obsolete.
+//! `BlockStartCursors` was introduced in the internal model/storage layer to group the replay
+//! cursor fields, but the wire representation did not change. Because this version still encodes
+//! the same payload shape, that refactor did not require introducing `v3`.
```
itegulov: I think the comment is excessive, these files (v1.rs, v2.rs etc) are here exactly because we want separation between storage and network models

PR #1067, opened 2026-03-24
fix(priority-tree): run initialization in background to avoid shutdown bug

kostia244:
## Summary

I make the heavy part of init run together with the main priority tree task, insite the reth-task.

run() function is added to PriorityTreeManager. This reduces code duplication.

I'm not sure about PriorityTreePipelineStep, is it in the right palce?

Context:
PriorityTreeManager initialization is currently done inside run. This component waits for background processes to do their job without accounting for their crashes(wait_for_replays never returns). Since graceful shutdown logic relies on the run() function finishing, this fix is necessary.


lib/priority_tree/src/lib.rs
```diff
@@ -31,34 +31,52 @@ pub struct PriorityTreeManager<ReplayStorage, Finality> {
...
+            committed_batch_provider,
+            last_executed_batch_on_init: 0,
+            initial_block_number,
+        })
+    }
+
+    /// Performs the async initialization: replays any blocks that are already executed on L1
+    /// but not yet reflected in the persisted priority tree. Must be called before any other
+    /// method that depends on `last_executed_batch_on_init`.
+    pub async fn init(&mut self) -> anyhow::Result<()> {
```
itegulov: Can't we fully move `init` into `prepare_execute_commands`? AFAIU the other task does not require initialization but please double check.

PR #1069, opened 2026-03-24
feat: add remote fork storage for rpc simulation

itegulov-claude:
## Summary
- add unstable RPC methods for reading raw storage values and preimages from a node
- wrap RPC state access with a remote fork fallback backed by 
- enable the fallback for node RPC simulation flows and document the behavior

## Verification
- cargo check -p zksync_os_rpc -p zksync_os_server

### DIFF
--- a/lib/rpc/src/rpc_storage.rs
+++ b/lib/rpc/src/rpc_storage.rs
@@ -4,8 +4,9 @@ use std::{ops::RangeInclusive, sync::Arc};
 use zksync_os_interface::traits::{PreimageSource, ReadStorage};
 use zksync_os_merkle_tree_api::MerkleTreeProver;
 use zksync_os_storage_api::{
-    ReadBatch, ReadFinality, ReadReplay, ReadRepository, ReadStateHistory, RepositoryError,
-    RepositoryResult, StateError, StateResult, ViewState, notifications::SubscribeToBlocks,
+    ReadBatch, ReadFinality, ReadReplay, ReadRepository, ReadStateHistory, RepositoryBlock,
+    RepositoryError, RepositoryResult, StateError, StateResult, ViewState,
+    notifications::SubscribeToBlocks,
 };
 
 pub trait ReadRpcStorage: ReadStateHistory + Clone {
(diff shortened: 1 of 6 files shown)


node/bin/src/lib.rs
```diff
@@ -712,13 +712,23 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+        .map(Arc::new)
+        .map(|client| client as Arc<_>);
+    let rpc_storage = RpcStorage::with_remote_fork(
         repositories.clone(),
         block_replay_storage.clone(),
         finality_storage.clone(),
         persistent_batch_storage.clone(),
         state.clone(),
         tree_for_rpc,
+        remote_fork_storage,
```
itegulov: This means only RPC is using forked storage. All other components (such as sequencer) must also use forked storage.

lib/rpc/src/rpc_storage.rs
```diff
@@ -215,3 +219,615 @@ pub enum RpcStorageError {
...
+    fn get_stored_transaction(&self, hash: TxHash) -> Option<StoredTxData>;
+    fn get_latest_block(&self) -> Option<BlockNumber>;
+    fn get_replay_record(&self, block_number: BlockNumber) -> Option<ReplayRecord>;
+    fn get_latest_replay_record(&self) -> Option<BlockNumber>;
+}
+
+#[derive(Clone, Debug)]
+pub struct ForkedRepository<Local> {
+    local: Local,
+    remote_fork: Option<Arc<dyn RemoteForkSource>>,
```
itegulov: Make non-optional here and in other places

lib/storage_fork/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+    fn get_latest_replay_record(&self) -> Option<BlockNumber> {
+        self.with_runtime(self.client.get_latest_replay_record_number(), |err| {
+            tracing::warn!(%err, "failed to fetch remote fork latest replay record");
+        })
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ForkedRepository<Local> {
+    local: Local,
```
itegulov: `Local` here must always be in-memory implementation, add it if it's missing

lib/storage_fork/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+        match self {
+            Self::Local(local) => local.subscribe_to_blocks(),
+            Self::Forked(forked) => forked.subscribe_to_blocks(),
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct ForkedReplayStorage<Local> {
+    local: Local,
```
itegulov: Local here must always be in-memory implementation, add it if it's missing

lib/storage_fork/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+        self.local
+            .add_block_result(block_number, storage_diffs, new_preimages, override_allowed)
+    }
+}
+
+#[derive(Clone, Debug)]
+pub enum StateHistoryStorage<Local> {
+    Local(Local),
+    Forked(ForkedStateHistory<Local>),
+}
```
itegulov: You have two ReadStateHistory implementations

PR #1073, opened 2026-03-24
feat(config): generate declarative config validation

perekopskiy-ai:
## Summary
- add a proc-macro-based config validation system that generates `Config::validate()` from field annotations
- move main/external node config requirements next to the relevant fields and support async field validation
- recurse into subconfigs by default for `*_config` fields and `#[config(nest)]`, with an explicit opt-out

## Testing
- cargo test -p zksync_os_server config::tests::
- cargo clippy --all-targets --all-features --workspace -- -D warnings

### DIFF
--- a/node/bin/src/main.rs
+++ b/node/bin/src/main.rs
@@ -43,6 +43,11 @@ struct Cli {
 
     #[command(subcommand)]
     cmd: Option<CliCommand>,
+
+    /// Makes the node to stop before calling `run` method.
+    /// It is used to catch issues with configuration.
+    #[arg(long)]
+    no_run: bool,
 }
 
 fn load_config_defaults(config_sources: &mut ConfigSources, config_paths: Option<Vec<String>>) {
@@ -151,6 +156,12 @@ pub async fn main() {
     tracing::info!(?config, "Loaded config");
     load_internal_config(&mut config);
     config.validate().await.expect("invalid config");
+
+    if opt.no_run {
+        tracing::info!("Node config was loaded successfully, exiting due to --no-run flag");
+        return;
+    }
+
     // ======= Run tasks ===========
     let ephemeral_enabled = config.general_config.ephemeral;
     let _ephemeral_guard = ephemeral_enabled.then(|| enable_ephemeral_mode(&mut config));


node/bin/src/config/mod.rs
```diff
@@ -149,6 +175,48 @@ impl Config {
...
+        let prove = match &l1_sender_config.operator_prove_sk {
+            Some(signer) => signer,
+            None => return Ok(()),
+        };
+        let execute = match &l1_sender_config.operator_execute_sk {
+            Some(signer) => signer,
+            None => return Ok(()),
+        };
+
+        let commit_addr = commit.address().await.map_err(|err| {
```
itegulov: I believe this will need access to GCP when secret is in GCP KMS HSM. Will that work with gitops CI?

node/bin/src/main.rs
```diff
@@ -150,11 +150,9 @@ pub async fn main() {
     let mut config = build_external_config(config_repo).await;
     tracing::info!(?config, "Loaded config");
     load_internal_config(&mut config);
+    config.validate().await.expect("invalid config");
```
itegulov: Do you plan to use this in gitops CI? If so, then it probably makes to also add `Validate` variant to `ConfigCommand` with matching implementation. Also, I'd make sure it prints all errors at once as opposed to one at a time (i.e. not fail fast).

node/bin/src/config/mod.rs
```diff
@@ -257,19 +330,31 @@ pub struct GeneralConfig {
...
+    ))]
     pub enabled: bool,
     /// The node's secret key (256-bit ECDSA), from which the node's identity is derived. Used during
     /// initial RLPx handshake.
     #[config(secret)]
     #[config(default, with = SecretKeyDeserializer)]
+    #[config_validate(validate(
+        |root: &Config, value: &Option<SecretKey>| !root.network_config.enabled || value.is_some(),
+        "is required when `network.enabled=true`"
+    ))]
```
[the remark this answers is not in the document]
perekopskiy: The problem is that we need validation between different structs. I'm not sure how to do that in a good way with smart-config. My goal was to patch our specific use-case

node/bin/src/main.rs
perekopskiy: nvm, added to this PR

PR #1075, opened 2026-03-25
feat(network): support `network_interface` and DNS boot nodes

itegulov-claude:
## Summary

- add `network.interface` / `network_interface` as a separate field that resolves a local network interface at startup and overrides `network.address`
- keep `network.boot_nodes` DNS-aware by parsing them as `TrustedPeer`, so entries like `enode://<pk>@<dns_name>:<port>` are accepted as-is
- document the updated external-node networking configuration

## Testing

- `cargo test -p zksync_os_server --lib config::`

### DIFF
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -219,7 +219,7 @@ impl Tester {
     ) -> anyhow::Result<Self> {
         let overrides_fun = |config: &mut Config| {
             config.general_config.node_role = NodeRole::ExternalNode;
-            config.network_config.boot_nodes = vec![self.node_record];
+            config.network_config.boot_nodes = vec![self.node_record.into()];
             config.general_config.main_node_rpc_url = Some(self.l2_rpc_address.clone());
             config.l1_sender_config.pubdata_mode = None;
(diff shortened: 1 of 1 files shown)


node/bin/src/config/mod.rs
```diff
@@ -271,18 +271,20 @@ pub struct NetworkConfig {
     #[config(secret)]
     #[config(default, with = SecretKeyDeserializer)]
     pub secret_key: Option<SecretKey>,
-    /// IPv4 address to use for Node Discovery Protocol v5 (discv5) and RLPx Transport Protocol (rlpx).
-    #[config(default_t = Ipv4Addr::UNSPECIFIED, with = Serde![str])]
+    /// IPv4 address to use for Node Discovery Protocol v5 (discv5) and RLPx Transport Protocol
+    /// (rlpx). Can also be provided via `network.interface` / `network_interface`.
+    #[config(default_t = Ipv4Addr::UNSPECIFIED, with = Serde![str], alias = "interface")]
```
itegulov: `interface` should not be an alias to `address`, but rather a different field that regulates which networking interface (aka net_if) to use to resolve ip address INSTEAD of what is specified in `address`

PR #1082, opened 2026-03-26
fix(network): retry boot node DNS resolution before startup

itegulov-claude:
## Summary
- resolve boot node DNS before building the reth network config so reth does not do the one-shot resolution itself
- retry unresolved boot nodes with exponential backoff during startup and proceed once at least one boot node resolves
- add unit coverage for immediate resolution, retry behavior, and backoff capping

## Testing
- cargo test -p zksync_os_network --lib

### DIFF
--- a/lib/network/src/service.rs
+++ b/lib/network/src/service.rs
@@ -4,7 +4,7 @@ use crate::version::{ZksProtocolV1, ZksProtocolV2};
 use crate::wire::replays::RecordOverride;
 use alloy::eips::eip2124::Head;
 use alloy::primitives::BlockNumber;
-use backon::{ConstantBuilder, Retryable, Sleeper};
+use backon::{ConstantBuilder, Retryable};
 use futures::future::join_all;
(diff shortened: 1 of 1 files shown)


lib/network/src/service.rs
```diff
@@ -25,6 +29,95 @@ use zksync_os_types::NodeRole;
...
+const MAX_BOOT_NODE_RESOLUTION_BACKOFF: Duration = Duration::from_secs(30);
+
+async fn resolve_boot_nodes_with_backoff(
+    boot_nodes: Vec<TrustedPeer>,
+) -> Result<Vec<TrustedPeer>, NetworkError> {
+    resolve_boot_nodes_with_backoff_using(
+        boot_nodes,
+        |boot_node| {
+            async move {
+                tokio::task::spawn_blocking(move || boot_node.resolve_blocking())
```
itegulov: Why `resolve_blocking` and not regular `resolve`?

PR #1098, opened 2026-03-26
fix: add integration test for rebuild after emptying historical block

romanbrodetski-ai:
## Summary
- add a restart helper that allows config overrides in integration tests
- add a rebuild integration test that empties an older block and verifies a later block is still rebuilt
- use a second sender in the final block so the test distinguishes nonce-broken historical transactions from unaffected rebuilt transactions

## Notes
- this PR intentionally contains only the integration test coverage
- the new test is expected to fail on its own until the separate rebuild invalid-tx handling fix is merged

### DIFF
--- a/integration-tests/tests/node/rebuild.rs
+++ b/integration-tests/tests/node/rebuild.rs
@@ -1,0 +1,188 @@
+use alloy::network::{EthereumWallet, TransactionBuilder};
+use alloy::primitives::{Address, U256};
+use alloy::providers::{Provider, ProviderBuilder};
+use alloy::rpc::types::TransactionRequest;
+use alloy::signers::local::LocalSigner;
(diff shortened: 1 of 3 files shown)


integration-tests/tests/node/rebuild.rs
```diff
@@ -0,0 +1,188 @@
...
+                    .with_value(U256::from(1u64)),
+            )
+            .await?
+            .expect_successful_receipt()
+            .await?;
+        primary_last_block = receipt
+            .block_number
+            .expect("transfer receipt should have a block number");
+    }
+    let second_sender_receipt = second_signer
```
RomanBrodetski: let's add a comment that we are trying to get this second wallet's tx to the very last block

PR #1108, opened 2026-03-27
fix: make block executor wait for block applier

romanbrodetski-ai:
## Summary
Wait for the previous block to be fully applied before executing the next one, using the startup block boundary as intial value

## Why
This removes a race between BlockExecutor and BlockApplier during replay and rebuild.

This fix is necessary for #1107 to succeed.

## Validation
- the targeted rebuild integration test in #1107 passes with this fix in place


lib/sequencer/src/execution/block_executor.rs
```diff
@@ -182,6 +191,40 @@ where
     }
 }
 
+async fn wait_for_block_applier(
```
perekopskiy: Now that there is no boolean flag I think that a separate method is useless, you can just call `wait_for`

PR #1123, opened 2026-03-30
feat: pipeline backpressure handle

antonbaliasnikov:


lib/observability/src/component_health_reporter.rs
```diff
@@ -0,0 +1,230 @@
...
+    pub state_entered_at: Instant,
+    /// Block number of the last item successfully processed. `None` until first call to
+    /// `record_processed`.
+    pub last_processed_block_number: Option<u64>,
+    /// Block timestamp of the last processed block. `None` if not yet processed or unavailable
+    /// (e.g. batch-level components that call `record_processed` with `None`).
+    pub last_processed_block_timestamp: Option<u64>,
+    /// When record_processed was last called (None until first call).
+    /// Independent from state_entered_at — tracks processing rate, not state duration.
+    /// This is [`tokio::time::Instant`]; pair with `tokio::time::Instant::now()` for durations.
```
RomanBrodetski: nit:
>     /// This is [`tokio::time::Instant`]; pair with `tokio::time::Instant::now()` for durations.

Does this really deserve a doc comment? Won't type system prevent an error there?

lib/observability/src/generic_component_state.rs
```diff
@@ -1,29 +1,35 @@
...
-    // we cannot effectively distinguish between Processing and Waiting for input,
-    // as both happen simultaneously
-    ProcessingOrWaitingRecv,
+    /// No work available — waiting for upstream to produce more.
+    Idle,
+    /// Actively processing an item.
+    Active,
+    /// Has work queued but blocked on an external resource
+    /// (e.g. L1 confirmation, prover job slots, service reconnect).
+    Throttled,
```
RomanBrodetski: I like this separation!

RomanBrodetski (commented):
Need to jump before finishing the review, but publishing what I already have.


Oveall looks really good - can't way to merge this! I think we are close

lib/sequencer/src/execution/block_canonizer.rs
```diff
@@ -121,7 +122,19 @@ where
...
+                            .send(BlockPayload {
+                                output: block_output,
+                                record: produced_replay,
+                                command_type: cmd_type,
+                            })
+                            .is_err()
+                        {
+                            anyhow::bail!("Outbound channel closed");
+                        }
+                        self.health_reporter.record_processed(block_number, Some(block_ts));
```
RomanBrodetski: https://github.com/matter-labs/zksync-os-server/pull/1123#discussion_r3046966834 - if we did it like I propose here, we'd be able to remove ~10 lines from here - and this logic (sending + recording) anyway belongs elsewhere - now it's indise a component that does some specific business-logic.

lib/status/src/pipeline.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// storage writes or downstream population complete. Do not treat this as a durability
+    /// guarantee — a block may appear here while it is still being persisted.
+    pub last_processed_block: u64,
+    /// Blocks behind the pipeline head (BlockExecutor). Always present.
+    pub head_block_lag: u64,
+    /// Blocks behind this component's direct upstream neighbour. This is the value compared
+    /// against `max_block_lag` for backpressure. Absent for components with no registered
+    /// upstream adjacency (e.g. BlockExecutor itself), which are measured head-relative.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub adjacent_block_lag: Option<u64>,
```
RomanBrodetski: Honestly it took me a while to understand exactly what this represents - the comment is only partially helpful.

This is the number of blocks between the last block fully processed by this component and the last block fully processed by the prev. component.

Part of the confusin is because in fact each component may need to have two pointers - last processed block (like now) and last PICKED block (not sure if tracked now). Some components (most notably proves) work an a whole range of blocks. Are such cases properly wiorking now?

We may do that as a follow up, but we should still be explicit that we only track the last processed one in this PR.

antonbaliasnikov:
Addressed review and superceded by https://github.com/matter-labs/zksync-os-server/pull/1190.

PR #1130, opened 2026-04-01
feat(batcher): SLI-based sealing

dimazhornyk:
## Summary

The settlement SLI wasn't maintained properly in the previous implementation. If the server restarted, the batching deadline would get reset. 
This implementation:
- reads L1 for the latest `executeBatches` on server startup and stores its timestamp
- watches the executed batches when the server is running
- uses values from the two sources above to determine when to seal a batch, leaving some buffer for proving

### DIFF
--- a/node/bin/src/batcher/mod.rs
+++ b/node/bin/src/batcher/mod.rs
@@ -315,10 +315,7 @@ impl<ReadState: ReadStateHistory + Clone + Send + 'static> Batcher<ReadState> {
                                     self.batcher_config.batch_timeout,
                                 );
                                 tracing::info!(
-                                    batch_number,
-                                    first_block_timestamp,
-                                    unix_deadline,
-                                    "Armed batch deadline from first block timestamp."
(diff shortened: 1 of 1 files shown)


node/bin/src/batcher/mod.rs
```diff
@@ -292,6 +294,28 @@ impl<ReadState: ReadStateHistory + Clone + Send + 'static> Batcher<ReadState> {
                                 leaf_count,
                             };
 
+                            // Arm the deadline on the first eligible block using its L2 timestamp.
+                            // Using an absolute wall-clock instant derived from the block timestamp
+                            // makes the deadline restart-resilient: the block timestamp is part of
+                            // chain state and does not reset when the server restarts.
+                            if deadline.is_none() {
+                                if block_number >= self.startup_config.last_persisted_block {
```
perekopskiy: I think we should remove this line, and `last_persisted_block` from `startup_config`, otherise your change will have no effect

PR #1145, opened 2026-04-06
fix(batch_verification): remove blocks only when processing the next request

dimazhornyk:
## Summary

After a restart, the server's last_committed_batch_number is captured from L1 state at startup and never updates. Meanwhile, L1CommitWatcher starts concurrently and can observe a batch commitment whose L1 tx was still pending at startup. This advances last_committed_block (via the finality watch channel) past a batch the server still considers uncommitted.
That leads to the removal of blocks needed for the validation of the current batch from the cache.


RomanBrodetski:
As discussed in slack ,this is not the behavikour we desire

PR #1149, opened 2026-04-07
feat(network)!: move batch verification to devp2p

EmilLuta:
## Summary

Move batch verification transport from the legacy batch-verification wire format
onto the `zks/3` devp2p subprotocol.

This PR adds verifier auth plus `VerifyBatch` / `VerifyBatchResult` over RLPx,
threads MN/EN verification flow through `lib/network`, and updates batch
verification orchestration to dispatch only to authorized and sufficiently
caught-up `zks/3` peers.

Reviewer guidance:
- Start with `docs/src/design/devp2p.md`
- Then look at `lib/network/src/service.rs`
- Then `lib/network/src/protocol/mn.rs` and `lib/network/src/protocol/en.rs`
- Then `lib/batch_verification/src/main_node/component.rs` and
  `lib/batch_verification/src/verifier/mod.rs`
- Finish with `lib/network/tests/e2e.rs` and the new `service.rs` tests

## Breaking Changes

- Affected users: main nodes and external nodes that participate in batch verification.
- Batch verification no longer uses the old `lib/batch_verification` wire-format transport.
  Verification traffic now runs over the `zks/3` devp2p subprotocol.
- Nodes that need batch verification must run code that supports the new `zks/3`
  verifier auth and `VerifyBatch` / `VerifyBatchResult` flow.
- The old `wire_format` transport and its fixture tests are removed as part of the move to
  `zks/3` devp2p transport. The new `zks/3` messages are covered by unit and e2e tests in
  `lib/network`.
- Related docs: `docs/src/design/devp2p.md`.

## Rollout Instructions

- Roll out main nodes and verifier-capable ENs together; mixed old/new batch-verification
  transport is not supported.
- Verify that ENs connect over `zks/3`, complete verifier auth, and that MN logs show
  `VerifyBatch` dispatch only to authorized, caught-up peers.
- Monitor for:
  - verifier authorization failures
  - missing `VerifyBatchResult` responses
  - `MaxActiveConnectionsExceeded`
- Roll back by reverting MN and EN nodes together to the previous release; partial rollback
  leaves the batch-verification transport mismatched.


lib/network/src/service.rs
```diff
@@ -316,29 +393,206 @@ impl NetworkService {
...
+        let Some(connection) = connection else {
+            tracing::warn!(
+                peer_id = %peer_id,
+                request_id = request.request_id,
+                batch_number = request.batch_number,
+                "skipping verify request: missing active connection"
+            );
+            continue;
+        };
+        if connection.version != crate::version::ZksVersion::Zks3 {
```
perekopskiy: is it possible to check < Zks3 here?

perekopskiy (dismissed):

cytadela8 (commented):
Partial review

perekopskiy (approved):

lib/network/src/service.rs
```diff
@@ -314,31 +391,208 @@ impl NetworkService {
...
+        if connection.version < crate::version::ZksVersion::Zks3 {
+            tracing::warn!(
+                peer_id = %peer_id,
+                request_id = request.request_id,
+                batch_number = request.batch_number,
+                version = ?connection.version,
+                "skipping verify request: peer is not on zks/3"
+            );
+            continue;
+        }
```
cytadela8, line 536: This check seems redundant. You are already getting only peers that are `authorized_verifier_peers`

cytadela8: No reason really to skip it also.

lib/network/src/service.rs
```diff
@@ -314,31 +391,208 @@ impl NetworkService {
...
+        if connection
+            .outbound_tx
+            .send(ZksMessage::<ZksProtocolV3>::VerifyBatch(request.clone()).encoded())
+            .await
```
cytadela8, line 540: I think outbound_tx is a channel. This will block if it feels up.

cytadela8 (commented):
The networking code is sprinkled with channels that fill up, resulting in availability issues. Although you are not first to add such code, I think in case of batch_verification it's easier to avoid it by just using try_send. If we skip sending a batch_verification request, nothing bad will happen.

PR #1152, opened 2026-04-07
feat(revm-consistency-checker): Add metrics with last divergence timestamp

kostia244:
## Summary

Add a 2 metrics with timestamps of last divergent block: 1 for case where the block is rebuilt and one for case where it's not.

I use LabeledFamily, becuase it has lazy initialization. Otherwise the metric would be reset to zero which is misleading. Becuase it's reported rarely we'll still need to use `last_over_time`

Alert condition would be something like
```

      watchdog_transfer_failing_or_missing:
        promql: '(max(last_over_time(revm_consistency_checker_last_divergence_timestamp_seconds{namespace="{{ .Release.Namespace }}"}[7d])) or vector(0))'
        compare: ">"
        thresholds:
          P3: 0
        duration: 1m
        annotations:
          description: "REVM divergence detected"
          summary: "REVM divergence detected"
          runbook_url: "..."
        labels:
          owner: zksync-os-team
```

### DIFF
--- a/lib/revm_consistency_checker/src/metrics.rs
+++ b/lib/revm_consistency_checker/src/metrics.rs
@@ -1,0 +1,27 @@
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use vise::{Gauge, Metrics};
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "revm_consistency_checker")]
+pub(crate) struct RevmConsistencyCheckerMetrics {
+    /// Unix timestamp of the most recent detected inconsistency.
+    pub last_inconsistency_timestamp: Gauge<u64>,
+    /// Block number of the most recent detected inconsistency.
+    pub last_inconsistent_block_number: Gauge<u64>,
+}
+
+#[vise::register]
+pub(crate) static METRICS: vise::Global<RevmConsistencyCheckerMetrics> = vise::Global::new();
+
+impl RevmConsistencyCheckerMetrics {
+    pub fn record_inconsistency(&self, block_number: u64) {
+        self.last_inconsistent_block_number.set(block_number);
+        self.last_inconsistency_timestamp.set(
+            SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .unwrap_or_default()
+                .as_secs(),
+        );
+    }
+}
--- a/lib/revm_consistency_checker/src/metrics.rs
+++ b/lib/revm_consistency_checker/src/metrics.rs
@@ -4,7 +4,7 @@ use vise::{Gauge, Metrics, Unit};
 
 #[derive(Debug, Metrics)]
 #[metrics(prefix = "revm_consistency_checker")]
-pub(crate) struct RevmConsistencyCheckerMetrics {
+pub struct RevmConsistencyCheckerMetrics {
     /// Unix timestamp of the most recent detection -- used for alerts
     #[metrics(unit = Unit::Seconds)]
     pub last_inconsistency_timestamp: Gauge<u64>,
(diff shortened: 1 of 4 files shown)


node/bin/src/main.rs
```diff
@@ -288,6 +484,8 @@ fn load_internal_config(config: &mut Config) {
         .l2_signer_blacklist
         .extend(internal_config.l2_signer_blacklist);
     if let Some(failing_block) = internal_config.failing_block {
+        zksync_os_revm_consistency_checker::metrics::METRICS
+            .record_divergence(RevmDivergenceOutcome::Reverted);
```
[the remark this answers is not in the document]
slowli: FWIW, I recall @tomg10 [adding code to the push variant of the metrics exporter](https://github.com/matter-labs/vise/commit/1c9cc500e92cf9ea052b230e114a6f9cce4fb2c1) specifically to address pushing metrics on shutdown. It could be possible to add similar logic to the pull version and have a future that resolves after the next metrics push / pull. IMO, this isn't particularly good (although better than the current approach) because it makes the divergence component aware of the metrics *exporter*, and one should be careful to have a timeout for the aforementioned push / pull future because it may take unpredictable time to complete (I guess this could be built into the API that would produce such futures).

lib/revm_consistency_checker/src/node.rs
```diff
@@ -49,7 +50,18 @@ where
...
+        if report.is_empty() {
+            return Ok(());
+        }
+
+        let message = format!(
+            "REVM consistency check failed for block number {}, block hash {}",
+            replay_record.block_context.block_number,
+            block_output.header.hash(),
+        );
+        tracing::warn!(message);
```
slowli, line 62: Nit: Using a separate `format!` with tracing macros looks weird; it's possible to just inline its args into the tracing macro (or better yet, define interpolated args as values).

lib/revm_consistency_checker/src/metrics.rs
```diff
@@ -0,0 +1,32 @@
...
+    Accepted,
+    Reverted,
+}
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "revm_consistency_checker")]
+pub struct RevmConsistencyCheckerMetrics {
+    /// Unix timestamp of the most recent divergence by outcome -- used for alerts.
+    #[metrics(unit = Unit::Seconds, labels = ["outcome"])]
+    pub last_divergence_timestamp: LabeledFamily<RevmDivergenceOutcome, Gauge<u64>>,
```
slowli, line 17: I can't say that I'm very familiar with alerting, but it *looks* like it would make sense to use a counter here, where increase events correspond to divergences, and the alerting rule checks whether there are any increases in the considered time interval. Since Prometheus is fundamentally a time series DB, the timestamp of these events should be available inherently (I guess it could be possible that it has insufficient precision, but the alerting rule mentioned in the PR doesn't seem to require significant accuracy).

kostia244: counters are problematic because they do not persist between runs. I feel like it would be misleading, the value won't really represent anything meaningful

lib/revm_consistency_checker/src/node.rs
kostia244: The same message is then being passed to write_config_and_panic, I'm not sure how to avoid duplication if this part is changed.

node/bin/src/main.rs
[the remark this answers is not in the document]
kostia244: I can't find a perfectly clean way to handle it. And I don't want to rely on prometheus specifics too much. The best idea I had is:

Add `unexpected_event_timestamp: LabeledFamily<&'static str, Gauge<i64>>` to general metrics for such cases, then add a similar variable to `internal_config` to report the cases that include a restart.

For now I suggest to simply not report this case for the first version to verify that divergencies are not happening.

lib/revm_consistency_checker/src/metrics.rs
slowli: AFAIU, Prometheus APIs (can) account for counter resets between runs (e.g., via `increase` from PromQL), since otherwise counters would be unreasonably difficult to use in general. Won't insist on the counter use; looks not that important.

PR #1155, opened 2026-04-07
feat: bytecodes supplier support

tomg10:
## Summary

Adding support for fetching bytecodes from L1 instead of using hardcoded list


integration-tests/src/upgrade/builder.rs
```diff
@@ -108,16 +119,17 @@ impl ProtocolUpgradeBuilder {
...
-            // TODO: the current implementation is faulty, since it uses `full_bytecode_len`; the reason for that
-            // is the fact that we deploy preimages using creation bytecode (which includes bytecode artifacts),
-            // but for "real" force deployments we need to use the actual deployed bytecode.
-            // Once BytecodesSupplier is ready for zksync-os, we need to change this logic to use observable bytecode len.
+            // NOTE: `setBytecodeDetailsEVM` in era-contracts uses `bytecodeSize` for both
+            // `bytecode_length` and `observable_bytecode_length`. In production this should be the
+            // observable (raw EVM) bytecode length. Here we use `full_bytecode_len()` because the
+            // test registers preimages via L2 deploy, which records the full padded size. Using
+            // `observable_bytecode_len` would cause a preimage cache panic at runtime.
             let deployed_bytecode_info = super::interfaces::ForceDeploymentBytecodeInfo {
```
perekopskiy: Can it be changed so it matches production?

integration-tests/src/upgrade/builder.rs
```diff
@@ -28,8 +28,11 @@ pub struct ProtocolUpgradeBuilder {
     /// Timestamp after which upgrade can be executed
     /// If not provided, default value will be used (e.g. upgrade whenever)
     timestamp: U256,
-    // I don't need more parameters here for now, but in the future if you want
-    // to extend this builder, feel free to do it.
+    /// Whether to include bytecode hashes in `factory_deps` of the upgrade tx.
+    /// When true, the server's `fetch_force_preimages` will look up bytecodes
+    /// from the L1 `BytecodesSupplier`. When false, the server relies on
+    /// preimages already known from L2 deploys.
+    include_factory_deps: bool,
```
perekopskiy, line 39: IIUC we can get rid of it after we drop support for v30, always use BytecodesSupplier, right? Perhaps worth adding a comment

lib/l1_watcher/src/upgrade_tx_watcher.rs
```diff
@@ -151,6 +165,9 @@ impl L1UpgradeTxWatcher {
                 .event_signature(NewUpgradeCutData::SIGNATURE_HASH)
                 .topic1(*raw_protocol_version);
             upgrade_cut_data_logs = self.provider_sl.get_logs(&filter).await?;
+            if from_block == start_block {
+                break;
+            }
```
perekopskiy: why is this needed? won't loop exit because of while condition?

lib/l1_watcher/src/upgrade_tx_watcher.rs
```diff
@@ -221,53 +238,121 @@ impl L1UpgradeTxWatcher {
...
+                }
+            }
+
+            // Stop early once all requested hashes are found.
+            if by_hash.len() == requested.len() {
+                break;
+            }
+            if from_block == start_block {
+                break;
+            }
```
perekopskiy: same here

lib/l1_watcher/src/upgrade_tx_watcher.rs
tomg10: without this the code doesn't handle the cases there start_block = 0 (goes into infinite loop)

lib/l1_watcher/src/upgrade_tx_watcher.rs
tomg10: without this the code doesn't handle the cases there start_block = 0

integration-tests/src/upgrade/builder.rs
tomg10: added

lib/l1_watcher/src/upgrade_tx_watcher.rs
perekopskiy:
```
let start_block = current_block
            .saturating_sub(UPGRADE_DATA_LOOKBEHIND_BLOCKS) // Upgrade could've been set a long time ago.
            .max(1u64);
```
start_block >=1

tomg10: fair enough, removing

integration-tests/src/upgrade/builder.rs
tomg10: corrected

perekopskiy (approved):

PR #1158, opened 2026-04-08
feat(storage): build RocksDB log index on block write/rollback

LogvinovLeon:
Second in the `eth_getLogs` index series ([#1156](https://github.com/matter-labs/zksync-os-server/pull/1156) added the trait). This PR builds and maintains the actual index — the next PR wires it into `scan_logs`.

## The problem

`eth_getLogs` currently does an O(block_range) bloom scan: every block is fetched from RocksDB and its bloom filter checked. A 10k-block range with a busy contract takes ~1.5s. With an address constraint, most of those blocks are irrelevant but we have no way to skip them.

## What this adds

A roaring-bitmap inverted index in two new RocksDB column families:
- `LogBlocksByAddress` — key: `address[20] ++ chunk_start[8]`, value: bitmap of block numbers
- `LogBlocksByTopic` — key: `topic[32] ++ chunk_start[8]`, value: bitmap of block numbers

**Chunk size is 2^16 = 65536 blocks.** This aligns with roaring32's internal container boundary: all block numbers within a chunk share the same high 16 bits, so each chunk serializes to exactly one roaring container (≤8 KB). A query spanning N chunks reads and ORs N keys.

**Write path** (`write_block`): for each transaction receipt, every log's address and topics are added to the relevant chunk bitmap in the same RocksDB write batch as the block data — so the index is always consistent with block storage.

**Rollback path** (`rollback`): iterates blocks being rolled back, removes each log from its chunk bitmap (deleting the key if the bitmap becomes empty), and rolls back the `log_index_last_block` coverage marker.

**Coverage metadata** (`log_index_first_block` / `log_index_last_block` in the `Meta` CF): tracks which block range is indexed. `first_block` is written once on the first indexed block and never changes; `last_block` is updated on every write and rollback. Callers intersect their query range with coverage and fall back to bloom scan for any uncovered prefix.

## Reviewing guide

All the interesting logic is in `lib/storage/src/db/repository/log_index.rs` (~510 lines). Start there. `repository.rs` changes are mechanical plumbing (new CFs, passing `log_index_started` flag, calling into the submodule).

The 8 unit tests at the bottom of `log_index.rs` each cover one specific property — read them alongside the functions they exercise:

| Test | What it pins down |
|---|---|
| `address_index_round_trip` | basic write → query |
| `topic_index_round_trip` | same for topic CF |
| `query_respects_requested_range` | result is clipped to requested range, not full coverage |
| `deindex_removes_block` | rollback correctly removes from bitmap and updates coverage |
| `multi_block_rollback_removes_all_blocks` | rolling back multiple blocks in one batch removes all of them (regression) |
| `no_index_returns_empty` | empty DB returns empty bitmap + empty range |
| `different_addresses_do_not_interfere` | address prefix correctly isolates entries |
| `chunk_boundary_spanning_query` | multi-chunk queries OR bitmaps correctly |

See the [testing notes comment](https://github.com/matter-labs/zksync-os-server/pull/1158#issuecomment-4214119226) for E2E test results, the rollback bug found, and the `WriteBatchWithIndex` alternative that was considered and rejected.

## Overhead

**Storage:** ~5 bytes per log per block (roaring bitmap amortized). At genlayer's current density (~3.3 KB/block): <0.2% overhead. At Ethereum mainnet density: ~1.5% overhead.

**Write path:** same number of `db.write(batch)` calls (1 per block). The batch gains 1 write per unique `(address, chunk)` and `(topic, chunk)` pair. Each unique key also requires one DB read (to load the existing bitmap into `BitmapCache`) — worth watching `persist_block` latency after deployment.

## Next PR

PR 3 adds `index_candidates()` to `scan_logs` — uses the bitmap to skip blocks outside the result set, with bloom fallback for the uncovered prefix. Also adds Roman's shadow-mode comparison to validate correctness against the existing bloom scan before cutting over.

Design doc: https://www.notion.so/335a48363f23810e9755cd8028b76973

### DIFF
--- a/lib/rocksdb/src/db.rs
+++ b/lib/rocksdb/src/db.rs
@@ -17,7 +17,7 @@ use std::{
 
 use rocksdb::{
     BlockBasedOptions, Cache, ColumnFamily, ColumnFamilyDescriptor, DB, DBPinnableSlice, Direction,
-    IteratorMode, Options, PrefixRange, ReadOptions, WriteOptions, perf, properties,
+    IteratorMode, Options, PrefixRange, ReadOptions, SliceTransform, WriteOptions, perf, properties,
 };
 use thread_local::ThreadLocal;
 use vise::MetricsFamily;
@@ -44,6 +44,15 @@ pub trait NamedColumnFamily: 'static + Copy {
     /// of compaction / memtables.
     fn requires_tuning(&self) -> bool {
         false
+    }
+
+    /// Returns the byte length of the fixed-length key prefix to use as a RocksDB prefix extractor
+    /// for this CF, or `None` if no prefix extractor should be configured.
+    ///
+    /// When set, RocksDB builds per-prefix bloom filters that allow `range_iterator_cf` to skip
+    /// SST files that contain no keys with the requested prefix.
+    fn prefix_extractor_len(&self) -> Option<usize> {
+        None
     }
 }
 
@@ -397,7 +406,7 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
 
         let cfs_and_options: HashMap<_, _> = CF::ALL
             .iter()
-            .map(|cf| (cf.name(), cf.requires_tuning()))
+            .map(|cf| (cf.name(), (cf.requires_tuning(), cf.prefix_extractor_len())))
             .collect();
         let obsolete_cfs: Vec<_> = existing_cfs
             .iter()
@@ -423,8 +432,8 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
         let cf_names = cfs_and_options.keys().copied().collect();
         let all_cfs_and_options = cfs_and_options
             .into_iter()
-            .chain(obsolete_cfs.into_iter().map(|name| (name, false)));
-        let cfs = all_cfs_and_options.map(|(cf_name, requires_tuning)| {
+            .chain(obsolete_cfs.into_iter().map(|name| (name, (false, None))));
+        let cfs = all_cfs_and_options.map(|(cf_name, (requires_tuning, prefix_extractor_len))| {
             let mut block_based_options = BlockBasedOptions::default();
             block_based_options.set_bloom_filter(10.0, false);
             if let Some(cache) = &caches.shared {
@@ -436,7 +445,10 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
             }
 
             let memtable_capacity = options.large_memtable_capacity.filter(|_| requires_tuning);
-            let cf_options = Self::rocksdb_options(memtable_capacity, Some(block_based_options));
+            let mut cf_options = Self::rocksdb_options(memtable_capacity, Some(block_based_options));
+            if let Some(len) = prefix_extractor_len {
+                cf_options.set_prefix_extractor(SliceTransform::create_fixed_prefix(len));
+            }
             ColumnFamilyDescriptor::new(cf_name, cf_options)
         });
 
@@ -649,6 +661,31 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
         self.inner
             .db
             .iterator_cf(cf, IteratorMode::From(keys.start, Direction::Forward))
+            .map(Result::unwrap)
+            .fuse()
+        // ^ unwrap() is safe for the same reasons as in `prefix_iterator_cf()`.
+    }
+
+    /// Iterates over key-value pairs in `cf` in lexical order from `from` (inclusive) up to
+    /// `to` (exclusive).  When the column family has a prefix extractor configured and `from` and
+    /// `to` share the same extracted prefix, RocksDB uses prefix bloom filters to skip SST files
+    /// that contain no keys with that prefix.
+    pub fn range_iterator_cf<'a>(
+        &'a self,
+        cf: CF,
+        from: &'a [u8],
+        to: Vec<u8>,
+    ) -> impl Iterator<Item = (Box<[u8]>, Box<[u8]>)> + 'a {
+        let cf_handle = self.column_family(cf);
+        let mut options = ReadOptions::default();
+        options.set_iterate_upper_bound(to);
+        // When the CF has a prefix extractor configured, this tells RocksDB to use
+        // prefix bloom filters for the initial seek — checking each SST file only if
+        // it contains keys with the same prefix as the seek key.
+        options.set_prefix_same_as_start(true);
+        self.inner
+            .db
+            .iterator_cf_opt(cf_handle, options, IteratorMode::From(from, Direction::Forward))
             .map(Result::unwrap)
             .fuse()
         // ^ unwrap() is safe for the same reasons as in `prefix_iterator_cf()`.
--- a/lib/rocksdb/src/db.rs
+++ b/lib/rocksdb/src/db.rs
@@ -668,9 +668,9 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
         // ^ unwrap() is safe for the same reasons as in `prefix_iterator_cf()`.
     }
 
-    /// Iterates over key-value pairs in `cf` in lexical order from `from` (inclusive) up to
-    /// `to` (exclusive).  When the column family has a prefix extractor configured and `from` and
-    /// `to` share the same extracted prefix, RocksDB uses prefix bloom filters to skip SST files
+    /// Iterates over key-value pairs in `cf` in lexical order over `range` (start inclusive,
+    /// end exclusive).  When the column family has a prefix extractor configured and both bounds
+    /// share the same extracted prefix, RocksDB uses prefix bloom filters to skip SST files
     /// that contain no keys with that prefix.
     pub fn range_iterator_cf<'a>(
         &'a self,
--- a/lib/rocksdb/src/db.rs
+++ b/lib/rocksdb/src/db.rs
@@ -5,7 +5,7 @@ use std::{
     fmt, iter,
     marker::PhantomData,
     num::NonZeroU32,
-    ops,
+    ops::{Range, RangeFrom, RangeToInclusive},
     path::Path,
     sync::{
         Arc, Condvar, Mutex, Weak,
@@ -84,7 +84,7 @@ impl<CF: NamedColumnFamily> WriteBatch<'_, CF> {
         self.inner.delete_cf(cf, key);
     }
 
-    pub fn delete_range_cf(&mut self, cf: CF, keys: ops::Range<&[u8]>) {
+    pub fn delete_range_cf(&mut self, cf: CF, keys: Range<&[u8]>) {
         let cf = self.db.column_family(cf);
         self.inner.delete_range_cf(cf, keys.start, keys.end);
     }
@@ -657,7 +657,7 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
     pub fn from_iterator_cf(
         &self,
         cf: CF,
-        keys: ops::RangeFrom<&[u8]>,
+        keys: RangeFrom<&[u8]>,
     ) -> impl Iterator<Item = (Box<[u8]>, Box<[u8]>)> + '_ {
         let cf = self.column_family(cf);
         self.inner
@@ -675,12 +675,11 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
     pub fn range_iterator_cf<'a>(
         &'a self,
         cf: CF,
-        from: &'a [u8],
-        to: Vec<u8>,
+        range: Range<&'a [u8]>,
     ) -> impl Iterator<Item = (Box<[u8]>, Box<[u8]>)> + 'a {
         let cf_handle = self.column_family(cf);
         let mut options = ReadOptions::default();
-        options.set_iterate_upper_bound(to);
+        options.set_iterate_upper_bound(range.end.to_vec());
         // When the CF has a prefix extractor configured, this tells RocksDB to use
         // prefix bloom filters for the initial seek — checking each SST file only if
         // it contains keys with the same prefix as the seek key.
@@ -690,7 +689,7 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
             .iterator_cf_opt(
                 cf_handle,
                 options,
-                IteratorMode::From(from, Direction::Forward),
+                IteratorMode::From(range.start, Direction::Forward),
             )
             .map(Result::unwrap)
             .fuse()
@@ -702,7 +701,7 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
     pub fn to_iterator_cf(
         &self,
         cf: CF,
-        keys: ops::RangeToInclusive<&[u8]>,
+        keys: RangeToInclusive<&[u8]>,
     ) -> impl Iterator<Item = (Box<[u8]>, Box<[u8]>)> + '_ {
         let cf = self.column_family(cf);
         self.inner
--- a/lib/rocksdb/src/db.rs
+++ b/lib/rocksdb/src/db.rs
@@ -17,7 +17,8 @@ use std::{
 
 use rocksdb::{
     BlockBasedOptions, Cache, ColumnFamily, ColumnFamilyDescriptor, DB, DBPinnableSlice, Direction,
-    IteratorMode, Options, PrefixRange, ReadOptions, SliceTransform, WriteOptions, perf, properties,
+    IteratorMode, Options, PrefixRange, ReadOptions, SliceTransform, WriteOptions, perf,
+    properties,
 };
 use thread_local::ThreadLocal;
 use vise::MetricsFamily;
@@ -445,7 +446,8 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
             }
 
             let memtable_capacity = options.large_memtable_capacity.filter(|_| requires_tuning);
-            let mut cf_options = Self::rocksdb_options(memtable_capacity, Some(block_based_options));
+            let mut cf_options =
+                Self::rocksdb_options(memtable_capacity, Some(block_based_options));
             if let Some(len) = prefix_extractor_len {
                 cf_options.set_prefix_extractor(SliceTransform::create_fixed_prefix(len));
             }
@@ -685,7 +687,11 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
         options.set_prefix_same_as_start(true);
         self.inner
             .db
-            .iterator_cf_opt(cf_handle, options, IteratorMode::From(from, Direction::Forward))
+            .iterator_cf_opt(
+                cf_handle,
+                options,
+                IteratorMode::From(from, Direction::Forward),
+            )
             .map(Result::unwrap)
             .fuse()
         // ^ unwrap() is safe for the same reasons as in `prefix_iterator_cf()`.
--- a/lib/storage/src/db/repository.rs
+++ b/lib/storage/src/db/repository.rs
@@ -6,7 +6,10 @@ use alloy::{
     primitives::{Address, BlockHash, BlockNumber, TxHash, TxNonce},
     rlp::{Decodable, Encodable},
 };
-use log_index::{chunk_start, deindex_logs, index_logs, rollback_coverage, update_coverage};
+use log_index::{
+    BitmapCache, chunk_start, deindex_logs, flush_bitmap_cache, index_logs, rollback_coverage,
+    update_coverage,
+};
 use std::path::Path;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -21,7 +24,7 @@ use zksync_os_types::{ZkEnvelope, ZkReceiptEnvelope, ZkTransaction};
 
 mod log_index;
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum RepositoryCF {
     // block hash => (block header, array of tx hashes)
     BlockData,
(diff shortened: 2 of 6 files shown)


LogvinovLeon:
## Testing notes

### E2E write-path test

Ran the node locally against the v30.2 chain config, deployed a `Pinger` contract, and sent several transactions that emit events. After the node had produced blocks 5–7 (all containing Pinger logs), we used the new `read-log-index` tool to inspect the RocksDB `log_blocks_by_address` and `log_blocks_by_topic` column families directly. The Pinger address and its topic appeared in blocks `[5, 6, 7]` with `log_index_last_block = 7`, confirming the write path is correct.

### E2E rollback test — bug found

To test the rollback path without relying on timing, we inserted a deliberate panic in `RepositoryManager::populate` immediately after `RepositoryDb::rollback()` returns (before the persist loop can re-fill the DB), and set `force_starting_block_number: 4` in `local_dev.yaml` so the first `populate` call would trigger a rollback to block 3.

After the node crashed at the panic, `read-log-index` showed:

```
log_index_last_block = 3       ✓ (was 7)
Pinger address: blocks [5, 6]  ✗ (should be empty — blocks 4–7 were rolled back)
```

Block 7 was correctly removed, but blocks 5 and 6 were not. **Root cause:** `with_chunk` always read the bitmap from the DB and wrote the result directly into the write batch. When rolling back blocks 5, 6, and 7 in a single batch, each call read the original `{5, 6, 7}` bitmap from the DB (ignoring earlier batch writes), so later writes silently overwrote earlier ones — only block 7's removal survived.

### Fix

Introduced `BitmapCache` — a `HashMap<(RepositoryCF, Vec<u8>), RoaringBitmap>` that acts as a read-your-own-writes overlay for a single batch. `with_chunk` now reads from the cache on subsequent accesses to the same key, accumulating all mutations in memory. Callers flush the cache to the write batch once via `flush_bitmap_cache` before committing. Both `index_logs` and `deindex_logs` use the same path.

### `WriteBatchWithIndex`

RocksDB has a built-in `WriteBatchWithIndex` type that provides `GetFromBatchAndDB` — read-your-own-writes at the raw-bytes level — which would have avoided this class of bug automatically. However, neither `librocksdb-sys` nor the `rocksdb` Rust crate expose it (the `rocksdb_writebatch_wi_*` symbols only appear as internals of the transaction API with no public bindings). Adding it would require writing raw `extern "C"` FFI declarations and a safe wrapper in `zksync_os_rocksdb`. Given that `BitmapCache` solves the problem at a higher level in ~30 lines and also avoids repeated deserialize/serialize cycles on repeated access to the same bitmap, we decided not to pursue the `WriteBatchWithIndex` route.

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,504 @@
...
+pub(super) type BitmapCache = HashMap<(RepositoryCF, Vec<u8>), RoaringBitmap>;
+
+fn with_chunk(
+    db: &RocksDB<RepositoryCF>,
+    cache: &mut BitmapCache,
+    cf: RepositoryCF,
+    key: &[u8],
+    f: impl FnOnce(&mut RoaringBitmap),
+) -> RepositoryResult<()> {
+    let cache_key = (cf, key.to_vec());
```
slowli: I'd much prefer having 2 separate `HashMap`s in cache for the 2 possible CFs. This would be more typesafe and would allow to avoid an awkward allocation here.

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,504 @@
...
+        let mut key = key_prefix.to_vec();
+        key.extend_from_slice(&chunk.to_be_bytes());
+        if let Some(bytes) = db.get_cf(cf, &key)? {
+            result |= deserialize_bitmap(&bytes)?;
+        }
+        chunk += CHUNK_SIZE;
+    }
+
+    // Mask to the requested range (block numbers fit in u32 given current chain sizes).
+    result &= RoaringBitmap::from_iter(range.start as u32..range.end as u32);
```
slowli: Constructing a container with a potentially large range doesn't look efficient; AFAIU, there are no optimizations in the `RoaringBitmap` to efficiently store full ranges specifically. From this perspective, using two [`remove_range()`](https://docs.rs/roaring/latest/roaring/bitmap/struct.RoaringBitmap.html#method.remove_range) calls looks more efficient.

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,504 @@
...
+    while chunk <= to_chunk {
+        let mut key = key_prefix.to_vec();
+        key.extend_from_slice(&chunk.to_be_bytes());
+        if let Some(bytes) = db.get_cf(cf, &key)? {
+            result |= deserialize_bitmap(&bytes)?;
+        }
+        chunk += CHUNK_SIZE;
+    }
+
+    // Mask to the requested range (block numbers fit in u32 given current chain sizes).
```
slowli: Wouldn't the approach break with `u64` block numbers since `RoaringBitmap` doesn't support `u64` values? If so, why use 64-bit chunk keys at all? If it's for future-proofing, I think it makes sense to either to tag values, or store `RoaringBitmap` for block number *offsets* (i.e., relative to the chunk start – in this case, the values in the bitmap cannot overflow `u32` by construction).

lib/storage/src/db/repository.rs
```diff
@@ -184,6 +239,10 @@ impl RepositoryDb {
             &initiator_and_nonce_key,
             tx_hash.as_slice(),
         );
+
+        index_logs(db, bitmap_cache, block_number, chunk, tx.receipt.logs())?;
```
slowli: Given that we always (de)index logs with block granularity, it may be more efficient to first compose addresses / topics for all transactions in the block (essentially, get unique ones in the current impl), and only then touch the relevant bitmaps.

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,504 @@
...
+) -> RepositoryResult<RoaringBitmap> {
+    let from_chunk = chunk_start(range.start);
+    let to_chunk = chunk_start(range.end.saturating_sub(1));
+
+    let mut result = RoaringBitmap::new();
+    let mut chunk = from_chunk;
+    while chunk <= to_chunk {
+        let mut key = key_prefix.to_vec();
+        key.extend_from_slice(&chunk.to_be_bytes());
+        if let Some(bytes) = db.get_cf(cf, &key)? {
```
slowli: It could be possible to use [RocksDB prefix iteration](https://github.com/facebook/rocksdb/wiki/Prefix-Seek) instead here, which would theoretically be more efficient for huge `range`s. (The prefix would be the address / topic, so it would be possible to efficiently rule out non-existing ones.)

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,504 @@
...
+                    bm.insert(block_number);
+                },
+            )?;
+        }
+    }
+    Ok(())
+}
+
+/// Removes all logs from a single transaction from the bitmap cache.
+pub(super) fn deindex_logs<'a>(
```
slowli, line 174: Given there's [a specific method](https://docs.rs/roaring/latest/roaring/bitmap/struct.RoaringBitmap.html#method.remove_range) to remove a range of values from a `RoaringBitmap`, it may be more efficient to do batch removal for the entire range of removed block instead. See also the other comment regarding collecting addresses / topics from the removed blocks in advance.

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+}
+
+/// Updates the log index coverage metadata in `batch`.
+/// `log_index_started` indicates whether `log_index_first_block` is already set in the DB;
+/// if not, it is written now (it is only ever set once).
+pub(super) fn update_coverage(
+    batch: &mut WriteBatch<RepositoryCF>,
+    block_number_bytes: &[u8],
+    log_index_started: bool,
```
slowli: This variable and the corresponding `Arc<AtomicBool>` in the database look too non-local. Wouldn't it be sufficient to read `log_index_first_block_key()` before the write and only perform the write if this key is missing?

lib/storage/src/db/repository.rs
```diff
@@ -31,14 +38,26 @@ pub enum RepositoryCF {
     TxMeta,
     // (initiator address, nonce) => tx hash
     InitiatorAndNonceToHash,
-    // meta fields: currently only latest block number
+    // meta fields: latest block number, log index first/last block
     Meta,
+    // (address[20] ++ chunk_start[8]) => roaring bitmap of block numbers containing logs from that address
```
slowli, line 39: Dumb question: Is *not* saving two least-significant bytes of `chunk_start` (which are always 0, IIUC) not worth the trouble?

slowli (commented):
- Dumb question: Does the topic position really not matter in practice? I.e., is the overhead incurred from splitting topic keys by the topic position (i.e., +1 byte to the key) not justified?
- Have you measured the storage overhead in practice? One thing to consider is that since the write pattern is not append-only (in the worst case, a single RocksDB key would be overwritten 65,536 times) there could be write amplification overhead resulting from persisting multiple versions of the value, at least before compaction (the quantity of which depends on a lot of factors, such as the configured memtable size). Similarly, it could regress read performance somewhat. Since the write pattern is still *roughly* append-only, I *think* the overhead should be manageable, but having concrete numbers would be the best.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Done — replaced the `HashMap<(RepositoryCF, Vec<u8>), RoaringBitmap>` type alias with a struct holding two typed maps (`by_address` / `by_topic`). `RepositoryCF` no longer needs `PartialEq + Eq + Hash`. `flush_bitmap_cache` became `BitmapCache::flush`. `with_chunk` now takes `&mut HashMap<Vec<u8>, RoaringBitmap>` directly, so the CF is no longer part of the cache key.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Fixed — replaced `from_iter(range)` with two `remove_range` calls.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Good catch. Went with the offset approach: bitmaps now store `block_number - chunk_start` rather than the absolute block number. Offsets are always in `[0, 65535]` by construction so they can never overflow `u32`. `read_range` shifts them back to absolute block numbers on the way out.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Agreed. `update_coverage` now checks `db.get_cf(Meta, log_index_first_block_key()).is_none()` directly. The key is written once and then always present in RocksDB's block cache, so the read is essentially free. Dropped the `Arc<AtomicBool>` field, the `log_index_started` parameter to `write_block_inner`, and the load/store in `write_block`.

lib/storage/src/db/repository.rs
LogvinovLeon: Not worth it in my opinion. The current encoding (`prefix ++ chunk_start_as_u64_be`) is self-documenting and maps directly to the code. Saving 2 bytes per key by stripping the always-zero trailing bytes would require a shift/mask on every read and write, and anyone inspecting keys on disk would need to know the value is `chunk_start >> 16` rather than `chunk_start`. The space saving is ~7% of key size and negligible in practice. Happy to revisit if you feel strongly though!

LogvinovLeon:
> Dumb question: Does the topic position really not matter in practice? I.e., is the overhead incurred from splitting topic keys by the topic position (i.e., +1 byte to the key) not justified?

In practice, ignoring position produces zero false positives rather than just "acceptable" ones. Topic[0] is always an event signature hash (e.g. `keccak256("Transfer(address,address,uint256")`), while topic[1..3] are always ABI-encoded arguments (addresses, uint256s, etc.). These two value spaces don't overlap — a 32-byte signature hash will never appear as a padded address or integer at another position in an unrelated event. So a query for `topic[0] = X` won't match a block where `X` appears at topic[1], because no real event produces a signature hash as a non-topic[0] value. The verification pass in PR 3 is a safety net, but it won't actually be needed for this class of false positives.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Done — implemented in 65874d69.

`LogBlocksByAddress` and `LogBlocksByTopic` now have a fixed-prefix `SliceTransform` configured (20 and 32 bytes respectively). `read_range` is rewritten to use a new `range_iterator_cf` helper (upper-bound + `set_prefix_same_as_start(true)`) instead of the per-chunk `get_cf` loop. This means:

- **Sparse address/topic over a wide block range**: previously `read_range` issued one `Get` per chunk in `[from_chunk, to_chunk]`, paying for empty chunks. The iterator now skips them entirely.
- **SST file pruning**: with a prefix extractor set, RocksDB consults per-prefix bloom filters on seek — letting it skip SST files that contain no keys for the queried address/topic at all. One-time O(SST files) check rather than per-chunk.

What doesn't change: single-chunk queries and dense addresses (one iterator step ≈ one Get); rollback path (unchanged).

lib/storage/src/db/repository.rs
LogvinovLeon: Rollback is a manual recovery operation, not on the hot path. Correctness matters more than performance here.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Done — already addressed in the reply above.

LogvinovLeon:
> Have you measured the storage overhead in practice? One thing to consider is that since the write pattern is not append-only (in the worst case, a single RocksDB key would be overwritten 65,536 times) there could be write amplification overhead resulting from persisting multiple versions of the value, at least before compaction (the quantity of which depends on a lot of factors, such as the configured memtable size). Similarly, it could regress read performance somewhat. Since the write pattern is still *roughly* append-only, I *think* the overhead should be manageable, but having concrete numbers would be the best.

Measured. Two data points:

**E2E (200 real log-emitting transactions):** log index CFs total 4,269 B — 3.1% of the repository DB. Amplification 1.0x: all bitmap updates fit within a single memtable flush, so only the final version ever hits disk.

**Synthetic worst case:** one address emitting in every block for a full 65,536-block chunk. The bitmap grows from ~10 B to ~8 KB over the chunk, accumulating ~260 MB of total writes, causing the 64 MB memtable to flush ~4 times mid-chunk. Each flush writes a stale version of the bitmap to an L0 SST file. Pre-compaction amplification: **7.1x**. Post-compaction: **1.0x**.

In practice the ceiling is lower. RocksDB compacts L0 after 4 files (the default `level0_file_num_compaction_trigger`), which for the repository DB at 100 tx/block fires roughly every 2,600 blocks — well within one chunk. So at most 4 stale versions ever accumulate, regardless of how many blocks have been indexed. Realistic pre-compaction amplification for even a very active address: **≤ 4x, transiently**. Always 1x after compaction.

lib/storage/src/db/repository.rs
slowli: Won't insist on it, just wanted to know that this decision is intentional.

lib/storage/src/db/repository/log_index.rs
slowli: I may be missing something obvious, but I don't see these changes in the changeset for this PR. Ditto for the following comment regarding range removal from `RoaringBitmap`s.

lib/rocksdb/src/db.rs
```diff
@@ -654,6 +666,31 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
...
 
+    /// Iterates over key-value pairs in `cf` in lexical order from `from` (inclusive) up to
+    /// `to` (exclusive).  When the column family has a prefix extractor configured and `from` and
+    /// `to` share the same extracted prefix, RocksDB uses prefix bloom filters to skip SST files
+    /// that contain no keys with that prefix.
+    pub fn range_iterator_cf<'a>(
+        &'a self,
+        cf: CF,
+        from: &'a [u8],
+        to: Vec<u8>,
```
slowli: Bikeshedding: using a slice vs `Vec` for bounds looks inconsistent. I'd suggest to use a `Range<&[u8]>` to clearly convey that the `to` bound is exclusive.

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,537 @@
...
+            .expect("chunk key suffix must be 8 bytes");
+        let chunk_base = u64::from_be_bytes(chunk_bytes) as u32;
+        result |= deserialize_bitmap(&value)?
+            .into_iter()
+            .map(|offset| chunk_base + offset)
+            .collect::<RoaringBitmap>();
+    }
+
+    result.remove_range(..range.start as u32);
+    result.remove_range(range.end as u32..);
```
slowli, line 263: Micro-nit: Could make sense to add a comment here explaining the reasoning (efficiency).

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,537 @@
...
+    Ok(())
+}
+
+/// Removes all logs from a single transaction from the bitmap cache.
+/// `block_offset` is the block number relative to `chunk` (i.e. `block_number - chunk`).
+pub(super) fn deindex_logs<'a>(
+    db: &RocksDB<RepositoryCF>,
+    cache: &mut BitmapCache,
+    block_offset: u32,
+    chunk: u64,
```
slowli: Bikeshedding: Since correctness is preferred over efficiency elsewhere (which I agree with), it would make sense to pass the block number here and split into into the chunk start and offset inside the method, so that it cannot be misused. Ditto for `index_logs()` above.

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,537 @@
...
+        };
+        bitmaps.entry(key.to_vec()).or_insert(bm)
+    };
+    f(bitmap);
+    Ok(())
+}
+
+/// Updates the log index coverage metadata in `batch`.
+/// Writes `log_index_first_block` if it is not already present in the DB
+/// (it is only ever set once); always updates `log_index_last_block`.
```
slowli, line 105: Nit: Could be worth specifying that this logic only works if `update_coverage()` is called once with the specified `batch`; otherwise, it's prone to the same issue you've fixed previously (changes not read back from the write batch).

lib/storage/src/db/repository/log_index.rs
```diff
@@ -0,0 +1,537 @@
...
+            )?;
+        }
+    }
+    Ok(())
+}
+
+fn read_u64_meta(db: &RocksDB<RepositoryCF>, key: &[u8]) -> RepositoryResult<Option<u64>> {
+    Ok(db
+        .get_cf(RepositoryCF::Meta, key)?
+        .map(|v: Vec<u8>| u64::from_be_bytes(v.as_slice().try_into().unwrap())))
```
slowli: Bikeshedding: Looks slightly inconsistent that bitmap deserialization errors have their own variant in `RepositoryError`, but deserialization errors here and in similar places are converted to panics. AFAIU, in both cases, the only possible cause of an error would be DB corruption.

slowli (commented):
Overall, looks pretty good 👍

lib/rocksdb/src/db.rs
LogvinovLeon: Done — changed to `Range<&'a [u8]>`.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Done — added a doc note.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Done — both functions now take `block_number: u64` and compute chunk/offset internally.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Done — added a comment.

lib/storage/src/db/repository/log_index.rs
LogvinovLeon: Done — changed `deserialize_bitmap` to panic on failure (like other DB corruption cases in the codebase), and removed the `BitmapDeserialize` error variant entirely. Also added `expect` message to `read_u64_meta` for consistency.

lib/rocksdb/src/db.rs
```diff
@@ -654,12 +668,40 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
         // ^ unwrap() is safe for the same reasons as in `prefix_iterator_cf()`.
     }
 
+    /// Iterates over key-value pairs in `cf` in lexical order from `from` (inclusive) up to
+    /// `to` (exclusive).  When the column family has a prefix extractor configured and `from` and
+    /// `to` share the same extracted prefix, RocksDB uses prefix bloom filters to skip SST files
+    /// that contain no keys with that prefix.
```
slowli: Micro-nit: The doc comment is outdated now that `range` is used instead of separate `from` and `to` args.

slowli (dismissed):

lib/rocksdb/src/db.rs
LogvinovLeon: Fixed.

slowli (approved):

PR #1160, opened 2026-04-09
feat: async batch provider init

perekopskiy:
## Summary

Resolves TODO for CommittedBatchProvider, now batches that were already committed on start-up are loaded in background and other components await them with `wait_for_batch`. `[last_committed_batch, last_proved_batch, last_executed_batch]` these batches should be loaded first because we need them for `commit_proof_execute_block_numbers` during node start-up.

### DIFF
--- a/lib/l1_watcher/src/committed_batch_provider.rs
+++ b/lib/l1_watcher/src/committed_batch_provider.rs
@@ -121,6 +121,12 @@ impl CommittedBatchProvider {
         }
     }
 
+    /// Returns `DiscoveredCommittedBatch` from in-memory map if available.
+    pub fn get(&self, batch_number: u64) -> Option<DiscoveredCommittedBatch> {
+        let inner = self.inner.read().expect("lock poisoned");
+        inner.batches.get(&batch_number).cloned()
+    }
+
     /// Fetches a batch set with bounded concurrency to reduce startup latency without issuing an
     /// unbounded number of L1 requests.
     async fn load_batch_numbers(
--- a/lib/l1_watcher/src/committed_batch_provider.rs
+++ b/lib/l1_watcher/src/committed_batch_provider.rs
@@ -4,7 +4,7 @@ use alloy::providers::DynProvider;
 use anyhow::Context;
 use futures::stream::{self, StreamExt};
 use rangemap::RangeInclusiveMap;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::sync::{Arc, RwLock};
 use std::time::Duration;
 use tokio::time::sleep;
@@ -73,21 +73,25 @@ impl CommittedBatchProvider {
     /// soon as possible. The remaining committed batches are fetched afterwards, with bounded
     /// parallelism.
     pub async fn init(&self, l1_state: &L1State, max_l1_blocks_to_scan: u64) -> anyhow::Result<()> {
-        self.load_batches_in_background(
+        let (prioritized_batch_numbers, remaining_batch_numbers) = startup_batch_numbers(
+            l1_state.last_committed_batch,
+            l1_state.last_proved_batch,
+            l1_state.last_executed_batch,
+        );
+
+        self.load_batch_numbers(
             l1_state.diamond_proxy_sl.clone(),
             max_l1_blocks_to_scan,
-            startup_priority_batch_numbers(
-                l1_state.last_committed_batch,
-                l1_state.last_proved_batch,
-                l1_state.last_executed_batch,
-            ),
-            startup_remaining_batch_numbers(
-                l1_state.last_committed_batch,
-                l1_state.last_proved_batch,
-                l1_state.last_executed_batch,
-            ),
+            prioritized_batch_numbers,
         )
-        .await
+        .await?;
+        self.load_batch_numbers(
+            l1_state.diamond_proxy_sl.clone(),
+            max_l1_blocks_to_scan,
+            remaining_batch_numbers,
+        )
+        .await?;
+        Ok(())
     }
 
     pub(crate) fn insert(&self, batch: DiscoveredCommittedBatch) {
@@ -116,30 +120,6 @@ impl CommittedBatchProvider {
             }
             sleep(WAIT_FOR_BATCH_POLL_INTERVAL).await;
         }
-    }
-
-    /// Fills the provider in two phases so startup-critical frontier batches become available
-    /// before the rest of the historical committed range.
-    async fn load_batches_in_background(
-        &self,
-        diamond_proxy_sl: ZkChain<DynProvider>,
-        max_l1_blocks_to_scan: u64,
-        prioritized_batch_numbers: Vec<u64>,
-        remaining_batch_numbers: Vec<u64>,
-    ) -> anyhow::Result<()> {
-        self.load_batch_numbers(
-            diamond_proxy_sl.clone(),
-            max_l1_blocks_to_scan,
-            prioritized_batch_numbers,
-        )
-        .await?;
-        self.load_batch_numbers(
-            diamond_proxy_sl,
-            max_l1_blocks_to_scan,
-            remaining_batch_numbers,
-        )
-        .await?;
-        Ok(())
     }
 
     /// Fetches a batch set with bounded concurrency to reduce startup latency without issuing an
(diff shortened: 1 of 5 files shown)


RomanBrodetski (commented):
Can you please add some doc comments with logic explanation etc? Especially the logic behind  priority/remaining batch numbers

node/bin/src/lib.rs
```diff
@@ -1360,8 +1364,8 @@ async fn commit_proof_execute_block_numbers(
         0
     } else {
         committed_batch_provider
-            .get(l1_state.last_committed_batch)
-            .expect("last committed batch was not discovered on L1")
+            .wait_for_batch(l1_state.last_committed_batch)
+            .await
```
kostia244: I'm pretty sure this can cause a shutdown bug, if committed batch provider gets an RPC error.

The easiest way to fix this is to select with tokio::select! with a shutdown signal, but I'd be happy if there is a cleaner way

lib/l1_watcher/src/committed_batch_provider.rs
```diff
@@ -36,56 +45,105 @@ impl CommittedBatchProvider {
...
+            sleep(WAIT_FOR_BATCH_POLL_INTERVAL).await;
+        }
+    }
+
+    async fn load_batches_in_background(
+        &self,
+        diamond_proxy_sl: ZkChain<DynProvider>,
+        max_l1_blocks_to_scan: u64,
+        prioritized_batch_numbers: Vec<u64>,
+        remaining_batch_numbers: Vec<u64>,
```
kostia244: Can't you call this function twice instead of passing two vectors? IMO having more generic function here is better.

lib/l1_watcher/src/committed_batch_provider.rs
```diff
@@ -96,3 +172,70 @@ impl Inner {
...
+    [last_committed_batch, last_proved_batch, last_executed_batch]
+        .into_iter()
+        .filter(|batch_number| *batch_number > 0)
+        .filter(|batch_number| seen.insert(*batch_number))
+        .collect()
+}
+
+/// Returns the rest of the startup committed range after removing the frontier batches that are
+/// loaded first.
+fn startup_remaining_batch_numbers(
```
kostia244: How about one function that returns two vectors and uses .partition instead of filter?

lib/l1_watcher/src/committed_batch_provider.rs
```diff
@@ -36,56 +49,119 @@ impl CommittedBatchProvider {
...
+                tracing::info!("waiting for committed batch {batch_number} to load");
+                logged_wait = true;
+            }
+            sleep(WAIT_FOR_BATCH_POLL_INTERVAL).await;
+        }
+    }
+
+    /// Fills the provider in two phases so startup-critical frontier batches become available
+    /// before the rest of the historical committed range.
+    async fn load_batches_in_background(
```
kostia244: In fact this is a pass-through function, maybe just inline it?

lib/l1_watcher/src/committed_batch_provider.rs
```diff
@@ -20,12 +28,17 @@ struct Inner {
 }
 
 impl CommittedBatchProvider {
```
RomanBrodetski, line 37: not directly rtelated to your PR, but I'd love to have a doc comment on what this provider actually does. Is this the one that's persisting things? Do qwe have other related but different providers?

lib/l1_watcher/src/committed_batch_provider.rs
```diff
@@ -36,56 +49,119 @@ impl CommittedBatchProvider {
...
-        })
+        Ok(provider)
+    }
+
+    /// Loads historical committed batches discovered on startup.
+    ///
+    /// The three frontier batches used by startup bookkeeping are fetched first so callers waiting
+    /// on `last_committed_batch`, `last_proved_batch`, or `last_executed_batch` can proceed as
+    /// soon as possible. The remaining committed batches are fetched afterwards, with bounded
+    /// parallelism.
```
RomanBrodetski: very nice and clear comment thanks!

lib/l1_watcher/src/committed_batch_provider.rs
kostia244: It would be nice to have these 2 things stated explicitly for each watcher:
1. What data it fetches? - some watchers fetch/store same information, but visibility into this is kinda low unless you keep all 10 of them in your head.
2. Who are the consumers? (main pipeline vs EN vs RPC ..)

If nobody else does it, I'll add it to my backlog as I'm already looking into L1 RPC reads.

node/bin/src/batcher/mod.rs
```diff
@@ -136,13 +130,8 @@ impl<ReadState: ReadStateHistory + Clone + Send + 'static> PipelineComponent
...
                     let committed_batch = self
                         .committed_batch_provider
-                        .get(prev_batch_info.batch_number + 1)
-                        .with_context(|| {
-                            format!(
-                                "committed batch {} must have been discovered on L1",
-                                prev_batch_info.batch_number + 1
-                            )
-                        })?;
+                        .wait_for_batch(prev_batch_info.batch_number + 1)
```
RomanBrodetski, line 133: I'm just thinking.. if there is a bug or some L1 issue it may be hard to say why node is hanging, right? Can we do something with it? One option is to simply add an. `INFO` log inside `wait_for_batch` that would only print if batch is not available yet (this way we only print when we need to wait and don't pullte general logs)

RomanBrodetski (commented):
Please consider the minor comments I added

lib/l1_watcher/src/committed_batch_provider.rs
perekopskiy: it's not the one that persists, I will add comment

node/bin/src/batcher/mod.rs
perekopskiy: there is already such log, I can add one more log on return, this way it will be totally clear that this the reason for node handing

RomanBrodetski (dismissed):

lib/l1_watcher/src/committed_batch_provider.rs
perekopskiy: @kostia244 makes sense, I can add in a separate PR

node/bin/src/lib.rs
perekopskiy: Nice catch, fixed

node/bin/src/lib.rs
```diff
@@ -302,21 +302,33 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+            .unwrap()
+    })
     .await
     .expect("failed to init CommittedBatchProvider");
 
+    let committed_batch_provider_for_init = committed_batch_provider.clone();
+    let l1_state_for_init = l1_state.clone();
+    let max_blocks_to_process = config.l1_watcher_config.max_blocks_to_process;
+    let runtime_for_init = runtime.clone();
+    runtime.spawn_critical_with_shutdown_signal("committed batch provider init", |shutdown| async move {
```
kostia244: This doesn't work. It You can test by adding anyhow::ensure(false); in init()

kostia244: Note that you are starting tasks AFTER receiving shutdown signal, this is probably the reason

node/bin/src/lib.rs
```diff
@@ -302,21 +302,33 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+        tokio::pin!(shutdown);
+
+        tokio::select! {
+            result = committed_batch_provider_for_init.init(&l1_state_for_init, max_blocks_to_process) => {
+                if let Err(err) = result {
+                    tracing::error!(%err, "failed to initialize CommittedBatchProvider");
+                    let _ = runtime_for_init.initiate_graceful_shutdown();
```
kostia244: IMO we should only send shutdown signal in one place to avoid confusion. Perhaps just changing this part would fix everything.

node/bin/src/lib.rs
```diff
@@ -1353,37 +1367,63 @@ fn check_batch_verification_mismatch(
...
+    runtime: &Runtime,
+    committed_batch_provider: &CommittedBatchProvider,
+    batch_number: u64,
+) -> anyhow::Result<zksync_os_batch_types::DiscoveredCommittedBatch> {
+    let shutdown = runtime.on_shutdown_signal().clone();
+    tokio::pin!(shutdown);
+
+    tokio::select! {
+        batch = committed_batch_provider.wait_for_batch(batch_number) => Ok(batch),
+        _ = &mut shutdown => anyhow::bail!(
```
kostia244: Note that if we decide to only send shutdown signal after run finishes.
Run doesn't return an error, so this would cause a panic I think? I'm not sure if the other tasks would shut down this way

kostia244: Honestly I would initialize priority batches in a blocking way, because it's much easier and you end up waiting for them very shorly anyway

perekopskiy: agree, I think it resolves other comments as well, right?

kostia244: Sorry, I miscalculated. If we only send shutdown signal in main, this literally does not get the shutdown signal until we exit run. So using this in run is not better than using the raw function. We'll keep waiting forever anyway.

I wouldn't want to discuss a proper fix in here as it's probably out of scope. Maybe let's just get these batches without retries(assert that they are present?).

perekopskiy: done

kostia244 (approved):
Looks fine, I assume you have tested the crash case.

PR #1166, opened 2026-04-10
fix: record sl block used for l1 state fetch

### LINKED ISSUE #1164: fix: EN crash-loop from L1 watcher race between commit and execute watchers
## Summary

External Nodes can enter a crash-loop when the L1 commit watcher's historical-commit skip guard (`should_skip_historical_commit`) incorrectly drops a legitimate batch commit event. This leaves the `CommittedBatchProvider` without the batch data that the execute watcher needs, causing a `BatchNotCommitted` panic that kills the process.

## Root Cause

There is a race condition between the commit watcher, execute watcher, and `CommittedBatchProvider::init()` during EN restarts.

### The specific chain of events (observed 2026-04-09):

1. **EN crashes** for any reason (in this case, `BatchNotCommitted(74530)` from a prior timing issue)
2. **Between crash and restart**, batch 74531 gets committed on L1 (block 10622613)
3. **EN restarts** — L1 contract state returns `last_committed_batch: 74530` (the query just barely misses 74531)
4. `CommittedBatchProvider::init()` pre-loads batches `74530..=74530` — batch 74531 is outside this range
5. Commit watcher sets `startup_latest_l1_block: 10622614`, `startup_last_committed_batch: 74530`
6. **Commit watcher encounters batch 74531's commit event** on L1 block 10622613
7. `should_skip_historical_commit` evaluates:
   - `log_block_number (10622613) <= startup_latest_l1_block (10622614)` → true
   - `batch_number (74531) > startup_last_committed_batch (74530)` → true
   - **Result: SKIP** — "likely reverted before startup"
8. **Batch 74531 falls through a crack** — not in `init()`, not in commit watcher
9. Execute watcher sees batch 74531 executed → `committed_batch_provider.get(74531)` → `None` for 30s → `BatchNotCommitted(74531)` → **panic**
10. Docker restarts EN → same cycle repeats until L1 state advances enough

### Why it eventually self-resolves

After enough restarts, `last_committed_batch` and `last_executed_batch` both reach 74531 in the L1 contract state, so `init()` pre-loads batch 74531 directly. No gap, no crash.

### Log evidence

```
10:43:02 CRITICAL: watcher failed: BatchNotCommitted(74530) → crash #1
10:43:59 initializing L1 commit watcher: last_committed_batch=74530, current_l1_block=10622614
10:44:09 WARN: skipping historical committed batch above startup frontier (batch 74531, L1 block 10622613)
10:45:10 CRITICAL: watcher failed: BatchNotCommitted(74531) → crash #2
10:45:22 L1 state: last_committed_batch=74531, last_executed_batch=74531
10:45:26 discovered committed batch on startup: batch 74531 → no crash
```

## The bug in `should_skip_historical_commit`

Introduced in #1096 to handle **reverted** batches — commits that landed on L1 then got rolled back before the node started. The guard assumes any commit event in the historical L1 range for a batch above the startup frontier must be stale/reverted.

This assumption breaks when a batch is **legitimately committed** on L1 between the EN crash and restart. The commit event is in the historical range and above the startup frontier, but it's valid — not reverted.

## Secondary issue: panic on `BatchNotCommitted`

`watcher.rs:51-53` treats `BatchNotCommitted` as fatal:

```rust
if let Err(e) = self.poll().await {
    tracing::error!("l1 watcher fatal error: {e}");
    panic!("watcher failed: {e}");
}
```

`BatchNotCommitted` is a timing issue, not a data integrity problem. It should not kill the process.

## Proposed Fix

1. **In `should_skip_historical_commit`**: when the guard triggers, verify against L1 contract state (`storedBatchHash(batch_number)`) before skipping. If the hash is non-zero, the batch is still committed — process it normally. If zero, it was truly reverted — skip it. This preserves #1096's revert handling while fixing the false-positive skip.

2. **In `watcher.rs`**: make `BatchNotCommitted` a recoverable error that retries on the next poll cycle instead of panicking.

## Impact

- Affects External Nodes only (not the main node)
- Causes EN downtime during the crash-loop (2+ minutes of unavailability)
- Self-resolves eventually but produces unnecessary restarts and block sync gaps
- Triggered by any EN restart that coincides with L1 batch activity

perekopskiy:
## Summary

  - Record the settlement-layer block number used to initialize L1State.
  - Fetch committed, proved, and executed batch counters from the same SL snapshot.
  - Pass that startup SL block into the commit watcher and use it to filter out historical commit events for batches that were already reverted before startup.

###  Why

  - Startup finality should be derived from one consistent SL view instead of mixing values from different tips.
  - The commit watcher should compare historical events against the same snapshot that initialized finality, so that it doesn't skip valid events in case of race condition (see issue below for more details).
  
Fixes https://github.com/matter-labs/zksync-os-server/issues/1164

### DIFF
--- a/lib/contract_interface/src/l1_discovery.rs
+++ b/lib/contract_interface/src/l1_discovery.rs
@@ -6,7 +6,7 @@ use alloy::primitives::{Address, U256, address};
 use alloy::providers::{DynProvider, Provider};
 use anyhow::Context;
 use backon::{ConstantBuilder, Retryable};
-use std::fmt::{Debug, Display};
+use std::fmt::Debug;
 use std::time::Duration;
 
 const L2_BRIDGEHUB_ADDRESS: Address = address!("0x0000000000000000000000000000000000010002");
@@ -34,9 +34,18 @@ pub struct L1State {
     pub last_committed_batch: u64,
     pub last_proved_batch: u64,
     pub last_executed_batch: u64,
+    /// Block number on SL that was used to query `last_committed_batch`, `last_proved_batch`, `last_executed_batch`.
+    pub sl_block_number: u64,
     pub da_input_mode: BatchDaInputMode,
     pub l1_chain_id: u64,
     pub sl_chain_id: u64,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct BatchFinality {
+    last_committed_batch: u64,
+    last_proved_batch: u64,
+    last_executed_batch: u64,
 }
 
 impl L1State {
@@ -86,10 +95,16 @@ impl L1State {
         let diamond_proxy_sl = bridgehub_sl.zk_chain().await?;
         let validator_timelock_sl = bridgehub_sl.validator_timelock_address().await?;
 
-        let latest = BlockId::latest();
-        let last_committed_batch = diamond_proxy_sl.get_total_batches_committed(latest).await?;
-        let last_proved_batch = diamond_proxy_sl.get_total_batches_proved(latest).await?;
-        let last_executed_batch = diamond_proxy_sl.get_total_batches_executed(latest).await?;
+        let latest_sl_block_number = diamond_proxy_sl.provider().get_block_number().await?;
+        let last_committed_batch = diamond_proxy_sl
+            .get_total_batches_committed(latest_sl_block_number.into())
+            .await?;
+        let last_proved_batch = diamond_proxy_sl
+            .get_total_batches_proved(latest_sl_block_number.into())
+            .await?;
+        let last_executed_batch = diamond_proxy_sl
+            .get_total_batches_executed(latest_sl_block_number.into())
+            .await?;
 
         let pubdata_pricing_mode = diamond_proxy_sl.get_pubdata_pricing_mode().await?;
         let da_input_mode = match pubdata_pricing_mode {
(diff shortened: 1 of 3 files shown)


lib/contract_interface/src/l1_discovery.rs
```diff
@@ -34,11 +34,20 @@ pub struct L1State {
...
     pub l1_chain_id: u64,
     pub sl_chain_id: u64,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+struct BatchFinality {
+    last_committed_batch: u64,
+    last_proved_batch: u64,
+    last_executed_batch: u64,
+}
```
RomanBrodetski, line 49: Shall we maybe reuse this in `L1State` itself? It has these three fields as well.

RomanBrodetski (approved):
Approved, but please consider the optional comment

lib/contract_interface/src/l1_discovery.rs
perekopskiy: Why would we? What's the point?

RomanBrodetski: I'm not saying it's large but:

Image
this will be unnesearry

And also, we'll have two less fields in this strcut - less cognitive load when reading this code. And I don't see a downside.

perekopskiy: But then every access becomes `l1_state.<smth>.last_committed_batch`, IMO it's better to keep this struct flattened. This struct was added for internal method to not return a tuple of 3 ints but a struct

PR #1170, opened 2026-04-14
refactor: Simplify batch metadata and related types

kostia244:
### Main points:
- Remove unused fields from BatchMetadata.
- Get rid of BatchInfo.
- Rename & move CommittedBatch to batcher_types.
- Move batcher_model to batcher_types crate
- Create batcher_metrics crate and put batcher metrics there

### Justification

BatchInfo makes little sense to me as a data structure. It has a couple useful functions but the data flow is not right: most of the fields act like BatchMetadata members instead.

As for the functions, into_stored() was really made for CommittedBatch so it makes sense to move it there.


lib/batch_types/src/batcher_model.rs
```diff
@@ -20,25 +21,24 @@ use zksync_os_types::{ProtocolSemanticVersion, ProvingVersion};
...
 ///
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct BatchMetadata {
     pub previous_stored_batch_info: StoredBatchInfo,
     // This is not purely commitment information, but we keep old serialization name for
     // backwards-compatibility.
     #[serde(rename = "commit_batch_info")]
-    pub batch_info: BatchInfo,
+    pub batch_info: CommitBatchInfoExt,
+    pub chain_address: Address,
```
kostia244, line 34: Keeping it here only because of senders -- I could pass it to sol_cal but then the argument list gets bloated, it's a separate issue IMO.

lib/l1_sender/src/batcher_model.rs
```diff
@@ -20,25 +21,24 @@ use zksync_os_types::{ProtocolSemanticVersion, ProvingVersion};
...
+///  This isn't really a problem as we only store the recent ones
 ///
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct BatchMetadata {
     pub previous_stored_batch_info: StoredBatchInfo,
     // This is not purely commitment information, but we keep old serialization name for
     // backwards-compatibility.
     #[serde(rename = "commit_batch_info")]
-    pub batch_info: BatchInfo,
+    pub batch_info: CommitBatchInfoExt,
```
kostia244: Should I implement deref to make access easier?

lib/batch_types/src/batcher_model.rs
```diff
@@ -1,16 +1,17 @@
...
 use time::UtcDateTime;
-use zksync_os_batch_types::{BatchInfo, BatchSignatureSet};
+use zksync_os_batch_types::{BatchSignatureSet, CommitBatchInfoExt};
 use zksync_os_contract_interface::models::{L2Log, StoredBatchInfo};
 use zksync_os_observability::LatencyDistributionTracker;
+use zksync_os_types::ProvingVersion;
 use zksync_os_types::PubdataMode;
-use zksync_os_types::{ProtocolSemanticVersion, ProvingVersion};
 // todo: these models are used throughout the batcher subsystem - not only l1 sender
 //       we will move them to `types` or `batcher_types` when an analogous crate is created in `zksync-os`
```
kostia244, line 16: I don't understand this fully, is it already time to move it?

lib/l1_sender/src/batcher_model.rs
itegulov: I'd implement `Deref` yeah

lib/batch_types/src/batcher_model.rs
itegulov: Yes, the comment was written loooong before `batch_types` crate was even a thing. Let's move it there if possible.

itegulov (dismissed):
LGTM overall

lib/batch_types/src/batch_info.rs
```diff
@@ -14,38 +14,36 @@ use zksync_os_types::{
 
 const PUBDATA_SOURCE_CALLDATA: u8 = 0;
 
+/// Commitment information about a batch.
+/// Contains enough data to restore `StoredBatchInfo` that got applied on-chain.
+/// Contains enough data to construct public input hash.
+/// todo: these fields should be a part of `CommitBatchInfo` but needs to be changed on L1 contracts' side first
 #[derive(Clone, Serialize, Deserialize, Debug)]
-pub struct BatchInfo {
+pub struct CommitBatchInfoExt {
```
slowli: Bikeshedding: `_Ext` suffix makes it look this is an extension trait. Maybe, use `ExtendedCommitBatchInfo`?

lib/l1_sender/src/batcher_model.rs
slowli: Bikeshedding: Personally, I agree with *Rust API Guidelines* that [`Deref` / `DerefMut` should be implemented for smart pointers only](https://rust-lang.github.io/api-guidelines/predictability.html#only-smart-pointers-implement-deref-and-derefmut-c-deref), since otherwise, reasoning about field accesses / method calls quickly becomes a mess. But this rule is already violated in the codebase, so 🤷

lib/batch_types/src/batch_info.rs
```diff
@@ -14,38 +14,36 @@ use zksync_os_types::{
...
+impl CommitBatchInfoExt {
     #[allow(clippy::too_many_arguments)]
-    pub fn new(
+    pub fn build(
         blocks: Vec<(
             &BlockOutput,
             &BlockContext,
             &[ZkTransaction],
             &zksync_os_merkle_tree::TreeBatchOutput,
         )>,
```
slowli, line 39: Not sure whether this is in the scope of the PR, but this looks like too much data supplied, since `BlockContext` and `TreeBatchOutput` are only read for the last block.

lib/l1_sender/src/commands/execute.rs
```diff
@@ -145,7 +146,12 @@ impl ExecuteCommand {
             }
             _ => panic!(
                 "Unsupported protocol version: {}",
-                self.batches.first().unwrap().batch.protocol_version
+                self.batches
+                    .first()
+                    .unwrap()
+                    .batch
+                    .batch_info
+                    .protocol_version
```
slowli: Micro-nit: You can extract this to a local var to use in the `match`ed expression so that it doesn't need to be duplicated here.

lib/l1_sender/src/batcher_model.rs
itegulov: Oh, I did not realize guidelines have such a strong wording on this. I have this pattern be used in a few codebases, interesting. I would not do that then, yeah

kostia244: This is used for several structs which aren't related to this. Perhaps I can make one more PR to get rid of this trait for all of them.

lib/batch_types/src/batch_info.rs
kostia244: I'd say a bit out of scope, I'll keep this in my backlog

kostia244:
@itegulov can you take a look at the new changes?

itegulov (approved):

PR #1172, opened 2026-04-14
feat(l1-sender): In-flight tx detection & minor enhancements

Artemka374:
## Summary                                                                                                                                                                                                                                                                                          
                                                                                                                                  
  - **In-flight transaction recovery on startup.** Previously, if the L1 sender was restarted while transactions were still pending in the mempool, it would crash when they eventually got mined. The sender now detects and resumes tracking any such transactions, falling back to the normal send 
  path if recovery isn't possible (e.g. unsupported RPC method, dropped mempool transaction).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                

 - **eth_getTransactionBySenderAndNonce** RPC method updated to also look in the mempool, so the feature also works if the SL is Gateway

## IMPORTANT NOTE

The recovery only works with Reth nodes v1.0.6(or higher), as other providers doesn't support `eth_getTransactionBySenderAndNonce` which is able to check the mempool

### Follow-ups
 - Revisit fee params estimation
 - Increase required confirmations to 3. Initialliy planned to be implemented in this PR, but seems there are some issues with tests
 - _Probably_ allow force-replacing the transactions

### DIFF
--- a/lib/l1_sender/src/lib.rs
+++ b/lib/l1_sender/src/lib.rs
@@ -39,9 +39,6 @@ use zksync_os_pipeline::PeekableReceiver;
 
 /// A code for "method not found" error response as declared in JSON-RPC 2.0 spec.
 const METHOD_NOT_FOUND_CODE: i64 = -32601;
-/// Estimated max amount of gas consumed by transaction sent by L1 sender is ~500k.
-/// We set the limit higher to be safe.
-const MAX_TX_GAS_USED: u64 = 2_000_000;
 /// Number of L1 confirmations required before a transaction is considered final.
 const REQUIRED_CONFIRMATIONS: u64 = 1;
 
@@ -198,6 +195,7 @@ pub async fn run_l1_sender<Input: SendToL1>(
                         operator_address,
                         config.max_fee_per_gas_wei,
                         config.max_priority_fee_per_gas_wei,
+                        config.gas_limit,
                     )
                     .await?
                     .with_to(to_address)
@@ -531,6 +529,7 @@ async fn tx_request_with_gas_fields(
     operator_address: Address,
     max_fee_per_gas: u128,
     max_priority_fee_per_gas: u128,
+    gas_limit: u64,
 ) -> anyhow::Result<TransactionRequest> {
     let eip1559_est = provider.estimate_eip1559_fees().await?;
     L1_SENDER_METRICS.report_l1_eip_1559_estimation(eip1559_est)?;
@@ -569,7 +568,7 @@ async fn tx_request_with_gas_fields(
         .with_from(operator_address)
         .with_max_fee_per_gas(capped_max_fee_per_gas)
         .with_max_priority_fee_per_gas(capped_max_priority_fee_per_gas)
-        .with_gas_limit(MAX_TX_GAS_USED);
+        .with_gas_limit(gas_limit);
     Ok(tx)
 }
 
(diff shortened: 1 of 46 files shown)


lib/l1_sender/src/lib.rs
```diff
@@ -259,6 +311,145 @@ pub async fn run_l1_sender<Input: SendToL1>(
...
+    #[derive(Debug, serde::Deserialize)]
+    struct TxResponse {
+        hash: alloy::primitives::B256,
+        input: alloy::primitives::Bytes,
+    }
+
+    // Probe whether the provider supports `eth_getTransactionBySenderAndNonce` before
+    // iterating over all pending nonces.
+    if let Err(TransportError::ErrorResp(ref e)) = provider
+        .raw_request::<_, Option<TxResponse>>(
```
itegulov: There is existing `get_transaction_by_sender_nonce` that you can use instead

lib/l1_sender/src/lib.rs
```diff
@@ -259,6 +311,145 @@ pub async fn run_l1_sender<Input: SendToL1>(
...
+            );
+            return Ok(vec![]);
+        }
+        anyhow::bail!("Error while probing eth_getTransactionBySenderAndNonce support: {e}");
+    }
+
+    // For each pending nonce, fetch the in-flight tx then peek at the next queued command.
+    // If the command's calldata matches what is on-chain, consume and pair it. On the first
+    // mismatch, stop — the unmatched command stays in `inbound` and will be re-sent by the
+    // normal send path (replacing the in-flight tx at that nonce).
```
itegulov, line 435: Hmm, there seems to be a race condition between batcher that relies on last committed batch as discovered during startup and this component that might be doing recovery at a slightly different point in time.

We need a way to align them somehow. Either make this logic resistant to receiving older calldata that is no longer in-flight. Or moving this logic somewhere earlier (but even then I am not sure if we'd be able to avoid race condition).

lib/l1_sender/src/lib.rs
```diff
@@ -259,6 +311,145 @@ pub async fn run_l1_sender<Input: SendToL1>(
...
+    let mut paired = Vec::with_capacity(in_flight_count);
+    for nonce in latest_nonce..pending_nonce {
+        let tx = match provider
+            .raw_request::<_, Option<TxResponse>>(
+                "eth_getTransactionBySenderAndNonce".into(),
```
itegulov: Same thing about `get_transaction_by_sender_nonce` here

lib/l1_sender/src/lib.rs
```diff
@@ -111,21 +151,32 @@ pub async fn run_l1_sender<Input: SendToL1>(
...
         latency_tracker.enter_state(L1SenderState::SendingToL1);
         let range = Input::display_range(&commands); // Only for logging
         tracing::info!(command_name, range, "sending L1 transactions");
         L1_SENDER_METRICS.parallel_transactions[&command_name].set(commands.len() as u64);
+
+        // On the first iteration, create receipt watchers for any in-flight transactions
+        // recovered on startup and prepend them. Their nonces are lower than the ones we
+        // are about to assign, so they must be first in the ordering. On all subsequent
+        // iterations `recovered` is empty and this produces nothing.
+        let mut pending_txs: Vec<(TransactionReceiptFuture, Input, Instant)> = recovered
```
itegulov: I think it's reasonable to wait for all in-flight transactions first before entering this loop. Current state of code is becoming very convoluted... We have `pending_txs`, `new_txs`, `recovered`, `commands` etc. And all of them are mutable and hard to reason about.

lib/l1_sender/src/lib.rs
Artemka374: Should be fixed now by querying last_{command}_batch and transaction_count(latest) at the same block number - then we can query pending transactions separately and handle such gracefully

Artemka374:
Initial gas limit was set to 2M gas, but seems like sometimes it's not enough for interop tests. Increased to 5M by default(3 and 4 didn't work as well) and made the value configurable - so we can change it further if needed

itegulov (approved):

PR #1175, opened 2026-04-15
feat: allow l1 poll interval to be configured

### LINKED ISSUE #1176: Allow l1 provider polling time to be configured to have faster batches in development environments
# Problem

When zksync-os server is used as a development dependency the time to close a batch is to slow. That causes tests to take a lot of time. There should a way to have faster batches for development.

# Solution

The slowest part to close a batch seems to be polling l1. At the moment the codebase is using the default polling time for alloy, which is 7 seconds. That time should be possible to configure, to have faster batches in development.

calvogenerico:
## Summary

<!-- Briefly explain what this PR does. What problem does it solve? -->

zksyncos is using the default alloy poll time. That's 7 seconds. When zksyncos is used in test environments that's too long.
This PR adds the possibility to configure shorter polling times for development environments:

```yaml
# config.yaml
general:
  rocks_db_path: /db/node1                                                                                                                                                                                                                                                           
  l1_provider_poll_interval: 100ms
```

## Related:

Closes #1176

### DIFF
--- a/node/bin/src/config/mod.rs
+++ b/node/bin/src/config/mod.rs
@@ -371,7 +371,7 @@ pub struct GeneralConfig {
     /// Poll interval used by the L1/Gateway alloy provider when waiting for transaction receipts.
     /// Alloy's default is 7 seconds for HTTP transports, using the same criteria here.
     #[config(default_t = 7 * TimeUnit::Seconds)]
-    pub l1_provider_poll_interval: Duration,
+    pub l1_rpc_poll_interval: Duration,
 
     /// Gateway's JSON RPC API.
     /// Must be present if the chain is currently settling to Gateway.
--- a/node/bin/src/config/mod.rs
+++ b/node/bin/src/config/mod.rs
@@ -368,7 +368,7 @@ pub struct GeneralConfig {
     #[config(default_t = "http://localhost:8545".into())]
     pub l1_rpc_url: String,
 
-    /// Poll interval used by the L1/Gateway alloy provider when waiting for transaction receipts.
+    /// Poll interval used by the L1 alloy provider when waiting for transaction receipts.
     /// Alloy's default is 7 seconds for HTTP transports, using the same criteria here.
     #[config(default_t = 7 * TimeUnit::Seconds)]
     pub l1_rpc_poll_interval: Duration,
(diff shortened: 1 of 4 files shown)


node/bin/src/config/mod.rs
```diff
@@ -368,6 +368,11 @@ pub struct GeneralConfig {
     #[config(default_t = "http://localhost:8545".into())]
     pub l1_rpc_url: String,
 
+    /// Poll interval used by the L1/Gateway alloy provider when waiting for transaction receipts.
+    /// Alloy's default is 7 seconds for HTTP transports, using the same criteria here.
+    #[config(default_t = 7 * TimeUnit::Seconds)]
+    pub l1_provider_poll_interval: Duration,
```
slowli: Bikeshedding: I'd call the config param `l1_rpc_poll_interval` to have a (reasonable) common prefix with `l1_rpc_url` above.

node/bin/src/lib.rs
```diff
@@ -189,9 +189,15 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
-    let l1_provider = build_node_provider(&config.general_config.l1_rpc_url).await;
+    let l1_provider = build_node_provider(
+        &config.general_config.l1_rpc_url,
+        config.general_config.l1_provider_poll_interval,
+    )
+    .await;
     let gateway_provider = match &config.general_config.gateway_rpc_url {
-        Some(url) => Some(build_node_provider(url).await),
+        Some(url) => {
+            Some(build_node_provider(url, config.general_config.l1_provider_poll_interval).await)
```
slowli: Is it always the case that the user would want to set both L1 and gateway poll intervals to the same value? I.e., would a separate `gateway_rpc_poll_interval` param not make sense? (If necessary, it could e.g. be optional to copy from the L1 param if not set.)

node/bin/src/config/mod.rs
itegulov: Could you please set this to something smaller for integration tests? Presumably this will speed up some stuff there too. Good catch btw

MexicanAce: Should this be something that's overriden as a config for integration tests? Would this affect production if we change the default values?

If not, do you have a recommendation for what this value should be defaulted to if it's different than what `alloy` has?

itegulov: No-no, I meant in `integration-tests/src/lib.rs` change initialization so the value is overriden with something smaller. 7 seconds default is fine here IMHO

calvogenerico: Done!

node/bin/src/lib.rs
calvogenerico: I went for the easiest path of creating a new configuration for the gateway, independent from l1. Let me know what you think.

node/bin/src/config/mod.rs
calvogenerico: Done! I'm not super familiarized with these tests, but it looks like slightly decreased the time in my env. It went from 11m42s  to 10m45s. I didn't run them more than once, so I'm not 100% convinced that the change was thanks to this config.

itegulov (approved):

PR #1179, opened 2026-04-16
test(integration-tests): disable p2p service for tests by default

itegulov-claude:
## Summary
- stop enabling the p2p network service for integration-test nodes by default
- add explicit builder opt-ins for tests that launch external nodes or use batch verification

### DIFF
--- a/integration-tests/tests/node/external_node.rs
+++ b/integration-tests/tests/node/external_node.rs
@@ -16,7 +16,6 @@ use zksync_os_integration_tests::{
 use zksync_os_server::config::Config;
 
 #[test_multisetup([CURRENT_TO_L1, NEXT_TO_GATEWAY])]
-#[test_builder(|builder| builder.enable_p2p())]
 async fn batch_verification_works(builder: TesterBuilder) -> anyhow::Result<()> {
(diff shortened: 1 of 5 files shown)


integration-tests/tests/node/external_node.rs
```diff
@@ -73,6 +75,7 @@ async fn batch_verification_without_enough_ens(builder: TesterBuilder) -> anyhow
 }
 
 #[test_multisetup([CURRENT_TO_L1])]
+#[test_builder(|builder| builder.enable_p2p())]
 async fn batch_verification_with_2_ens(builder: TesterBuilder) -> anyhow::Result<()> {
```
perekopskiy, line 76: I think it's not needed for `batch_verification_with_2_ens`, `batch_verification_without_enough_ens`, `batch_verification_works` because they do `builder.batch_verification(..)`

itegulov-claude: Removed the redundant `enable_p2p()` annotations from those batch-verification tests. `TesterBuilder::batch_verification()` already enables p2p.

perekopskiy (approved):

PR #1182, opened 2026-04-17
feat(rpc): use log index in eth_getLogs to skip candidate blocks

LogvinovLeon:
## Summary

- Adds a log index lookup before the bloom filter scan in `eth_getLogs`. Blocks within the index's covered range that are not in the candidate bitmap are skipped entirely; blocks outside fall back to bloom filter.
- Extracts index composition logic into `eth_filter/index.rs` with a `Candidates` struct supporting `|` (OR within group) and `&` (AND across groups) operators.
- Adds per-call Prometheus metrics: block disposition by kind (`total`, `skipped_by_index`, `bloom_true_positive`, `bloom_false_positive`, `bloom_negative`, `logs_returned`), plus counters for unconstrained and truncated queries.
- Unit tests for `candidates()` composition logic in `index.rs`.

## Test plan

- [ ] Unit tests pass: `cargo nextest run -p zksync_os_rpc --lib`
- [ ] Clippy/fmt clean: `cargo fmt --all -- --check && cargo clippy --all-targets --all-features --workspace --exclude zksync_os_integration_tests -- -D warnings`
- [ ] Deploy to stage and verify `get_logs_scanned_blocks[skipped_by_index]` metric is non-zero for constrained filters

### DIFF
--- a/lib/rpc/src/eth_filter/index.rs
+++ b/lib/rpc/src/eth_filter/index.rs
@@ -1,0 +1,128 @@
+use std::ops::{BitAnd, BitOr, Range};
+
+use alloy::{
+    primitives::{Address, B256},
+    rpc::types::{Filter, FilterSet},
+};
+use roaring::RoaringBitmap;
+use zksync_os_storage_api::{LogIndex, RepositoryResult};
+
+trait IteratorExt: Iterator {
+    /// Same as nightly `Iterator::try_reduce`, for stable Rust.
+    fn try_reduce<T, E, F>(mut self, mut f: F) -> Result<Option<T>, E>
+    where
+        Self: Sized + Iterator<Item = Result<T, E>>,
+        F: FnMut(T, T) -> T,
+    {
+        let first = match self.next() {
+            None => return Ok(None),
+            Some(Err(e)) => return Err(e),
+            Some(Ok(x)) => x,
+        };
+        self.try_fold(first, |acc, item| item.map(|x| f(acc, x)))
+            .map(Some)
+    }
+}
+
+impl<I: Iterator> IteratorExt for I {}
+
+/// Builds a candidate-block bitmap from the log index for `filter` over `range`.
+///
+/// Returns the candidates for the filter over `range`.
+/// Blocks outside `covered` must be checked via bloom filter.
+/// Returns empty candidates if the filter has no address or topic constraints.
+pub(crate) fn candidates(
+    repo: &dyn LogIndex,
+    filter: &Filter,
+    range: Range<u64>,
+) -> RepositoryResult<Candidates> {
+    // Within a group bitmaps are OR'd (any match); groups are AND'd (all must match).
+    let mut groups: Vec<Candidates> = vec![];
+
+    if !filter.address.is_empty() {
+        groups.push(address_candidates(repo, &filter.address, range.clone())?);
+    }
+    for topics in filter.topics.iter().filter(|ts| !ts.is_empty()) {
+        groups.push(topic_candidates(repo, topics, range.clone())?);
+    }
+
+    Ok(groups.into_iter().reduce(|a, b| a & b).unwrap_or_default())
+}
+
+/// OR's the bitmaps for all addresses in the filter.
+fn address_candidates(
+    repo: &dyn LogIndex,
+    addresses: &FilterSet<Address>,
+    range: Range<u64>,
+) -> RepositoryResult<Candidates> {
+    Ok(addresses
+        .iter()
+        .map(|addr| {
+            repo.blocks_for_address(*addr, range.clone())
+                .map(Candidates::from)
+        })
+        .try_reduce(|a, b| a | b)?
+        .unwrap_or_default())
+}
+
+/// OR's the bitmaps for all topics in a single topic position.
+fn topic_candidates(
+    repo: &dyn LogIndex,
+    topics: &FilterSet<B256>,
+    range: Range<u64>,
+) -> RepositoryResult<Candidates> {
+    Ok(topics
+        .iter()
+        .map(|topic| {
+            repo.blocks_for_topic(*topic, range.clone())
+                .map(Candidates::from)
+        })
+        .try_reduce(|a, b| a | b)?
+        .unwrap_or_default())
+}
+
+/// A set of candidate blocks from the log index, together with the range of blocks the index covers.
+/// Blocks outside `covered` must be checked via bloom filter regardless of `bitmap`.
+pub struct Candidates {
+    pub bitmap: RoaringBitmap,
+    pub covered: Range<u64>,
+}
+
+impl BitOr for Candidates {
+    type Output = Self;
+    fn bitor(self, other: Self) -> Self {
+        Self {
+            bitmap: self.bitmap | other.bitmap,
+            covered: intersect(self.covered, other.covered),
+        }
+    }
+}
+
+impl BitAnd for Candidates {
+    type Output = Self;
+    fn bitand(self, other: Self) -> Self {
+        Self {
+            bitmap: self.bitmap & other.bitmap,
+            covered: intersect(self.covered, other.covered),
+        }
+    }
+}
+
+impl Default for Candidates {
+    fn default() -> Self {
+        Self {
+            bitmap: RoaringBitmap::new(),
+            covered: 0..0,
+        }
+    }
+}
+
+impl From<(RoaringBitmap, Range<u64>)> for Candidates {
+    fn from((bitmap, covered): (RoaringBitmap, Range<u64>)) -> Self {
+        Self { bitmap, covered }
+    }
+}
+
+fn intersect(a: Range<u64>, b: Range<u64>) -> Range<u64> {
+    a.start.max(b.start)..a.end.min(b.end)
+}
--- a/lib/rpc/src/eth_filter/index.rs
+++ b/lib/rpc/src/eth_filter/index.rs
@@ -126,3 +126,130 @@ impl From<(RoaringBitmap, Range<u64>)> for Candidates {
 fn intersect(a: Range<u64>, b: Range<u64>) -> Range<u64> {
     a.start.max(b.start)..a.end.min(b.end)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloy::primitives::{Address, B256};
+    use std::collections::HashMap;
+    use zksync_os_storage_api::RepositoryResult;
+
+    #[derive(Debug, Default)]
+    struct MockIndex {
+        addresses: HashMap<Address, (RoaringBitmap, Range<u64>)>,
+        topics: HashMap<B256, (RoaringBitmap, Range<u64>)>,
+    }
+
+    impl MockIndex {
+        fn with_address(mut self, addr: Address, blocks: &[u64], covered: Range<u64>) -> Self {
+            self.addresses
+                .insert(addr, (blocks.iter().map(|&b| b as u32).collect(), covered));
+            self
+        }
+
+        fn with_topic(mut self, topic: B256, blocks: &[u64], covered: Range<u64>) -> Self {
+            self.topics
+                .insert(topic, (blocks.iter().map(|&b| b as u32).collect(), covered));
+            self
+        }
+    }
+
+    impl LogIndex for MockIndex {
+        fn blocks_for_address(
+            &self,
+            address: Address,
+            _range: Range<u64>,
+        ) -> RepositoryResult<(RoaringBitmap, Range<u64>)> {
+            Ok(self.addresses.get(&address).cloned().unwrap_or_default())
+        }
+
+        fn blocks_for_topic(
+            &self,
+            topic: B256,
+            _range: Range<u64>,
+        ) -> RepositoryResult<(RoaringBitmap, Range<u64>)> {
+            Ok(self.topics.get(&topic).cloned().unwrap_or_default())
+        }
+    }
+
+    fn blocks(c: &Candidates) -> Vec<u32> {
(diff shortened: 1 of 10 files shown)


lib/storage/src/log_index_filter.rs
```diff
@@ -0,0 +1,255 @@
...
+) -> RepositoryResult<Candidates> {
+    // Within a group bitmaps are OR'd (any match); groups are AND'd (all must match).
+    let mut groups: Vec<Candidates> = vec![];
+
+    if !filter.address.is_empty() {
+        groups.push(address_candidates(repo, &filter.address, range.clone())?);
+    }
+    for topics in filter.topics.iter().filter(|ts| !ts.is_empty()) {
+        groups.push(topic_candidates(repo, topics, range.clone())?);
+    }
```
slowli, line 47: I wonder what kind of gains would be possible by parallelizing this logic (e.g., by using `rayon`). Ultimately, it would probably be I/O-bottlenecked, but w/o parallelization it *could* be the case that this bottleneck isn't hit.

lib/rpc/src/eth_filter/index.rs
```diff
@@ -0,0 +1,255 @@
...
+
+impl<I: Iterator> IteratorExt for I {}
+
+/// Builds a candidate-block bitmap from the log index for `filter` over `range`.
+///
+/// Returns the candidates for the filter over `range`.
+/// Blocks outside `covered` must be checked via bloom filter.
+/// Returns empty candidates if the filter has no address or topic constraints.
+pub(crate) fn candidates(
+    repo: &dyn LogIndex,
```
slowli: Bikeshedding: IMO, it'd make more sense to structure this function and dependencies as an inherent method of `dyn LogIndex` (moving them to the corresponding crate). To me, it looks like they have a clear receiver, and the containing `storage_api` crate depends on `alloy` RPC types anyway.

lib/rpc/src/eth_filter/scan.rs
```diff
@@ -1,27 +1,39 @@
...
     let mut logs = Vec::new();
 
     for number in from..=to {
+        // Within the covered range, skip blocks the index says have no matching logs.
+        if candidates.covered.contains(&number) && !candidates.bitmap.contains(number as u32) {
+            stats.skipped_by_index += 1;
+            continue;
+        }
+
         let Some(block) = repo.get_block_by_number(number)? else {
```
slowli, line 33: This may be off-topic, but without introducing new indexes, a low-hanging optimization opportunity would be batching I/O by requesting block headers with a set of numbers. This *could* result in unnecessary headers returned (if `max_logs` is set for a multi-block request), but since block headers are reasonably small, it could be balanced.

lib/rpc/src/eth_filter/scan.rs
```diff
@@ -98,27 +112,45 @@ fn collect_matching_logs(filter: &Filter, stored_txs: Vec<StoredTxData>, out: &m
...
     }
 }
 
 impl Drop for BlockScanStats {
     fn drop(&mut self) {
         API_METRICS.get_logs_scanned_blocks[&"total"].observe(self.total);
-        API_METRICS.get_logs_scanned_blocks[&"true_positive"].observe(self.true_positive);
-        API_METRICS.get_logs_scanned_blocks[&"false_positive"].observe(self.false_positive);
-        API_METRICS.get_logs_scanned_blocks[&"negative"].observe(self.negative);
+        API_METRICS.get_logs_scanned_blocks[&"skipped_by_index"].observe(self.skipped_by_index);
```
slowli: - Nit: It would be more typesafe to make the label an enum with derived `EncodeLabelValue` / `EncodeValueSet`.
- IMO, it would make sense to split metrics by unconstrained / constrained category (maybe, with greater granularity, e.g., "unconstrained" / "has address" / "has address + 1st topic" / "everything else") by adding an additional label for the `get_logs_scanned_blocks` family (i.e., have both kind and filter category as 2 distinct labels).

lib/rpc/src/eth_filter/scan.rs
```diff
@@ -98,27 +112,45 @@ fn collect_matching_logs(filter: &Filter, stored_txs: Vec<StoredTxData>, out: &m
 /// Observes Prometheus metrics when dropped, ensuring they are recorded on all exit paths.
 struct BlockScanStats {
     total: u64,
-    true_positive: u64,
-    false_positive: u64,
-    negative: u64,
+    skipped_by_index: u64,
+    bloom_true_positive: u64,
+    bloom_false_positive: u64,
+    bloom_negative: u64,
```
slowli, line 115: It may make sense to also record ratios of these values vs the number of scanned blocks; it looks more easily interpretable (AFAIU, it's impossible to get precise ratios from recorded bucket values).

lib/storage/src/log_index_filter.rs
```diff
@@ -0,0 +1,255 @@
...
+        .try_reduce(|a, b| a | b)?
+        .unwrap_or_default())
+}
+
+/// A set of candidate blocks from the log index, together with the range of blocks the index covers.
+/// Blocks outside `covered` must be checked via bloom filter regardless of `bitmap`.
+pub struct Candidates {
+    pub bitmap: RoaringBitmap,
+    pub covered: Range<u64>,
+}
```
slowli, line 102: Nit: The struct may have `pub(crate)` visibility. Its fields may be made private by introducing a getter like this:

```rust
impl Candidates {
    pub(crate) fn may_contain(&self, block_number: u64) -> bool {
        !self.covered.contains(&block_number) || self.bitmap.contains(block_number as u32)
    }
}
```

lib/storage/src/log_index_filter.rs
LogvinovLeon: The way I anticipate - typical queries will look like after talking with clients is one of those two:
* Get all events from contract X (address only). (more frequent one)
* Get all events from contract X with topic0 = Y (address + topic0)

Therefore - I don't expect anyone to use multiple topic filters. It's just implemented to be compliant with the spec.

Happy to re-address later if we have other clients with other query patterns but at the current stage - don't think it's worth the complexity here.

lib/rpc/src/eth_filter/index.rs
LogvinovLeon: Moved to `zksync_os_storage::log_index_filter` (not `storage_api`, since this is logic rather than a trait contract). `index.rs` in the rpc crate is gone entirely — `scan.rs` imports directly from the storage crate.

lib/rpc/src/eth_filter/scan.rs
LogvinovLeon: Agreed the batching would help, but this PR already does something strictly better: instead of N block header reads for an N-block range, we do a single bitmap lookup (one disk seek) and only touch headers for the blocks the index says are candidates. For a constrained filter over 10k blocks with a small hit rate, that's ~1 seek instead of ~10k, not ~1 batched seek instead of ~10k sequential ones. Batching would still be a win for unconstrained queries, but that's probably a separate PR.

lib/storage/src/log_index_filter.rs
LogvinovLeon: Done — fields are private, `may_contain()` is public. The struct itself stays `pub` since it's now in `zksync_os_storage` and used from `zksync_os_rpc`, but the encapsulation is the same.

lib/rpc/src/eth_filter/scan.rs
LogvinovLeon: Done — added `FilterCategory` enum (`unconstrained` / `address_only` / `topic_only` / `address_and_topic`) as a second label on `get_logs_scanned_blocks`. The enum also derives `From<&Filter>` so the categorisation logic is in one place. `get_logs_unconstrained` counter is superseded by `{filter="unconstrained", kind="total"}`.

lib/rpc/src/eth_filter/scan.rs
LogvinovLeon: Added three per-call ratio histograms broken down by filter category: `get_logs_index_skip_ratio` (skipped / total), `get_logs_index_coverage` (index-covered range / total range), and `get_logs_bloom_fp_rate` (false positives / bloom-checked blocks). Buckets are non-uniform with finer resolution near 0.0 and 1.0 where the interesting edge cases live. Aggregate ratios are also queryable via `_sum` on the existing histograms.

lib/rpc/src/eth_filter/scan.rs
slowli: I'm inclined to disagree that the current approach is *strictly* better; it's possible to first filter blocks by the log index(es) and then batch-request block headers. The hard part is accounting for `max_logs` limiter, but as I've said previously, there is probably a balance point between over-requesting headers and optimizing I/O. However, this is outside the scope of this PR, so may get implemented later (or never) 🤷

lib/rpc/src/metrics.rs
```diff
@@ -1,14 +1,67 @@
...
 #[derive(Debug, Metrics)]
 pub struct ApiMetrics {
-    #[metrics(labels = ["method"], buckets = BLOCK_COUNTS)]
-    pub get_logs_scanned_blocks: LabeledFamily<&'static str, Histogram<u64>>,
+    /// Block disposition per `eth_getLogs` call, broken down by outcome kind and filter category.
+    #[metrics(labels = ["kind", "filter"], buckets = BLOCK_COUNTS)]
+    pub get_logs_scanned_blocks: LabeledFamily<(GetLogsStat, FilterCategory), Histogram<u64>, 2>,
+    /// Per-call fraction of blocks skipped by the log index (skipped / total).
+    #[metrics(labels = ["filter"], buckets = RATIO_BUCKETS)]
+    pub get_logs_index_skip_ratio: LabeledFamily<FilterCategory, Histogram<f64>>,
```
slowli: Nit: You can use `#[derive(EncodeLabelSet)]` on `FilterCategory` (in fact, `#[metrics(label = _)]` is useless w/o such a derive), and use a `Family` here.

lib/rpc/src/metrics.rs
```diff
@@ -1,14 +1,67 @@
...
+    /// Per-call bloom filter false-positive rate among blocks that reached the bloom check.
+    #[metrics(labels = ["filter"], buckets = RATIO_BUCKETS)]
+    pub get_logs_bloom_fp_rate: LabeledFamily<FilterCategory, Histogram<f64>>,
```
slowli: Micro-nit: in machine learning, [there is](https://en.wikipedia.org/wiki/Precision_and_recall) a similar *precision* ratio defined as `TP / (TP + FP)` (i.e., 1 - this metric). IMO, it'd be more logical to use it.

lib/rpc/src/metrics.rs
```diff
@@ -1,14 +1,67 @@
...
+    /// Per-call fraction of the queried block range covered by the log index.
+    #[metrics(labels = ["filter"], buckets = RATIO_BUCKETS)]
+    pub get_logs_index_coverage: LabeledFamily<FilterCategory, Histogram<f64>>,
+    /// Number of `eth_getLogs` calls truncated due to exceeding `max_logs`.
+    pub get_logs_truncated: Counter,
```
slowli: Does it not make sense to split this counter by `FilterCategory`?

slowli (commented):
Looks good, just got a couple of nits and a question regarding metrics definition.

lib/rpc/src/metrics.rs
LogvinovLeon: Done — renamed to `get_logs_bloom_precision` and flipped the formula to `TP / (TP + FP)` in 0b5205e.

lib/rpc/src/metrics.rs
LogvinovLeon: Done in e973720f — added `EncodeLabelSet` to `FilterCategory` (the `label = "filter"` attribute is now meaningful), switched the three single-label ratio fields from `LabeledFamily` to `Family`, and dropped the redundant `labels = ["filter"]` from each field site.

lib/rpc/src/metrics.rs
LogvinovLeon: Done in 7cb9dd0f — changed to `Family<FilterCategory, Counter>`.

lib/rpc/src/eth_filter/scan.rs
LogvinovLeon: Agree. I would leave it for the future though

lib/rpc/src/metrics.rs
```diff
@@ -46,22 +49,32 @@ impl From<&Filter> for FilterCategory {
     }
 }
 
+/// Metrics for `eth_getLogs` scan behaviour.
 #[derive(Debug, Metrics)]
-pub struct ApiMetrics {
-    /// Block disposition per `eth_getLogs` call, broken down by outcome kind and filter category.
+#[metrics(prefix = "get_logs")]
```
slowli, line 54: Bikeshedding: I never noticed that the original metrics don't have a common prefix defined. Subjectively, metrics should strart with a component prefix (i.e., something like `rpc` or `api` or `web3` in this case; `api` looks best given that it's used by the common metrics below), otherwise they quickly become hard to maintain.

slowli (approved):

lib/rpc/src/metrics.rs
LogvinovLeon: I've added the api_ prefix already. For that exact reason. `cancellation` is not unique enough in the context of the whole server. api_cancellation is.

This will unfortunately break all the dashboards that rely on it and all allerts.

I'm the de-facto DRI for this section of dashboards so I'll fix them when this gets deployed.
As for alerts - I'll grep the gitops repo but also let devops know cause maybe they have a broader context

PR #1185, opened 2026-04-20
feat: detect SL intervals using L1

itegulov:
## Summary

Adds logic that detects all SL intervals where chain continuously settled on either L1 or GW.

This logic will be used by all watchers in the future to detect which SL to watch (or even both sequentially if we need to load batches from two adjacent intervals).


lib/contract_interface/src/settlement_layer_intervals.rs
```diff
@@ -0,0 +1,221 @@
...
+/// The intervals are reconstructed from `IL1ChainAssetHandler.migrationInterval(chainId, i)`
+/// for each known migration slot (`i ∈ [1, migrationNumber(chainId)]`):
+///
+/// - Each populated slot describes one L1 → Gateway → L1 round-trip, giving us the chain's own
+///   batch number at which the migration to the Gateway happened (`migrateToGWBatchNumber`) and
+///   the one at which it returned (`migrateFromGWBatchNumber`, or `isActive = true` if the
+///   chain has not returned yet).
+/// - Slot `0` is reserved for the legacy Gateway and is intentionally skipped — legacy-GW chains
+///   are not supported here.
+/// - `MAX_ALLOWED_NUMBER_OF_MIGRATIONS = 2` on-chain, so at most two cycles are supported.
```
perekopskiy, line 110: Nice comment! It takes a while to understand it but after you understand it perfectly makes sense 😄

perekopskiy (approved):

PR #1188, opened 2026-04-21
fix(rpc): fix monitoring middleware sync handler measurement and panic handling

LogvinovLeon:
## Summary
- **Sync handlers not measured**: \`inner.call(request)\` was invoked before the \`async move\` block, so sync handlers (dispatched eagerly by jsonrpsee) ran outside \`CallGuard\`'s timer. Fixed by cloning \`inner\` and moving the call to after \`CallGuard::new()\` inside the async block.
- **Panics drop the connection**: for HTTP, a handler panic propagates through hyper's connection task, dropping the TCP connection with no response. Fixed by wrapping the inner call as \`AssertUnwindSafe(async move { inner.call(request).await }).catch_unwind()\`. The extra \`async move\` is required because sync handler panics fire during \`inner.call()\` (before the future is polled), so a plain \`AssertUnwindSafe(inner.call(request)).catch_unwind()\` would not catch them. Panics now return a \`-32603 Internal error\` response. A new \`panicked\` counter (per method) is recorded through \`CallGuard\`, consistent with how \`cancelled\` works. Panics are no longer misrecorded as cancellations.
- **\`notification()\` had no guard**: added a \`kind\` field (\`Call\`/\`Notification\`) to \`CallGuard\` and unified \`notification()\` with the same guard-based instrumentation as \`call()\`, giving notifications proper cancellation and panic coverage.

### DIFF
--- a/lib/rpc/src/monitoring_middleware.rs
+++ b/lib/rpc/src/monitoring_middleware.rs
@@ -1,12 +1,15 @@
 use crate::metrics::API_METRICS;
+use crate::result::internal_rpc_err;
+use futures::FutureExt as _;
 use jsonrpsee::core::middleware::{Batch, BatchEntry, Notification};
 use jsonrpsee::server::middleware::rpc::{RpcService, RpcServiceT};
 use jsonrpsee::types::Request;
 use jsonrpsee::{BatchResponseBuilder, MethodResponse};
 use std::collections::HashMap;
+use std::panic::AssertUnwindSafe;
 use std::time::{Duration, Instant};
 
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug)]
 pub enum CallKind {
     Call,
     Notification,
@@ -29,20 +32,24 @@ impl Monitoring {
 
 /// Ensures latency is recorded even if the future is dropped mid-flight (client disconnected).
 struct CallGuard {
+    kind: CallKind,
     method: String,
     started: Instant,
     request_size: usize,
     /// `Some((output_size, error_code))` once the future has resolved.
     completed: Option<(usize, Option<i32>)>,
+    panicked: bool,
 }
 
 impl CallGuard {
-    fn new(method: String, request_size: usize) -> Self {
+    fn new(kind: CallKind, method: String, request_size: usize) -> Self {
         Self {
+            kind,
             method,
             started: Instant::now(),
             request_size,
             completed: None,
+            panicked: false,
         }
     }
 }
@@ -97,8 +104,11 @@ impl Drop for CallGuard {
         if cancelled {
             API_METRICS.cancelled[&self.method].inc();
         }
+        if self.panicked {
+            API_METRICS.panicked[&self.method].inc();
+        }
         log_and_report(
-            CallKind::Call,
+            self.kind,
             &self.method,
             elapsed,
             self.request_size,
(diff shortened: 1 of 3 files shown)


lib/rpc/src/monitoring_middleware.rs
```diff
@@ -120,11 +130,19 @@ impl RpcServiceT for Monitoring {
...
 
         async move {
-            let mut guard = CallGuard::new(method, request_size);
-            let out = fut.await;
+            let id = request.id.clone().into_owned();
+            let mut guard = CallGuard::new(CallKind::Call, method.clone(), request_size);
+            let result = AssertUnwindSafe(async move { inner.call(request).await })
+                .catch_unwind()
+                .await;
+            guard.panicked = result.is_err();
```
slowli: AFAIU, you could achieve a similar result by checking [`thread::panicking()`](https://doc.rust-lang.org/std/thread/fn.panicking.html) in `impl Drop for CallGuard`, although it wouldn't influence how panicking RPC handlers are translated into responses (which is obviously important, so it probably should stay as-is).

lib/rpc/src/monitoring_middleware.rs
```diff
@@ -120,11 +130,19 @@ impl RpcServiceT for Monitoring {
...
+            let out = result.unwrap_or_else(|_| {
+                tracing::error!(method = %method, "RPC handler panicked");
```
slowli: Nit: This logging can be moved to `CallGuard` so that `method` doesn't need to be cloned additionally.

lib/rpc/src/monitoring_middleware.rs
```diff
@@ -97,8 +104,11 @@ impl Drop for CallGuard {
         if cancelled {
             API_METRICS.cancelled[&self.method].inc();
         }
+        if self.panicked {
+            API_METRICS.panicked[&self.method].inc();
+        }
         log_and_report(
```
slowli: Nit: `log_and_report` is now only used here and can be inlined.

lib/rpc/src/monitoring_middleware.rs
```diff
@@ -203,22 +221,19 @@ impl RpcServiceT for Monitoring {
...
-                CallKind::Notification,
-                &method,
-                started.elapsed(),
-                request_size,
-                output_size,
-                out.as_error_code(),
-                false,
-            );
+            let mut guard = CallGuard::new(CallKind::Notification, method.clone(), request_size);
+            let result = AssertUnwindSafe(async move { inner.notification(n).await })
```
slowli: Nit: I wonder whether it would make sense to extract this logic to a method in `CallGuard` so it's not repeated for calls and notifications:

```rust
let guard = CallGuard::new(CallKind::Notification, method.clone(), request_size);
guard
    .handle_result(
        async move { inner.notification(n).await },
        MethodResponse::notification, // erroneous response generator
    )
    .await
```

lib/rpc/src/monitoring_middleware.rs
LogvinovLeon: Done — moved into `CallGuard::Drop`, which also removed the need to clone `method`.

lib/rpc/src/monitoring_middleware.rs
LogvinovLeon: Done — inlined and removed the free function.

lib/rpc/src/monitoring_middleware.rs
LogvinovLeon: Done — extracted as `CallGuard::handle_result`. Agreed it's cleaner: the guard now fully owns its lifecycle, and callers just provide the handler future and a panic fallback.

lib/rpc/src/monitoring_middleware.rs
LogvinovLeon: Good point — thread::panicking() would work for the metric, but since we need catch_unwind anyway to convert panics into error responses, the panicked flag falls out naturally from result.is_err() in handle_result. No extra mechanism needed.

slowli (approved):

PR #1189, opened 2026-04-21
feat(l1-sender): increase L1 required confirmations

Artemka374:
## What                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                      
  - Raise required transaction confirmations from 1 to 3 for direct L1 submissions. Gateway submissions stay at 1, since low block production there makes 3 confirmations impractical. Named constants document the rationale.
  - Emit metrics for alternative fee estimation strategies - mostly for observability to make a decision in case other fixes won't be enough                                                                                                                                                           
  - Set `block_time(1)` on the Anvil L1 instance in integration tests so the 3-confirmation wait doesn't stall.

### DIFF
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -1221,6 +1221,7 @@ impl AnvilL1 {
                 .port(locked_port.port)
                 .chain_id(L1_CHAIN_ID)
                 .block_time(1)
+                .arg("--mixed-mining")
                 .arg("--load-state")
                 .arg(l1_state_path)
         })?;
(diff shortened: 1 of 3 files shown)


integration-tests/src/lib.rs
```diff
@@ -1220,6 +1220,7 @@ impl AnvilL1 {
             anvil
                 .port(locked_port.port)
                 .chain_id(L1_CHAIN_ID)
+                .block_time(1)
```
itegulov, line 1223: I believe applying this to all tests makes some of them take 1-2s slower. AFAIR there is a way to configure anvil so that it uses a combination of immediate block sealing AND block time based (i.e. seals at least one block every 1s but also immediately mines a block if there is a pending tx in mempool)

itegulov (approved):

PR #1192, opened 2026-04-21
test(integration-tests): add HTTP RPC timeline recorder for lifecycle tests

romanbrodetski-ai:
## Summary
- Adds `HttpRpcRecorder` that polls an HTTP RPC endpoint during lifecycle events (restart, rebuild) and records state transitions with timing
- Used in restart and rebuild tests to assert the node becomes ready and to log detailed RPC availability timelines
- Makes `reqwest` a non-optional dependency (was previously gated behind `prover-tests`)

## Output example

```
  Observed HTTP RPC detailed timeline during restart
    0.093s | rpc became available; latest block=2 hash=0x6cd47
           | 2 requests over 101ms; avg latency=1ms min=0ms max=1ms; all successful with unchanged latest block=2 hash=0x6cd47
  0.294s (after 0.201s) | rpc started failing after latest block=2 hash=0x6cd47; transport error: error sending request for url (http://localhost:37281/)
           | 41 requests over 4100ms; avg latency=0ms min=0ms max=1ms; 41x same failure: transport error: error sending request for url (http://localhost:37281/)
  4.495s (after 4.200s) | rpc recovered; latest block=3 hash=0x42a86
           | 1 requests over 975ms; avg latency=1ms min=1ms max=1ms; all successful with unchanged latest block=3 hash=0x42a86

```

### DIFF
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -4,6 +4,7 @@ use crate::network::Zksync;
 use crate::node_log::NodeLogState;
 use crate::prover_tester::ProverTester;
 use crate::provider::{ZksyncApi, ZksyncTestingProvider};
+use crate::rpc_recorder::{HttpRpcRecorder, RpcRecordConfig};
 use crate::test_config::{build_node_config, disable_prover_input_generation};
 use crate::utils::LockedPort;
 use alloy::network::EthereumWallet;
@@ -48,6 +49,7 @@ mod network;
 mod node_log;
 mod prover_tester;
 pub mod provider;
+pub mod rpc_recorder;
 pub mod test_config;
 pub mod upgrade;
 mod utils;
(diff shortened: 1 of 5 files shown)


integration-tests/src/lib.rs
```diff
@@ -389,6 +391,14 @@ impl Tester {
         &self.l2_rpc_address
     }
 
+    pub fn l2_http_rpc_url(&self) -> &str {
+        &self.l2_rpc_address
+    }
```
itegulov: Isn't this the same as `l2_rpc_url` above?

integration-tests/src/rpc_recorder.rs
```diff
@@ -0,0 +1,881 @@
...
+                "params": [],
+            },
+            {
+                "jsonrpc": "2.0",
+                "id": 3,
+                "method": "eth_getBlockByNumber",
+                "params": ["latest", false],
+            }
+        ]))
+        .send()
```
itegulov, line 742: This seems pretty low-level with how you manual construct requests and then manually decode batch responses. Is there some reason for it?

RomanBrodetski: added comment

integration-tests/src/lib.rs
RomanBrodetski: yep thx

PR #1199, opened 2026-04-22
fix(rpc): correct error codes and add detailed rejection metrics for eth_sendRawTransaction

LogvinovLeon:
## Summary

- **Fix error code classification**: `PoolError` was incorrectly mapped to `-32603` (server error) — pool rejections (duplicate, underpriced, invalid nonce) are client errors and now return `-32602`. `ForwardError` preserves the main node's original error code instead of always returning `-32603`. `EthSendRawTransactionSyncError::Regular` now delegates to the inner error's mapping instead of flattening everything to `-32603`.
- **Add `tx_submission_rejections{reason=...}` metric**: a `Family<TxRejectionReason, Counter>` with 25 distinct reason values covering every error path — decode failures, invalid signatures, blacklisted signers, forwarding errors (transport vs rejected), and every `PoolErrorKind`/`InvalidPoolTransactionError` variant individually. This replaces the opaque `-32602`/`-32603` split in Grafana with actionable per-reason visibility.

## Test plan

- [ ] `cargo check -p zksync_os_rpc` passes
- [ ] Verify `tx_submission_rejections` metric appears in Prometheus with correct `reason` labels
- [ ] Submit invalid tx (bad signature) → expect `-32602` and `reason="invalid_signature"` counter increment
- [ ] Submit tx via external node with main node down → expect `-32603` and `reason="forward_transport_error"`
- [ ] Submit duplicate tx → expect `-32602` and `reason="pool_already_imported"`

### DIFF
--- a/lib/rpc/src/metrics.rs
+++ b/lib/rpc/src/metrics.rs
@@ -1,5 +1,8 @@
 use std::time::Duration;
-use vise::{Buckets, Counter, Histogram, LabeledFamily, Metrics, Unit};
+use vise::{
+    Buckets, Counter, EncodeLabelSet, EncodeLabelValue, Family, Histogram, LabeledFamily, Metrics,
+    Unit,
+};
 
 const LATENCIES_FAST: Buckets = Buckets::exponential(0.000001..=32.0, 2.0);
 const BLOCK_COUNTS: Buckets = Buckets::exponential(1.0..=100000.0, 10.0);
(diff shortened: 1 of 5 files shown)


lib/rpc/src/metrics.rs
```diff
@@ -36,7 +39,67 @@ pub struct TxSubmissionMetrics {
...
+    /// Transaction priority fee is below the minimum.
+    PoolPriorityFeeBelowMinimum,
+    /// Consensus-level invalid transaction.
+    PoolConsensusError,
+    /// EIP-4844 blob transaction error.
+    PoolEip4844Error,
+    /// EIP-7702 transaction error.
+    PoolEip7702Error,
+    /// Other pool error.
+    PoolOther,
```
RomanBrodetski, line 104: why flatten the list? Seems like we have ~3 sources of errors - `EthSendRawTransactionError`/`InvalidPoolTransactionError`/`PoolErrorKiund` - maybe just make a wrapper for them?

LogvinovLeon: I think the label values need to be flat strings either way — `EncodeLabelValue` requires mapping to a string, so a wrapper type would still need the same match arms, just consolidated in one place instead of three `From` impls. The flat enum has the upside of making the full set of possible label values visible at the type level.

That said, happy to be corrected if you had something else in mind — what did you have in mind for the wrapper?

RomanBrodetski: Huh I see.. Good arguments - I agree with you.

RomanBrodetski (approved):

PR #1200, opened 2026-04-22
feat(rpc): add txpool_inspect/content/status RPC methods

LogvinovLeon:
## Summary

- Adds the standard geth `txpool` RPC namespace: `txpool_inspect`, `txpool_content`, `txpool_status`
- `txpool_inspect` returns a human-readable summary (to/value/gas/gasPrice) of pending and queued txs grouped by sender and nonce
- `txpool_content` returns full transaction objects for pending and queued txs
- `txpool_status` returns pending and queued tx counts

## Motivation

During a GenLayer testnet incident, transactions were accepted by the mempool but never mined. The root cause was a nonce gap: 1625 queued txs on the sequencer with 0 pending. Without `txpool_*` methods, diagnosing this required direct access to internal metrics. These endpoints make mempool state inspectable via standard JSON-RPC, the same way operators do it on geth/reth nodes.

## Implementation notes

- New `TxpoolApi` trait in `lib/rpc_api/src/txpool.rs` using `jsonrpsee` proc macros
- `TxpoolNamespace` in `lib/rpc/src/txpool_impl.rs` backed by `L2Subpool::all_transactions()`
- `TxpoolInspectSummary` uses `pool_tx.transaction.transaction.clone().into_inner().into()` — same pattern as reth's own txpool impl
- `ValidPoolTransaction` re-exported from `zksync_os_mempool` so `txpool_impl` doesn't need a direct `reth_transaction_pool` dep

## Test plan

- [ ] `curl -X POST localhost:3050 -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","method":"txpool_status","params":[],"id":1}'`
- [ ] Submit a tx with a nonce gap and verify it appears in `txpool_inspect` under `queued`
- [ ] Verify `txpool_content` returns full tx fields (from, nonce, gas, etc.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -684,6 +684,7 @@ dependencies = [
  "alloy-rpc-types-debug",
  "alloy-rpc-types-eth",
  "alloy-rpc-types-trace",
+ "alloy-rpc-types-txpool",
  "alloy-signer",
  "alloy-sol-types",
  "alloy-transport",
@@ -790,6 +791,7 @@ dependencies = [
  "alloy-rpc-types-engine",
  "alloy-rpc-types-eth",
  "alloy-rpc-types-trace",
+ "alloy-rpc-types-txpool",
  "alloy-serde",
  "serde",
 ]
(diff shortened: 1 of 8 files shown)


RomanBrodetski:
On the high level I like this - need @itegulov 's approval though :)

LogvinovLeon:
CI failure is because of this: https://github.com/matter-labs/zksync-os-server/pull/1201

itegulov (approved):

PR #1204, opened 2026-04-22
feat(network): add raft transport layer over devp2p

romanbrodetski-ai:
## Summary

Next PR in consensus series. Introduces the changes to networking - that will later be used by raft engine.

- Adds `openraft` dependency
- Adds `lib/consensus_types`: shared openraft type config (`RaftTypeConfig`) and `PeerId`-based node identity
- Adds `lib/network/src/raft/`: `RaftProtocolHandler` (a new RLPx sub-protocol), wire encoding for AppendEntries/Vote/InstallSnapshot RPCs, and `RaftRouter` for routing outbound RPCs and matching responses
- `NetworkService::new` gains an optional `raft_handler` parameter; registers the raft sub-protocol when provided
- Tightens peer backoff durations (30–180 s → 1–10 s) so consensus nodes reconnect quickly after restarts

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "lib/contract_interface",
+    "lib/consensus_types",
     "lib/crypto",
     "lib/genesis",
     "lib/l1_sender",
@@ -194,6 +195,7 @@ num_cpus = "1.17.0"
 rocksdb = "0.24.0"
 thread_local = "1.1.9"
 pin-project = "1.1.10"
+openraft = { version = "0.9.21", features = ["serde", "generic-snapshot-data", "storage-v2"] }
 assert_matches = "1.5"
 serde_yaml = "0.9"
 secrecy = "0.10.3"
@@ -250,6 +252,7 @@ zksync-os-revm = { git = "https://github.com/matter-labs/zksync-os-revm", tag = 
 
 # "Local" dependencies
 zksync_os_contract_interface = { version = "=0.19.0", path = "lib/contract_interface" }
+zksync_os_consensus_types = { version = "=0.19.0", path = "lib/consensus_types" }
 zksync_os_crypto = { version = "=0.19.0", path = "lib/crypto" }
 zksync_os_l1_sender = { version = "=0.19.0", path = "lib/l1_sender" }
 zksync_os_l1_watcher = { version = "=0.19.0", path = "lib/l1_watcher" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,7 +195,7 @@ num_cpus = "1.17.0"
 rocksdb = "0.24.0"
 thread_local = "1.1.9"
 pin-project = "1.1.10"
-openraft = { version = "0.9.21", features = ["serde", "generic-snapshot-data", "storage-v2"] }
+openraft = { version = "0.9.24", features = ["serde", "generic-snapshot-data", "storage-v2"] }
 assert_matches = "1.5"
 serde_yaml = "0.9"
 secrecy = "0.10.3"
--- a/lib/consensus_types/src/lib.rs
+++ b/lib/consensus_types/src/lib.rs
@@ -1,0 +1,29 @@
+use openraft::{BasicNode, Entry, EntryPayload};
+use reth_network_peers::PeerId;
+use std::io::Cursor;
+use zksync_os_storage_api::ReplayRecord;
+
+pub type RaftNode = BasicNode;
+
+openraft::declare_raft_types!(
+    pub RaftTypeConfig:
+        /// Application data carried in each normal log entry — a block to be replayed.
+        D = ReplayRecord,
+        /// Response returned from `client_write`; unused, we use the canonized-blocks channel instead.
+        R = (),
+        NodeId = PeerId,
+        Node = RaftNode,
+);
+
+/// Formats a Raft log entry as a short human-readable string for debug logging.
+pub fn debug_display_raft_entry(entry: &Entry<RaftTypeConfig>) -> String {
+    let payload = match &entry.payload {
+        EntryPayload::Blank => "blank".to_string(),
+        EntryPayload::Normal(r) => format!(
+            "block number {} (block output hash: {})",
+            r.block_context.block_number, r.block_output_hash
+        ),
+        EntryPayload::Membership(_) => "membership".to_string(),
+    };
+    format!("Entry(log_id_index={}, {})", entry.log_id.index, payload)
+}
--- a/lib/network/Cargo.toml
+++ b/lib/network/Cargo.toml
@@ -28,7 +28,6 @@ alloy = { workspace = true, default-features = false, features = [
 ] }
 alloy-rlp.workspace = true
 backon.workspace = true
-bincode.workspace = true
 dashmap.workspace = true
 openraft.workspace = true
 serde = { workspace = true, features = ["derive"] }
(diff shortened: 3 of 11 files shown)


lib/consensus_types/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+        /// Application data carried in each normal log entry — a block to be replayed.
+        D = ReplayRecord,
+        /// Response returned from `client_write`; unused, we use the canonized-blocks channel instead.
+        R = (),
+        NodeId = PeerId,
+        Node = RaftNode,
+);
+
+/// Formats a Raft log entry as a short human-readable string for debug logging.
+pub fn debug_display_raft_entry(entry: &Entry<RaftTypeConfig>) -> String {
```
itegulov, line 19: I don't see this being used anywhere but presumably it will be in a follow-up PR?

Cargo.toml
```diff
@@ -194,6 +195,7 @@ num_cpus = "1.17.0"
 rocksdb = "0.24.0"
 thread_local = "1.1.9"
 pin-project = "1.1.10"
+openraft = { version = "0.9.21", features = ["serde", "generic-snapshot-data", "storage-v2"] }
```
itegulov: Mega-nit: looks like latest version is 0.9.24

lib/network/src/service.rs
```diff
@@ -244,11 +246,14 @@ impl NetworkService {
...
+                    // stall raft leader election and block transaction processing.
                     .with_backoff_durations(PeerBackoffDurations {
-                        low: Duration::from_secs(30),
-                        medium: Duration::from_secs(60),
-                        high: Duration::from_secs(60 * 2),
-                        max: Duration::from_secs(60 * 3),
+                        low: Duration::from_secs(1),
+                        medium: Duration::from_secs(2),
+                        high: Duration::from_secs(5),
+                        max: Duration::from_secs(10),
```
itegulov, line 256: I don't think this will be much of an issue for now but note that this affects all network protocols. It might make sense to look into what high/max mean in this context (I think severity of network error, e.g. protocol-level breach vs simple connection severed)

lib/network/src/raft/wire.rs
```diff
@@ -0,0 +1,111 @@
...
+            id: RequestId,
+            resp: &'a Result<RaftResponse, String>,
+        }
+
+        let mut out = Vec::new();
+        match self {
+            RaftWireMessage::Request { id, req } => {
+                out.push(RAFT_REQUEST_MESSAGE_ID);
+                let payload = RequestPayload { id: *id, req };
+                out.extend(serde_json::to_vec(&payload).expect("serialize raft request payload"));
```
itegulov, line 61: Discussed on a call but I'll leave for history that this is using JSON instead of RLP (which is how it works in other places)

lib/network/src/raft/protocol.rs
```diff
@@ -0,0 +1,404 @@
...
+use std::fmt::Debug;
+use std::net::SocketAddr;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::task::{Context, Poll, ready};
+use tokio::sync::{mpsc, oneshot};
+use tokio::time::{Duration, Instant, sleep};
+
+pub const RAFT_PROTOCOL: &str = "raft";
```
itegulov: Nit: I'd use something more specific like `zks_raft`/`zks_consensus`

lib/network/src/raft/protocol.rs
```diff
@@ -0,0 +1,404 @@
...
+        req: RaftRequest,
+    ) -> Result<oneshot::Receiver<Result<RaftResponse, String>>, RaftTransportError> {
+        let Some(channels) = self.peers.get(&peer_id) else {
+            tracing::debug!(%peer_id, connected = self.peers.len(), "raft request failed: peer not connected");
+            return Err(RaftTransportError::NotConnected(peer_id));
+        };
+        let senders: Vec<_> = channels
+            .iter()
+            .map(|ch| (ch.connection_id, ch.sender.clone()))
+            .collect();
```
itegulov: Nit: isn't this the same as `channels.value().clone()` or something like that?

lib/network/src/raft/protocol.rs
```diff
@@ -0,0 +1,404 @@
...
+#[async_trait]
+pub trait RaftRequestHandler: Send + Sync + 'static {
+    async fn handle(&self, request: RaftRequest) -> Result<RaftResponse, String>;
+}
+
+#[derive(Debug, Clone)]
+pub struct RaftRouter {
+    next_request_id: Arc<AtomicU64>,
+    next_connection_id: Arc<AtomicU64>,
+    pending: Arc<DashMap<RequestId, oneshot::Sender<Result<RaftResponse, String>>>>,
```
itegulov: I think this map leaks when a connection is dropped while we are awaiting for the response. We only clean up when response arrives or if we failed to send request entirely.

lib/network/Cargo.toml
```diff
@@ -27,6 +28,12 @@ alloy = { workspace = true, default-features = false, features = [
 ] }
 alloy-rlp.workspace = true
 backon.workspace = true
+bincode.workspace = true
```
itegulov: Nit: unused?

lib/network/src/raft/protocol.rs
```diff
@@ -0,0 +1,404 @@
...
+        self,
+        supported: &reth_eth_wire::capability::SharedCapabilities,
+        _direction: Direction,
+        _peer_id: PeerId,
+    ) -> OnNotSupported {
+        if supported.iter_caps().any(|c| c.name() == RAFT_PROTOCOL) {
+            OnNotSupported::KeepAlive
+        } else {
+            OnNotSupported::Disconnect
+        }
```
itegulov: I think this will make us reject any peers that do not support at least one version of raft protocol. Probably not what you want here, i.e. peers that can only talk over zks should still be able to connect

lib/network/src/raft/protocol.rs
```diff
@@ -0,0 +1,404 @@
...
+
+        if let Some(fut) = &mut this.inflight
+            && let Poll::Ready(Some(buf)) = fut.as_mut().poll(cx)
+        {
+            this.inflight = None;
+            return Poll::Ready(Some(buf));
+        }
+
+        while let Poll::Ready(Some(outbound)) = this.outbound_rx.poll_recv(cx) {
```
itegulov: Note: this means we always prioritize replying over processing incoming messages. Not sure if this is a problem here but this can be problematic and lead to network "deadlock" in some cases

lib/network/src/raft/protocol.rs
```diff
@@ -0,0 +1,404 @@
...
+    outbound_queue: VecDeque<alloy::primitives::bytes::BytesMut>,
+    inflight: Option<
+        Pin<Box<dyn futures::Future<Output = Option<alloy::primitives::bytes::BytesMut>> + Send>>,
+    >,
+}
+
+impl Stream for RaftConnection {
+    type Item = alloy::primitives::bytes::BytesMut;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
itegulov: General comment: it's a lot harder to reason about these low-level Stream implementations. What I eventually did in zks protocol is I spawned a background async task that just writes replies into a channel and polled that channel in `Stream`. See `ZksConnection`

lib/network/src/service.rs
RomanBrodetski: provided comment with explanation

lib/network/src/raft/wire.rs
RomanBrodetski: added comment

RomanBrodetski:
@itegulov took your comments into account - please have another look

RomanBrodetski:
Nte that the PR to actually merge is : https://github.com/matter-labs/zksync-os-server/pull/1224, but leaving here to incremental review easioer

PR #1217, opened 2026-04-24
feat(mempool): validate intrinsic native resources for l2 txs

AntonD3:
## Summary

 Adds ZKsync OS specific stateless validation to the L2 mempool so that transactions whose gas_limit/fee caps cannot cover the intrinsic native resources (computational + pubdata) under the current chain fee params are rejected before reaching the pool. Previously, they would be dropped after an unsuccessful attempt to sequence, which opened a DDOS vector and affected ux.                                                     
                                              
  - Introduces ZkTransactionValidator, a wrapper around reth's EthTransactionValidator that carves out a seam in the validation pipeline (overriding validate_stateless, delegating validate_stateful).
  - Relocates FeeParams struct from zksync_os_sequencer::execution to zksync_os_types so the mempool can consume it without a sequencer dependency.
  - The validator holds RwLock<FeeParams>, L2Subpool::update_pending_fee_params propagates fresh params from the block context provider each time pending block fees are updated.                                                                                                                                                                                                                                                       
  - The check is stateless (no chain state needed — only self.fee_params and the tx), calls zk_os_api_dev::helpers::validate_l2_tx_intrinsic_native_resources, and reports failures as InvalidPoolTransactionError::IntrinsicGasTooLow.                                                                            
                                                                                                                                                                                                                                                                                                                   
 For now it's somewhat simplified, few improvements may be considered in the future:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
  - Fee params come from the latest block, If fee params change between admission and inclusion, a previously-admitted tx may become invalid (or the reverse). Can be refined later.                                                                                            
  - Validation uses zk_os_api_dev (pinned to the latest dev tag). On release this should move to latest dev at that point; longer-term the mempool may select a zksync-os version based on the chain's active protocol version.

## Breaking Changes

- Extra check in the mempool may affect API methods to send transaction

# NOTE: before 0.3.0 native intrinsic cost is much smaller, so merging this PR before will reject valid txs. Ideally should be merged when v31 upgrade is executed


lib/mempool/src/subpools/l2/validator.rs
```diff
@@ -0,0 +1,218 @@
...
+                    l.len() as u64,
+                    l.iter().map(|i| i.storage_keys.len()).sum::<usize>() as u64,
+                )
+            })
+            .unwrap_or((0, 0));
+        let authorization_list_num = transaction
+            .authorization_list()
+            .map(|l| l.len())
+            .unwrap_or(0) as u64;
+        validate_l2_tx_intrinsic_native_resources(
```
perekopskiy, line 100: The failing test (`low_fee_tx_does_not_hang_block_executor`) caught that `validate_l2_tx_intrinsic_native_resources` disallows txs with max_fee_per_gas less than base_fee which is incorrect behavior, it should be fixed. Please extend our test suite with positive and negative tests, i.e. validation passes for cases where it's supposed to pass and fails where it should fail.

perekopskiy (commented):
As discussed in slack, it won't be convenient to wait until v31 to merge it, instead I suggest enabling this new validation based on if current_protocol_version >= v31 (see node/bin/src/lib.rs)

lib/mempool/src/subpools/l2/validator.rs
AntonD3: Added tests. For now I just don't check intrinsic native limit for txs when max_fee_per_gas < base_fee, it's a bit complicated to estimate gas/native price in such case. And given that our base_fee is almost constant, it shouldn't be common case. WDYT?

lib/mempool/src/subpools/l2/validator.rs
```diff
@@ -0,0 +1,238 @@
...
+///             ├─ self.validate_stateless   (our override, then calls inner.validate_stateless)
+///             └─ inner.validate_stateful   (delegated to EthTransactionValidator)
+/// ```
+#[derive(Debug)]
+pub(crate) struct ZkTransactionValidator<Client, Tx> {
+    inner: EthTransactionValidator<Client, Tx, EthEvmConfig>,
+    fee_params: RwLock<FeeParams>,
+    /// Execution version expected for the next produced block. Drives version-gated
+    /// stateless checks (e.g. intrinsic native resources, available from V6).
+    execution_version: RwLock<ExecutionVersion>,
```
itegulov: Why not hold protocol version instead? AFAIU we want to get rid of execution version eventually as we established proper semantics for protocol versions now

From a brief look you do protocol_version<->execution_version conversions in many places anyway

lib/mempool/src/subpools/l2/validator.rs
perekopskiy: > And given that our base_fee is almost constant, it shouldn't be common case

Well, it fluctuates with base token price but indeed it shouldn't be the common case that someone sends tx with `max_fee_per_gas < base_fee`. IIUC the point of `validate_l2_tx_intrinsic_native_resources` is to improve UX by rejecting txs early and more importantly returning proper error message; I think it's ok to have this check skipped for txs with `max_fee_per_gas < base_fee`, such txs will likely sit in mempool for some time and so we indeed cannot do this validation check for them reliably

lib/mempool/src/subpools/l2/validator.rs
```diff
@@ -0,0 +1,235 @@
...
+    fee_params: RwLock<FeeParams>,
+    /// Protocol version expected for the next produced block. Drives version-gated
+    /// stateless checks (e.g. intrinsic native resources, available from v31 / execution V6).
+    protocol_version: RwLock<ProtocolSemanticVersion>,
+}
+
+impl<Client, Tx> ZkTransactionValidator<Client, Tx> {
+    pub(crate) fn new(
+        inner: EthTransactionValidator<Client, Tx, EthEvmConfig>,
+        protocol_version: ProtocolSemanticVersion,
```
itegulov: Let's not require `protocol_version` in constructor. I am actually trying to get rid of `current_protocol_version` usages in `node/bin/src/lib.rs`. We replay at least one block before we start producing more so IMHO this can start as `None` and then populated later via `update_protocol_version`

lib/mempool/src/subpools/l2/validator.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn validate_stateless(
+        &self,
+        origin: TransactionOrigin,
+        transaction: &Tx,
+    ) -> Result<(), InvalidPoolTransactionError> {
+        if self
+            .protocol_version
+            .read()
+            .expect("lock poisoned")
+            .is_post_v31()
```
itegulov: Is the logic going to be the same for v31/v32/etc? If not maybe it makes sense to future proof with match on protocol version so we don't forget about it in the future.

lib/mempool/src/pool.rs
```diff
@@ -354,13 +354,14 @@ impl<T: L2Subpool> Pool<T> {
 
     pub fn update_pending_block_fees(
```
itegulov, line 264: So this only gets called in `produce()`, meaning for ENs `fee_params` will always be zero base fee / zero pubdata / native price. I think this has some implications as ENs will accept such transactions into their mempool but MN will reject them, right?

node/bin/src/lib.rs
```diff
@@ -665,6 +659,14 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
         );
     }
 
+    tracing::info!("Initializing mempools");
+    let l2_subpool = zksync_os_mempool::subpools::l2::in_memory(
+        zk_provider_factory.clone(),
+        config.mempool_config.clone().into(),
+        config.tx_validator_config.clone().into(),
+        current_protocol_version.clone(),
+    );
```
itegulov: I think if you get rid of `current_protocol_version` you also won't need to move this code block down

integration-tests/tests/node/mempool.rs
```diff
@@ -226,3 +230,120 @@ async fn low_fee_tx_does_not_hang_block_executor(env: TestEnvironment) -> anyhow
...
+        .expect_successful_receipt()
+        .await?;
+
+    Ok(())
+}
+
+/// Gating: on V5 the V6 intrinsic-native check is bypassed. The same low-gas tx that the
+/// negative test rejects on V6 must be accepted at mempool ingress here.
+#[test_multisetup([CURRENT_TO_L1])]
+async fn intrinsic_native_check_disabled_pre_v6(env: TestEnvironment) -> anyhow::Result<()> {
```
itegulov: Nit: `pre_v6` -> `pre_v31`
I'd generally avoid mentioning V5/V6 across the code, we are trying to get rid of this abstraction anyway

lib/mempool/src/pool.rs
AntonD3: I also added a call in `on_canonical_state_change`, so should be called for EN when block is replayed. It's also called for MN, but should be no op(as values are the same as in produce call).

One more thing I noticed, but not directly related to my changes - I guess for same reason we may not update base_fee for the EN(for reth mempool), do you think it might be a problem?

lib/mempool/src/subpools/l2/validator.rs
AntonD3: For post_v31 we want to keep the intrinsic native check, but the only difference is that we may update constant and zksync_os_api version should be different

lib/mempool/src/pool.rs
itegulov: I think it's fine, we don't depend on it in RPC

lib/mempool/src/pool.rs
```diff
@@ -443,6 +444,21 @@ impl<T: L2Subpool> Pool<T> {
...
+            .update_pending_protocol_version(replay_record.protocol_version.clone());
+        // Refresh the validator's fee params from the executed block's context. This is the only
+        // fee source on nodes that don't produce blocks (external nodes never call
+        // `update_pending_block_fees`); on the main node these values are overwritten with the
+        // pending block's params at the start of each `produce()`.
+        self.l2_subpool.update_pending_fee_params(FeeParams {
+            eip1559_basefee: replay_record.block_context.eip1559_basefee,
+            native_price: replay_record.block_context.native_price,
+            pubdata_price: replay_record.block_context.pubdata_price,
+        });
```
itegulov, line 365: Hmm, can't `native_price` be `0` here if the chain has forced 0 fees? AFAIU this will make us reject all txs

lib/mempool/src/subpools/l2/validator.rs
```diff
@@ -0,0 +1,235 @@
...
+        if self
+            .protocol_version
+            .read()
+            .expect("lock poisoned")
+            .as_ref()
+            .is_some_and(ProtocolSemanticVersion::is_post_v31)
+        {
+            self.validate_intrinsic_native_resources(transaction)?;
+        }
+        self.inner.validate_stateless(origin, transaction)
```
itegulov: I think it's better to run `self.inner.validate_stateless` before `self.validate_intrinsic_native_resources`. Otherwise there will be some errors that are mislabeled as `IntrinsicGasTooLow`. For example if `max_priority_fee_per_gas > max_fee_per_gas` which is also asserted inside `validate_l2_tx_intrinsic_native_resources`.

itegulov (dismissed):
LGTM, a couple of final comments and I think we can merge

lib/mempool/src/pool.rs
AntonD3: Currently, nattive_price = 0 is not supported at all, zksync os execution will fail as well, for 0 fee chain should use base_fee 0, and native_price != 0. 

But actually good that you asked as with 0.4.0 native_price == 0 will be allowed(and handled as 0 fees), so I'll have update the `validate_l2_tx_intrinsic_native_resources` as well

itegulov (approved):

PR #1219, opened 2026-04-27
fix(sender): sender config no longer bottle-necked by pipeline constants

kostia244:
## Summary

We already have configurable limit on transactions sent to L1 in parallel. However the backpressure mechanism (OUTPUT_BUFFER_SIZE) limits us due to the way we read things from the channel. I increased it to 30, I think this is large enough limit.

### DIFF
--- a/lib/l1_sender/src/lib.rs
+++ b/lib/l1_sender/src/lib.rs
@@ -165,7 +165,7 @@ pub async fn run_l1_sender<Input: SendToL1>(
     // At this point, all in-flight transactions from the previous session are confirmed.
     // Only actual SendToL1 commands are expected from here on.
 
-    /// TODO: prefetched channel is a temporary solution, until OUTPUT_BUFFER_SIZE is replaced by a different mechanism.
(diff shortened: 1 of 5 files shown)


Artemka374:
so AFAICT there's an issue by design - we consume 16 commands from the channel, which's size is only 5, I wonder what will happen if we just increase the channel size

RomanBrodetski:
@Artemka374 very good point, thanks for noticing this

kostia244:
I did as Artem suggested. Are you happy with the current solutoin?

Artemka374 (approved):
Overall the solution looks logical to me, but I'm not sure about the tradeoffs we receive with it, so I suggest to look closely after deployment

kostia244:
I can test this with realistic scenario. I'll accumulate a lot of uncommitted batches on stage and see how it goes.
After running this test a couple times we can see if the config value is reasonable.
Mainnet is a bit different, I'll think about it once I have data from stage.
Any other suggestions on testing?

Artemka374:
yeah I think it should be good enough

PR #1221, opened 2026-04-27
feat(observability): add push exporter support & a metric to track REVM divergencies that can be used for alerting

kostia244:
## Summary

Add support for push exporter & push metrics. All metric groups with `PushMetrics` suffix use push exporter, the rest use pull.
Push exporter has a graceful shutdown, so we can report data right before shutdown.

New push metric `revm_divergences_detected` tracks divergencies, so we can finally set up the alert for them.


lib/revm_consistency_checker/src/node.rs
```diff
@@ -59,15 +72,10 @@ where
...
+                "Adding {} new addresses to L2 signer blacklist due to REVM divergence",
                 new_blacklist_size - initial_blacklist_size
             );
 
-            let message = format!(
-                "REVM consistency check failed for block number {}, block hash {}",
-                replay_record.block_context.block_number,
-                block_output.header.hash(),
-            );
             self.internal_config_manager
```
kostia244, line 81: We don't really have a choice but to crash here. No matter how the pipeline component stops, the previous one will panic because "outbound channel closed".

The other way would be wait for prometheus and then crash. Which just means building a few more blocks that will go to replay storage and later will have to be rebuilt. Not ideal.

node/bin/src/main.rs
```diff
@@ -188,13 +197,20 @@ pub async fn main() {
...
             if let Ok(Err(err)) = task_manager_result {
                 tracing::error!("shutting down due to error");
                 eprintln!("Error: {err:?}");
-                std::process::exit(1);
+                exit_code = 1;
             }
+            let prometheus_delay = prometheus_config.get_unexpected_event_shutdown_delay();
+            if !prometheus_delay.is_zero() {
+                runtime.graceful_shutdown_with_timeout(prometheus_delay);
+            }
```
kostia244: As long as we crash, which is the only option unless we add extra state to the node(to report when we restart, which starts to resemble an internal alert management system). We must have a timeout here for the reporting. 

Note that this only happens when we hit such events AND crash (i.e. revm_revert_on_divergence = true), which is already a messed up situation.

lib/observability/src/metrics.rs
```diff
@@ -26,6 +26,15 @@ pub struct GeneralMetrics {
 
     /// Number of blacklisted addresses in the internal config on server startup
     pub blacklisted_addresses_count: Gauge<usize>,
+
+    /// Counts unexpected events by event type.
```
RomanBrodetski:
```suggestion
    /// Counts unexpected events by event type. Can be used for alerting.
```

lib/observability/src/metrics.rs
```diff
@@ -26,6 +26,15 @@ pub struct GeneralMetrics {
...
+    /// Increment the counter once when the event occurs.
+    /// On internal crash shutdowns, the node may keep the Prometheus pull endpoint
```
RomanBrodetski:
```suggestion
    /// On internal crash shutdowns, the node will keep the Prometheus pull endpoint
```

lib/observability/src/prometheus.rs
```diff
@@ -51,14 +56,52 @@ impl PrometheusExporterConfig {
...
+        let transport = self.transport.clone();
+
+        // In pull mode, delay shutdown after unexpected events so Prometheus has one
+        // more chance to scrape the endpoint.
+        let shutdown = async {
+            let guard = shutdown_clone.await;
+            let delay = self.get_unexpected_event_shutdown_delay();
+            if !delay.is_zero() {
+                tracing::info!(
+                    "unexpected events were recorded; waiting {delay:?} before shutdown to allow one more Prometheus scrape"
```
RomanBrodetski: nit:
non-local reasoning / questinable method extraction: inside `get_unexpected_event_shutdown_delay` you c heck whether there were unexpected events, and if not it retunrns `0`. Then in the invocation code we rely on logic of `get_unexpected_event_shutdown_delay` - we depend on the fact that it returns zero when there are no unexpected events. 

Let's make a method `unexpected_events_reported -> bool` instead.

RomanBrodetski: But thinking further.. Are you sure we only want to sleep if there were unexpected events? Maybe always sleep? I think this will only affect kubernetes - locally with control+C it will stil stop immediately

node/bin/src/main.rs
RomanBrodetski: ok, we need @itegulov 's opinion/suggestion here.. I was thinking there is a way to do that without explicit wait in this file

lib/observability/src/prometheus.rs
```diff
@@ -51,14 +56,52 @@ impl PrometheusExporterConfig {
...
+                );
+                tokio::time::sleep(delay).await;
```
slowli: AFAIU, this can block node shutdown for longer than necessary (i.e., after the additional scrape we're waiting for has already occurred). Hence, I'd rather implement the termination functionality in `vise-exporter` itself, so that it's more efficient – i.e., allow configuring an additional scrape on shutdown with the configurable upper time bound. And I'd agree with @RomanBrodetski that this logic doesn't need to depend on existence of specific "important" metrics to report.

node/bin/src/config/mod.rs
```diff
@@ -1045,6 +1045,11 @@ pub struct PrometheusConfig {
     /// Port to expose Prometheus metrics on.
     #[config(default_t = 3312)]
     pub port: u16,
+
+    /// Extra pull-mode shutdown delay after unexpected events.
+    /// Set this slightly above the Prometheus scrape interval to allow one more scrape.
+    #[config(default_t = Duration::from_secs(60))]
```
slowli: Isn't this default value kind of high? I recall that at least for Era, the scrape interval was *much* smaller (order of 100ms). I understand that it the additional shutdown delay should almost never be triggered, but it would still degrade UX when it does.

kostia244:
```
❯ kubectl get servicemonitor sequencer-a -n testnet-genlayer -o yaml | grep metrics --context=5
  resourceVersion: "231668939"
  uid: 624cc16e-9707-4018-8ebd-4d8dd8a698a3
spec:
  endpoints:
  - interval: 30s
    path: /metrics
    port: metrics
    scrapeTimeout: 30s
  namespaceSelector:
    matchNames:
    - testnet-genlayer
  selector:
  ```
  
  It seems to be 30 for most atlas envs if I understand correctly?

Comment (kostia244) [reply node/bin/src/config/mod.rs]: @RomanBrodetski you were also interested in the downtime in this case.


Consider the case where consistency checker is enabled with reverts for inconsistencies. 

When we enter the `RevmConsistencyChecker` the block is already in the RPC. If it fails, we want to stop both sequencer and RPC immediately because we know that the new blocks will be rebuilt.

As for external nodes, they should have the revert enabled as well. Otherwise: 
- These nodes won’t restart and will keep serving RPC requests based on the incorrect block until the main node goes back up
- Batch Verification Client will break until restarted(which would have to be done manually; I haven’t figure out why this happens yet)

Waiting for prometheus scrape increases the amount of time main node is down.
We really don’t want to do this in general case:
- Incoming transactions are rejected while main node is down(ENs have nowhere to forward to)
- Many chains produce blocks at a faster pace than 30 seconds.

In case of divergence, the impact is higher:
- ENs with `revert=true` will be down until main node goes back up
- This means all the nodes are down for `scrape_interval`

Possible approaches I see:
Graceful shutdown for prometheus always:
 - big downtime (read-only RPCs still work)
 - Can we reduce the scraped interval, then it would just work / Why is it 30s? (cc: @RomanBrodetski )
 
Graceful shutdown when revms problems:
 - still quite a bit of downtime, but only in these rare cases.
 
Store last revert info together with internal config: 
 - “why don’t we use metrics or logs for this”
 - This avoids the downtime
 
Use push gateway:
- Avoids downtime
- Infra changes everywhere

File: lib/observability/src/prometheus.rs [hunk 8]
```diff
@@ -53,12 +53,21 @@ impl PrometheusExporterConfig {
...
+    }
+
     /// Runs the exporter. This future should be spawned in a separate Tokio task.
     pub async fn run(self, shutdown: GracefulShutdown) -> anyhow::Result<()> {
         tokio_runtime::register_monitor();
-        let registry = MetricsCollection::lazy().collect();
-        let metrics_exporter =
-            MetricsExporter::new(registry.into()).with_graceful_shutdown(shutdown.ignore_guard());
+        let registry = self.registry();
+        // ignore_guard will drop the guard too ealry, so clone is used.
```
slowli: Typo: "ealry" -> "early".

lib/revm_consistency_checker/src/node.rs
```diff
@@ -4,10 +4,11 @@ use reth_revm::ExecuteCommitEvm;
 use reth_revm::context::{Context, ContextTr};
 use reth_revm::db::CacheDB;
 use std::collections::HashSet;
+use std::time::{SystemTime, UNIX_EPOCH};
```
slowli: Unused imports.

node/bin/src/main.rs
```diff
@@ -186,14 +194,24 @@ pub async fn main() {
...
         .take_task_manager_handle()
         .expect("Runtime must contain a TaskManager handle");
 
     tokio::select! {
         task_manager_result = task_manager_handle => {
             if let Ok(Err(err)) = task_manager_result {
-                tracing::error!("shutting down due to error");
+                tracing::error!(%err, "shutting down due to critical task error");
+                /// Graceful shutdown for push exporter. Very fast.
+                /// This is needed for REVM alert.
```
slowli: Nit: Doc comments are weird when not placed on items.

lib/observability/src/metrics.rs
```diff
@@ -30,3 +30,18 @@ pub struct GeneralMetrics {
...
 pub static GENERAL_METRICS: vise::Global<GeneralMetrics> = vise::Global::new();
+
+/// These metrics are exported with push exporter
+/// Use case: report something right before stopping the node, without waiting for scrape
+/// so these metrics can be used for alerts
+///
+/// Be careful: There can be  some differences in how the metrics are handled by prometheus compared
+/// To pull exporter
+#[derive(Debug, Metrics)]
+pub struct PushMetrics {
```
slowli: Bikeshedding: Since metrics filtering for the exporter can be based on names, there's no need to define the metrics here; they could be defined at the use site (i.e., the `revm_consistency_checker` crate) and can be non-public.

slowli (commented):
I like the overall approach.

lib/observability/src/prometheus.rs
```diff
@@ -53,12 +53,21 @@ impl PrometheusExporterConfig {
         format!("{base_url}/metrics/job/{job_id}/namespace/{namespace}/pod/{pod}")
     }
 
+    /// Get the list of metrics that use this type of exporter (Push vs Pull)
+    fn registry(&self) -> Registry {
+        let is_push_exporter = matches!(self.transport, PrometheusTransport::Push { .. });
+        MetricsCollection::lazy()
+            .filter(|group| (group.name == "PushMetrics") == is_push_exporter)
```
RomanBrodetski: I didn't like that we rely on metrics name here to disttinguish.. so that if someone renames the struct in Rust, we'll silently stop pushing those metrics? But maybe there is no better way - then it's ok, but let's leave a comment where this struct is defined

node/bin/src/main.rs
```diff
@@ -186,14 +194,24 @@ pub async fn main() {
         }
     });
 
+    let prometheus_push_shutdown = if ephemeral_enabled {
+        tracing::info!("Ephemeral mode enabled, skipping Prometheus push exporter");
+        None
+    } else {
+        spawn_prometheus_push_exporter(&runtime, &prometheus_config)
+    };
+
```
RomanBrodetski, line 198: hmm can this statement be put inside the one above? We already deal with prometheus there, and there is already an `if` statement in the previous statement.

Ah maybe the other one is in a cloisure  and that's why it's not easy.. But please see if this can be simplify - feel free to leave as is if not

RomanBrodetski (commented):
Overall looks good and clean!

lib/observability/src/prometheus.rs
kostia244: I added a comment, this seems like the most appropriate solution to me. Other options:

- modify vise to have this distinction internally
- use marker variable (it will get exported as a metric)
- Move the Push struct to observability and only have a single struct for this push exporter
- keep things as they are, with a comment



WDYT?

slowli: I don't like the idea of introducing push / pull distinction into `vise` per se because it seems too specific. Right now, [metrics group metadata](https://matter-labs.github.io/vise/vise/descriptors/struct.MetricGroupDescriptor.html) doesn't contain much info useful for filtering besides the name. It would be relatively straightforward to add support of tags, i.e. something like

```rust
#[derive(Debug, Metrics)]
#[metrics(tags("push"))]
struct PushMetrics {
    // ...
}

// ...translated into
pub struct MetricGroupDescriptor {
    // other fields...
    pub tags: &'static [&'static str],
}
```

...and then, filtering could be performed by tags. 

> Move the Push struct to observability and only have a single struct for this push exporter

This could work, but IMO it's only feasible as a long-term solution if the push metrics are (almost) never extended in the future. Otherwise, this approach would eventually introduce a god object.

RomanBrodetski (approved):

PR #1222, opened 2026-04-27
feat(l1-sender): make configurable gas limit for L1

Artemka374:
## Summary

L1 sender now has different gas limits depending on SL. 
In case it's Gateway - the gas limit stays the same 15M, since transactions there can consume large amounts of gas(up to 10M even in interop integration tests)
In case it's L1 - there's a separate config value, which is by default set to 2M - judging by statistics, the max amount of gas consumed is ~500k. Left there as a config value just for a case it will need further adjustments in the future


cytadela8 (commented):
I think this will cause us to have transactions run out of gas when there is a lot of L2->L1 operations.

PR #1224, opened 2026-04-27
feat(network): add raft transport layer over devp2p

romanbrodetski-ai:
## Summary

Next PR in consensus series. Introduces the changes to networking - that will later be used by raft engine.

- Adds `openraft` dependency
- Adds `lib/consensus_types`: shared openraft type config (`RaftTypeConfig`) and `PeerId`-based node identity
- Adds `lib/network/src/raft/`: `RaftProtocolHandler` (a new RLPx sub-protocol), wire encoding for AppendEntries/Vote/InstallSnapshot RPCs, and `RaftRouter` for routing outbound RPCs and matching responses
- `NetworkService::new` gains an optional `raft_handler` parameter; registers the raft sub-protocol when provided
- Tightens peer backoff durations (30–180 s → 1–10 s) so consensus nodes reconnect quickly after restarts

### DIFF
--- a/lib/network/src/raft/protocol.rs
+++ b/lib/network/src/raft/protocol.rs
@@ -1,0 +1,426 @@
+use crate::raft::wire::{RaftRequest, RaftResponse, RaftWireMessage, RequestId};
+use async_trait::async_trait;
+use dashmap::DashMap;
+use futures::StreamExt;
+use reth_eth_wire::multiplex::ProtocolConnection;
+use reth_eth_wire::protocol::Protocol;
+use reth_network::Direction;
+use reth_network::protocol::{ConnectionHandler, OnNotSupported, ProtocolHandler};
+use reth_network::types::Capability;
+use reth_network_peers::PeerId;
+use std::fmt::Debug;
+use std::net::SocketAddr;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::task::{Context, Poll};
+use tokio::sync::{mpsc, oneshot};
+use tokio::time::{Duration, Instant, sleep};
+use tracing::Instrument;
+
+pub const RAFT_PROTOCOL: &str = "zks_raft";
+const RAFT_PROTOCOL_VERSION: usize = 1;
+// RLPx multiplexing uses the first byte as a sub-protocol message id.
+// Raft has two wire message kinds (request/response), so it needs 2 slots.
+const RAFT_PROTOCOL_MESSAGE_COUNT: u8 = 2;
+const RAFT_OUTBOUND_CHANNEL_CAPACITY: usize = 64;
+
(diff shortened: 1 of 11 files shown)


lib/network/src/service.rs
```diff
@@ -173,6 +174,7 @@ impl NetworkService {
         protocol_config: ZksProtocolConfig,
         replay: impl ReadReplay + Clone,
         client: impl ChainSpecProvider<ChainSpec: Hardforks> + BlockNumReader + 'static,
+        raft_handler: Option<RaftProtocolHandler>,
```
itegulov, line 177: Can be addressed later, but I do not think it's the right level of encapsulation to accept `RaftProtocolHandler` here. It should probably be initialized internally based on config

lib/network/src/raft/protocol.rs
```diff
@@ -0,0 +1,426 @@
...
+                    tracing::info!(%peer_id, connection_id, "raft connection closed by peer");
+                    break;
+                };
+                match RaftWireMessage::decode(&bytes[..]) {
+                    Ok(RaftWireMessage::Request { id, req }) => {
+                        tracing::debug!(%peer_id, request_id = id, "received raft request");
+                        let handler = handler.clone();
+                        let outbound_tx = outbound_tx.clone();
+                        tokio::spawn(async move {
+                            let resp = handler.handle(req).await;
```
itegulov, line 386: Hmm, do we need to spawn a concurrent job here? AFAIR canonical raft processes messages in order anyway. If we do need to spawn, then I'd at least accumulate these jobs in `JoinSet` so that we can abort them on drop.

lib/network/src/raft/protocol.rs
```diff
@@ -0,0 +1,426 @@
...
+    pub fn new(handler: impl RaftRequestHandler, router: RaftRouter) -> Self {
+        Self {
+            handler: Arc::new(handler),
+            router,
+        }
+    }
+
+    fn establish_connection(&self, peer_id: PeerId, conn: ProtocolConnection) -> RaftConnection {
+        let (outbound_tx, outbound_rx) = mpsc::channel(RAFT_OUTBOUND_CHANNEL_CAPACITY);
+        let (msg_tx, msg_rx) = mpsc::unbounded_channel();
```
itegulov, line 240: What's the semantics of outbound_{rx,tx} vs msg_{rx,tx} here? I am not sure if unbounded channel makes sense here but so far I failed to fully grasp the difference between the two. Why can't `RaftRouter::peers` hold outbound_tx instead?

itegulov (approved):
Left some final comments but overall happy to merge now and address in a follow-up

romanbrodetski-ai:
@itegulov thanks for comments - I'll take the opportunity and merge now, but will make sure to follow up (I have some other follow ups from previous comments of yours too!)

PR #1225, opened 2026-04-27
feat(consensus): Final consensus integration and integration tests

romanbrodetski-ai:
* enabled raft consensus (opt-in)
  * storage (very simple implementation of openraft trait - no atomicity assumptions etc)
  * `state_machine.rs` - main canonization logic - made as simple/clear as possible.
* adds comprehensive integration tests
  * multi_node.rs harness
  * consensus_tests.rs testsuite


docs/src/setup/multi_node_consensus.md
```diff
@@ -0,0 +1,126 @@
+# Multi-node consensus
+
+This document describes the two node roles in a multi-node setup:
+
+1. **ConsensusNode**: participates in Raft consensus and can become leader at any time.
+2. **ExternalNode**: does **not** participate in consensus; it downloads canonized blocks from `ZksProtocol` and replays them locally.
```
itegulov, line 6: Does EN understand anything about consensus? Can it query which node is the current leader and dynamically change the node it downloads blocks from?

docs/src/setup/multi_node_consensus.md
```diff
@@ -0,0 +1,126 @@
...
+A ConsensusNode will propose blocks when it is leader and will follow canonized blocks when it is a replica. An ExternalNode only replays canonized blocks and never proposes.
+
+> **Batcher subsystem (L1 settlement) is not yet highly available.** The batcher pipeline — proof generation, L1 batch submission — must be enabled on exactly one node via `batcher_enabled=true` (which is the default). The other consensus nodes should set `batcher_enabled=false`. Leader failover works for block production: if the node running the batcher fails, another node will be elected leader and continue producing blocks. However, those blocks will not be submitted to L1 until the original batcher node restarts or the cluster is manually reconfigured to enable the batcher on a different node.
+
+**ConsensusNode**
+
+Requirements:
+- Networking must be enabled (Raft RPC is transported over `lib/network`).
+- `consensus_enabled=true` and a list of Raft peer IDs (`consensus_peer_ids`).
+- Exactly one node should set `consensus_bootstrap=true` for initial cluster membership.
```
itegulov: Hmm, is this necessary? I imagine it will be awkward to maintain in infra

docs/src/setup/multi_node_consensus.md
```diff
@@ -0,0 +1,126 @@
...
+```bash
+ENODE_1="enode://246e07030b4c48b8f28ab1fdf797a02308b0ca724696b695aabee48ea48298ff221144a0c0f14ebf030aea6d5fb6b31bd3a02676204bb13e78336bb824e32f1d@127.0.0.1:3060"
+ENODE_2="enode://d2db8005d59694a5b79b7c58d4d375c60c9323837e852bbbfd05819621c48a4218cefa37baf39a164e2a6f6c1b34c379c4a72c7480b5fbcc379d1befb881e8fc@127.0.0.1:3061"
+ENODE_3="enode://2991880ae3ff81b881c86f54d2af0ee85a325231bb75f903f06f432020101614dbfbc75ddec885f5e101fed272bee661f492fb6dd80147b656da990635a7e581@127.0.0.1:3062"
+PEER_IDS_JSON='[
+  "0x246e07030b4c48b8f28ab1fdf797a02308b0ca724696b695aabee48ea48298ff221144a0c0f14ebf030aea6d5fb6b31bd3a02676204bb13e78336bb824e32f1d",
+  "0xd2db8005d59694a5b79b7c58d4d375c60c9323837e852bbbfd05819621c48a4218cefa37baf39a164e2a6f6c1b34c379c4a72c7480b5fbcc379d1befb881e8fc",
+  "0x2991880ae3ff81b881c86f54d2af0ee85a325231bb75f903f06f432020101614dbfbc75ddec885f5e101fed272bee661f492fb6dd80147b656da990635a7e581"
+]'
+BOOT_NODES="${ENODE_1},${ENODE_2},${ENODE_3}"
```
itegulov, line 48: This does not lead to discv5-level warnings that we are trying to re-discover ourselves? If you fixed it by filtering out our own peer id, then this is a welcome change (should've done so from the start). But wanted to flag in case you didn't

docs/src/setup/multi_node_consensus.md
itegulov: Potential simple idea is to have implicit requirement that first node from the consensus peer ids list must initialize the cluster membership

lib/raft/src/bootstrap.rs
```diff
@@ -0,0 +1,81 @@
...
+                tracing::info!("raft bootstrap process completed");
+                Ok(())
+            }
+            Err(openraft::error::RaftError::APIError(
+                openraft::error::InitializeError::NotAllowed(_),
+            )) => {
+                // Another node won the bootstrap race and initialized the cluster while
+                // we were waiting for peers. This is a normal multi-node startup scenario
+                // when multiple nodes have `bootstrap = true`. Safe to proceed.
+                tracing::info!("raft cluster became initialized meanwhile; skipping bootstrap");
```
itegulov, line 74: Hmm, it looks like having only one bootstrapping node is not a hard requirement?

lib/raft/src/leadership_monitor.rs
```diff
@@ -0,0 +1,99 @@
...
+/// The task exits automatically when the OpenRaft metrics channel closes or when all receivers
+/// for both output watch channels are dropped.
+pub fn spawn_leadership_monitor(
+    raft: Raft<RaftTypeConfig>,
+    node_id_str: String,
+    leader_tx: watch::Sender<ConsensusRole>,
+    status_tx: watch::Sender<Option<RaftConsensusStatus>>,
+) {
+    let mut metrics_rx = raft.metrics();
+    tokio::spawn(async move {
```
itegulov: Should this be spawned through `runtime` instead? Not sure if `Raft` can potentially hold any storage references etc. Also, this looks like a critical task to me judging by the fact sequencer depends on it

lib/raft/src/leadership_monitor.rs
```diff
@@ -0,0 +1,99 @@
...
+        let mut last_logged = None;
+        let mut leader_confirmed = false;
+        let mut prev_role = ConsensusRole::Replica;
+
+        loop {
+            if metrics_rx.changed().await.is_err() {
+                panic!("OpenRaft metrics channel closed unexpectedly; consensus subsystem is gone");
```
itegulov: Nit: log + break instead to not spam with panics

lib/raft/src/model.rs
```diff
@@ -1,15 +1,29 @@
...
     pub canonization_engine: BlockCanonizationEngine,
     pub leadership: LeadershipSignal,
-    pub network_protocol: ConsensusNetworkProtocol,
-    pub bootstrapper: ConsensusBootstrapper,
-    pub status: ConsensusStatusSource,
+    pub raft: Option<RaftRuntimeExtras>,
+    /// A handle to the OpenRaft engine for clean shutdown. The caller must await
+    /// `raft_handle.shutdown()` during node teardown so that the RaftCore task — and
+    /// the RocksDB handles it owns — are released before the process cleans up.
+    pub raft_handle: Option<Raft<RaftTypeConfig>>,
```
itegulov: Maybe it makes sense to spawn a critical task that waits for `raft_handle.shutdown()` internally here as opposed to `node/bin/src/lib.rs`?

itegulov: Nit: maybe I am missing something but can't this be a part of `RaftRuntimeExtras`? You won't need two `Option`s this way when they are logically tied if I understood the code correctly

lib/raft/src/storage.rs
```diff
@@ -0,0 +1,421 @@
...
+
+    /// Returns a clone of the underlying raft RocksDB handle.
+    pub(crate) fn db(&self) -> RocksDB<RaftColumnFamily> {
+        self.db.clone()
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Default)]
+pub(crate) struct RaftStateMachineMeta {
+    pub(crate) last_membership: Option<StoredMembership<PeerId, RaftNode>>,
```
itegulov, line 194: Is it a good idea to tie our storage format to `StoredMembership` (external type from openraft)?

itegulov (commented):
Partial review for now, will finish later

docs/src/setup/multi_node_consensus.md
romanbrodetski-ai: yeah this is a stale comment sorry for that

docs/src/setup/multi_node_consensus.md
RomanBrodetski: it doesn't. As long as it connects to a node that's in sync - it shouldn't make a difference, since blocks are currently only exposed once canonized. 

But indeed if the node is falling behind, it could see thta and flip `/health` endpoint so that traffic is not directed there. I have a follow up planed for such behavoiur.

lib/raft/src/storage.rs
RomanBrodetski: I agree it's not optimal - look, for now I don't plan to support backward compatibility, so will address this in a follow-up (I have a list)

docs/src/setup/multi_node_consensus.md
romanbrodetski-ai: No, no `warn!`-level noise — but not because we filter ourselves out, rather because upstream discv5 already filters silently.

Trace:

1. We pass `config.boot_nodes` verbatim to reth's `.boot_nodes(...)` (`lib/network/src/service.rs:201,205`); no filter for the local enode.
2. reth-discv5's bootstrap loop (`crates/net/discv5/src/lib.rs:551-562`) handles `BootNode::Enode(...)` via `discv5.request_enr(enode.to_string()).await` and logs failures at `debug!` only.
3. discv5 0.10.4 has a built-in self-request filter at `src/handler/mod.rs:472-475`:
   ```rust
   if self.listen_sockets.contains(&node_address.socket_addr) {
       debug!("Filtered request to self");
       return Err(RequestError::SelfRequest);
   }
   ```
   It short-circuits with `SelfRequest` and a `debug!` log. The error bubbles up and reth logs it at `debug!` too — nothing surfaces at default tracing level.
4. Separately, `src/service.rs:1255-1258` silently drops any ENR returned in query results that matches `local_id`.

One caveat worth flagging: the discv5 self-filter is **socket-based**, not peer-id-based. In the docs example each node's listen socket matches exactly one boot entry (its own), so it's fine. A misconfiguration that listed two enodes sharing the local node's port but with different node_ids would get one filtered as "self" regardless of node_id — but that's a config error, not something we'd design around.

lib/raft/src/leadership_monitor.rs
romanbrodetski-ai: Good catch — switched to `runtime.spawn_critical_task("raft leadership monitor", ...)` in 4cf4dbaa. Threaded `&Runtime` through `init_consensus` → `spawn_leadership_monitor`. Now panics propagate to `TaskManager` and the task is registered alongside the rest of the consensus subsystem.

lib/raft/src/model.rs
romanbrodetski-ai: You're right — they're `Some` together. Folded `raft_handle` into `RaftRuntimeExtras` as a `handle` field in acbb7e87, so the call site has just one `Option` to match on and the shutdown registration moved into that arm.

romanbrodetski-ai: Done in 6e43d69b — moved the `raft-shutdown` critical task spawn into `init_consensus`. The lifecycle is now owned by `lib/raft`, and `RaftRuntimeExtras` no longer needs to expose the handle (`OpenRaftCanonizationEngine` already holds its own clone for the canonization path).

lib/raft/src/leadership_monitor.rs
romanbrodetski-ai: Yes, still relevant — even more so after the previous comment moved this onto `spawn_critical_task`: the metrics channel closes on graceful shutdown when `raft.shutdown()` runs, so a panic would now trigger a TaskManager alarm on the clean teardown path. Switched to `tracing::info!` + `break` in 60e5fd9a.

node/bin/src/config/mod.rs
```diff
@@ -1996,4 +2116,80 @@ mod tests {
...
+            reset_timestamps: false,
+        });
+
+        let err = config.validate().await.unwrap_err().to_string();
+
+        assert!(err.contains("`sequencer.block_rebuild` requires `consensus.enabled=false`"));
+
+        config.consensus_config.enabled = false;
+        config.validate().await.unwrap();
+    }
```
RomanBrodetski: let's remove these tests

integration-tests/src/lib.rs
```diff
@@ -270,14 +272,15 @@ pub struct Tester {
...
+    ports: Ports,
 
     #[allow(dead_code)]
     tempdir: Arc<tempfile::TempDir>,
 
     // Needed to be able to connect external nodes
     node_record: NodeRecord,
     l2_rpc_address: String,
-    #[expect(dead_code)]
+    #[allow(dead_code)]
```
RomanBrodetski: can be removed?

.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ env:
   PROVER_GPU_TESTS: 'prover-gpu-tests'
   CUDAARCHS: '80;89;90'
   # Run tests 10 times in case of push to main branch to catch possible flakiness
-  NEXTEST_ITERATIONS: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && '--stress-count 10' || '' }}
+  NEXTEST_ITERATIONS: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && '--stress-count 10' || '--stress-count 30' }}
```
itegulov: I assume this is temporary? Don't forget to revert!

integration-tests/tests/consensus_node/mod.rs
itegulov, line 1: Please move this inside `consensus_node` directory and then add a `mod consensus_node` to `suite.rs`. Otherwise this will be a separate binary that we will need to link. This is why it used to take a long time to compile integration tests in the past (we had ~15 different binaries)

lib/raft/src/network.rs
```diff
@@ -0,0 +1,223 @@
...
+                .map_err(|e| e.to_string()),
+            RaftRequest::Vote(r) => self
+                .raft
+                .vote(r)
+                .await
+                .map(RaftResponse::Vote)
+                .map_err(|e| e.to_string()),
+            RaftRequest::InstallSnapshot(r) => {
+                #[allow(deprecated)]
+                let response = self.raft.install_snapshot(r).await;
```
itegulov, line 50: Why was this deprecated?

lib/raft/src/state_machine.rs
```diff
@@ -0,0 +1,216 @@
...
+                .iter()
+                .map(debug_display_raft_entry)
+                .collect::<Vec<_>>()
+                .join(", "),
+        );
+        let mut responses = Vec::new();
+
+        for entry in &entries {
+            match &entry.payload {
+                EntryPayload::Blank => responses.push(()),
```
itegulov, line 100: Is `Blank` needed for ping-pong or is there something else to it?

lib/raft/src/state_machine.rs
```diff
@@ -0,0 +1,216 @@
...
+
+        // If `RaftApplied` has entries beyond the latest WAL block, the process exited
+        // after `save_block_log_id` but before `BlockApplier` wrote those blocks to the WAL.
+        // OpenRaft will re-apply them — no data is lost, but worth logging.
+        self.log_pending_applied_blocks(latest_wal_block)?;
+
+        Ok((last_applied_log_id, membership))
+    }
+
+    async fn apply<I>(&mut self, entries: I) -> Result<Vec<()>, StorageError<PeerId>>
```
itegulov, line 81: The semantics of `Vec<()>` in the return type is not clear to me

itegulov: Oh, I see, it's just a requirement from the openraft trait

lib/raft/src/state_machine.rs
```diff
@@ -0,0 +1,216 @@
...
+        // on `wal.latest_record()`, we only advance `last_applied_log_id` once the block
+        // is durably in the WAL, so a crash before the WAL write causes OpenRaft to
+        // re-apply the missing entry on restart.
+        let latest_wal_block = self.wal.latest_record();
+        let last_applied_log_id = self.meta_store.load_block_log_id(latest_wal_block)?;
+
+        // If `RaftApplied` has entries beyond the latest WAL block, the process exited
+        // after `save_block_log_id` but before `BlockApplier` wrote those blocks to the WAL.
+        // OpenRaft will re-apply them — no data is lost, but worth logging.
+        self.log_pending_applied_blocks(latest_wal_block)?;
```
itegulov, line 76: So the method just logs and the actual re-application happens somewhere internally in Openraft -- correct?

lib/raft/src/network.rs
```diff
@@ -0,0 +1,223 @@
...
+        );
+        RaftNetworkClient {
+            router: factory.router,
+            peer_id: target,
+            timeout: factory.timeout,
+        }
+    }
+}
+
+pub type RaftNetworkFactory = RaftNetworkFactoryImpl;
```
itegulov: Nit: why not name the factory as `RaftNetworkFactory` in the first place?

lib/raft/src/network.rs
```diff
@@ -0,0 +1,223 @@
...
+}
+
+impl RaftNetwork<RaftTypeConfig> for RaftNetworkClient {
+    async fn append_entries(
+        &mut self,
+        rpc: AppendEntriesRequest<RaftTypeConfig>,
+        option: RPCOption,
+    ) -> Result<AppendEntriesResponse<PeerId>, RPCError<PeerId, RaftNode, RaftError<PeerId>>> {
+        let timeout_dur = std::cmp::min(self.timeout, option.hard_ttl());
+        let client = self.clone();
```
itegulov: Nit: do we have to clone here? Feels a bit weird

lib/raft/src/network.rs
```diff
@@ -0,0 +1,223 @@
...
+        }
+    }
+
+    async fn vote(
+        &mut self,
+        rpc: VoteRequest<PeerId>,
+        option: RPCOption,
+    ) -> Result<VoteResponse<PeerId>, RPCError<PeerId, RaftNode, RaftError<PeerId>>> {
+        let timeout_dur = std::cmp::min(self.timeout, option.hard_ttl());
+        let client = self.clone();
```
itegulov: Same thing here

lib/raft/src/network.rs
```diff
@@ -0,0 +1,223 @@
...
+    async fn install_snapshot(
+        &mut self,
+        rpc: InstallSnapshotRequest<RaftTypeConfig>,
+        option: RPCOption,
+    ) -> Result<
+        InstallSnapshotResponse<PeerId>,
+        RPCError<PeerId, RaftNode, RaftError<PeerId, openraft::error::InstallSnapshotError>>,
+    > {
+        let timeout_dur = std::cmp::min(self.timeout, option.hard_ttl());
+        let client = self.clone();
```
itegulov: And here

lib/status/src/lib.rs
```diff
@@ -1,27 +1,39 @@
...
-pub async fn run_status_server(addr: SocketAddr, shutdown: GracefulShutdown) {
+pub async fn run_status_server(
+    addr: SocketAddr,
+    shutdown: GracefulShutdown,
+    consensus_raft_status_rx: Option<watch::Receiver<Option<RaftConsensusStatus>>>,
+) -> anyhow::Result<()> {
     let app = Router::new()
         .route("/status/health", get(health))
-        .with_state(AppState {});
+        .route("/status", get(status))
```
itegulov, line 29: Hmm, is it a good idea to add extra endpoints for the status server? AFAIU it will not be exposed outside of the cluster so we won't be able to use it for debug. Maybe add new method to JSONRPC `unstable` or `debug` instead?

itegulov: Or will this be used by alerting somehow? Intent not clear to me

node/bin/src/node_state_on_startup.rs
```diff
@@ -30,13 +30,5 @@ impl NodeStateOnStartup {
             self.last_l1_proved_block,
             self.last_l1_executed_block,
         );
-        if self.node_role.is_main() {
-            assert!(
-                self.block_replay_storage_last_block >= self.last_l1_committed_block,
```
itegulov, line 35: This does not necessarily hold because they might be in consensus storage now and hence re-populated in runtime later?

node/bin/src/lib.rs
```diff
@@ -1688,6 +1739,44 @@ async fn find_last_matching_main_node_block(
...
+fn prepare_raft_storage(config: &Config) -> anyhow::Result<()> {
+    let raft_storage_path = config.general_config.rocks_db_path.join(RAFT_DB_NAME);
+    if config.consensus_config.force_clear_raft_history
+        && raft_storage_path_exists(&raft_storage_path)?
+    {
+        tracing::warn!(
+            path = %raft_storage_path.display(),
+            "force-clearing persisted raft history before startup"
+        );
+        std::fs::remove_dir_all(&raft_storage_path).with_context(|| {
```
itegulov: This feels a little scary. Is there no way to initialize RocksDB and then make it delete only files it tracks?

Otherwise we could nuke something much larger by accident

node/bin/src/config/mod.rs
```diff
@@ -534,6 +539,104 @@ impl NetworkConfig {
...
+    #[config_validate(custom(
+        |root: &Config, value: &bool| !*value || root.network_config.secret_key.is_some(),
+        "requires `network.secret_key`"
+    ))]
+    pub enabled: bool,
+    /// Delete persisted OpenRaft state before startup.
+    ///
+    /// This is intended for intentionally switching a node away from previously persisted
+    /// consensus history. Without clearing this state, starting with consensus disabled is
+    /// rejected because later re-enabling consensus could result in an invalid state.
```
itegulov, line 578: Why? Because we do not guarantee backwards compatibility or is there a deeper reason?

node/bin/src/config/mod.rs
```diff
@@ -534,6 +539,104 @@ impl NetworkConfig {
...
+                secret_key,
+            ).id;
+            value.contains(&local_peer_id)
+        },
+        "must include local peer id derived from `network.secret_key`"
+    ))]
+    pub peer_ids: Vec<PeerId>,
+    /// WARNING: Assumes all configured consensus nodes are already caught up to the same
+    /// canonical L2 state. Bootstrap does not catch up stale nodes before admitting them
+    /// to the cluster.
```
itegulov, line 607: What does this mean realistically? If one consensus node had fatal failure and needs to catch-up from scratch it can still do so using zks subprotocol, right?

RomanBrodetski: oh yes, it can definitely catch up. It's just that, before enabling consensus, one needs to prepare - make sure the nodes it wants to include into cluster are catched up. 
I plan to improve this in one of the follow-ups

node/bin/src/config/mod.rs
RomanBrodetski: If there were some blocks mined that are not reflected in raaft (no corresponding raft messages are recorded), we don't have a way to make raft catch up after already mined blocks. This can be fixed, but for simplicity I decided to make strongetr assumptions here. Node will detect such state and panic, so I don't see this as an added risk

lib/status/src/lib.rs
RomanBrodetski: now it's used in tests and locally.. 
JSON RPC is not ideal because you are anyway interested in indicvidual nodes - and JSON RPC is normally behind a loadbalancer.. So maybe the only way to use in our setup is to expose ports first? 
So, for now I don't see a better alternative.

lib/raft/src/state_machine.rs
RomanBrodetski: exactly!

lib/raft/src/state_machine.rs
RomanBrodetski: yeah..

lib/raft/src/network.rs
romanbrodetski-ai: openraft deprecated `install_snapshot` in favour of the streaming `full_snapshot` API. We do not use snapshotting at all — `full_snapshot` is implemented as a no-op error — but a remote peer built against an older openraft revision may still send an InstallSnapshot RPC, so we need the handler. `#[allow(deprecated)]` silences the compiler warning while we remain on the non-streaming path.

lib/raft/src/state_machine.rs
romanbrodetski-ai: Not for ping-pong — Blank entries are no-ops that a newly elected leader appends to its own log so it can commit all prior entries from its predecessor. OpenRaft calls `apply()` with them like any other entry; we must return a response per the trait contract. Without this arm the match would be non-exhaustive.

lib/raft/src/state_machine.rs
romanbrodetski-ai: Correct — this method only logs. Re-application is driven by openraft: `applied_state()` returns the WAL's last committed block ID (which may lag behind entries in `RaftApplied` if the process crashed between `save_block_log_id` and the WAL write). OpenRaft detects the gap and calls `apply()` with the missing entries on the next startup.

lib/raft/src/bootstrap.rs
romanbrodetski-ai: Correct — multiple nodes can have `bootstrap = true`. The first one to call `raft.initialize()` succeeds; any subsequent caller gets `NotAllowed` (the cluster is already initialized), which we handle by logging and continuing normally. This is the intended multi-node startup path.

lib/raft/src/network.rs
romanbrodetski-ai: Done — renamed `RaftNetworkFactoryImpl` to `RaftNetworkFactory` and removed the type alias. The openraft trait import is now aliased as `OpenraftNetworkFactory` to avoid the name collision.

lib/raft/src/network.rs
romanbrodetski-ai: Done — removed the clones in `append_entries`, `vote`, and `install_snapshot`. `send_rpc` takes `&self` so `self` is sufficient.

lib/raft/src/network.rs
romanbrodetski-ai: Same fix applied (see previous comment).

.github/workflows/ci.yml
romanbrodetski-ai: Yes — temporary. Reverted to the original (empty for PRs, stress-count 10 for main).

integration-tests/tests/consensus_node/mod.rs
romanbrodetski-ai: Done — moved to `integration-tests/tests/consensus_node/mod.rs` and added `mod consensus_node;` to `suite.rs`. The `#[path]` attribute on `restarted_node_catchup` was also removed since the submodule is now in the same directory.

node/bin/src/lib.rs
romanbrodetski-ai: Good catch. Replaced with `rocksdb::DB::destroy()` (via `zksync_os_rocksdb::rocksdb`) — it removes only files RocksDB tracks so a misconfigured path cannot wipe unrelated data. Followed by `std::fs::remove_dir` on the now-empty directory (which is a no-op if it is not empty, so still safe).

lib/raft/src/network.rs
romanbrodetski-ai: Same fix applied here as well.

lib/status/src/lib.rs
itegulov: Hmm, I expect it to not be that easy from infra perspective... I remember there were some issues with routing HTTP traffic not through LB in our setup.

In any case, not a blocker for merging this PR

itegulov (approved):

PR #1226, opened 2026-04-27
feat(observability): add tokio runtime and per-handler task metrics

LogvinovLeon:
## Summary

**Runtime-level metrics (`lib/observability`)**
- Adds `tokio-metrics` integration, sampling runtime state once per second from within the prometheus exporter task
- Exposes worker thread saturation (\`worker_busy_ratio\`), queue depths (\`global_queue_depth\`, \`total_local_queue_depth\`), and blocking pool usage (\`blocking_queue_depth\`, \`blocking_threads_count\`) as Prometheus gauges under the \`tokio_runtime_*\` prefix
- Enables \`tokio_unstable\` in \`.cargo/config.toml\` (required for queue depth and blocking pool metrics) and propagates it to the \`release-bins\` CI workflow where \`RUSTFLAGS\` is set explicitly and would otherwise override the config file

**Per-handler task metrics (`lib/rpc`)**
- Instruments specific RPC handlers with \`TaskMonitor\` (from \`tokio-metrics\`) via the existing monitoring middleware — no per-handler changes needed
- Exposes three Prometheus gauges per monitored method under the \`rpc_task_*\` prefix:
  - \`rpc_task_mean_scheduled_duration_seconds\` — time a handler task waits in the scheduler queue before a worker thread picks it up; rising values indicate worker saturation
  - \`rpc_task_mean_poll_duration_seconds\` — handler execution time (on-CPU or blocked on RocksDB); together with scheduled duration this decomposes \`api_response_time\` into queue-wait vs. actual work
  - \`rpc_task_slow_polls_count\` — number of polls exceeding the 10µs threshold in the last sample interval
- Sampled once per second from a dedicated background task; accumulates all events in the window so intra-second spikes are not missed
- Monitored methods: \`eth_call\`, \`eth_estimateGas\`, \`eth_feeHistory\`, \`eth_getLogs\`, \`eth_sendRawTransaction\`, \`eth_sendRawTransactionSync\`, \`debug_traceCall\`, \`debug_traceTransaction\`, \`debug_traceBlockByHash\`, \`debug_traceBlockByNumber\`

## Motivation

We observed residual \`eth_sendRawTransaction\` latency spikes visible in Traefik but absent from our own \`api_response_time\` metrics. The hypothesis is worker thread starvation: handlers like \`eth_call\` and \`eth_getLogs\` run synchronous RocksDB reads directly on tokio worker threads, blocking those threads without consuming CPU.

The runtime-level metrics confirm or rule out saturation at the global level. The per-handler task metrics pinpoint which specific method is causing it and whether the latency is from queue wait (starvation) or from the handler itself being slow.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,6 @@
+[build]
+rustflags = ["--cfg", "tokio_unstable"]
+
 [env]
 # Workspace directory to properly resolve artifacts relative paths
 # it will always be relative to the workspace and can be rewritten if required
--- a/.github/workflows/release-bins.yml
+++ b/.github/workflows/release-bins.yml
@@ -79,7 +79,7 @@ jobs:
       - name: Build
         env:
           # Build static binaries
-          RUSTFLAGS: '-C target-feature=+crt-static'
+          RUSTFLAGS: '-C target-feature=+crt-static --cfg tokio_unstable'
         run: |
           cargo build --profile "${RELEASE_PROFILE}" \
             --target "${{ matrix.target }}" --bin "${ZKSYNC_OS_SERVER_BIN}"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14263,7 +14263,7 @@ dependencies = [
 
 [[package]]
 name = "zksync_os_consensus_types"
-version = "0.19.1"
+version = "0.19.2"
 dependencies = [
  "openraft",
  "reth-network-peers",
@@ -14832,6 +14832,7 @@ dependencies = [
  "serde_json",
  "thiserror 2.0.17",
  "tokio",
+ "tokio-metrics",
  "tokio-stream",
  "tower",
  "tower-http",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3902,7 +3902,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8d162beedaa69905488a8da94f5ac3edb4dd4788b732fadb7bd120b2625c1976"
 dependencies = [
  "data-encoding",
- "syn 1.0.109",
+ "syn 2.0.117",
 ]
 
 [[package]]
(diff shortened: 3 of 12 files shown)


RomanBrodetski:
I lioke the idea of this afdditional observability but I want to understand the trade-off. In its current state, will it always be enabled? Are there anby downsides? Will this affect performance? Are there any reprecussions of tokio-unstable?

LogvinovLeon:
**Always enabled:** tied to the prometheus task — if prometheus is disabled (ephemeral mode) the monitor doesn't run. Otherwise yes, always on. No config flag, didn't think one was needed but happy to add one.

**Performance:** reads runtime atomics once per second and sets gauges. Effectively zero overhead.

**`tokio_unstable`:** fair concern. The stable subset (`workers_count`, `busy_ratio`, `global_queue_depth`, `live_tasks_count`) doesn't need the flag. The flag buys us `blocking_queue_depth`, `blocking_threads_count`, and `total_local_queue_depth` — which are specifically the metrics that would confirm or rule out the starvation hypothesis. In practice these APIs haven't changed across Tokio versions and the flag is widely used in production, but it is a project-wide commitment: if Tokio ever changes those APIs the build breaks everywhere, and tools like rust-analyzer need the flag set too (it's in `.cargo/config.toml` so it should be picked up automatically).

If you'd rather not take that dependency, I can drop the unstable metrics and remove the flag — the stable ones are still useful, just less specific to this investigation.

RomanBrodetski:
>  the stable ones are still useful, just less specific to this investigation.

what exactly do you mean?

LogvinovLeon:
The stable metrics tell you workers are occupied but not why — they can't distinguish between a worker doing real async work vs one blocked on a synchronous call.

The two unstable metrics that matter for this investigation:
- `mean_poll_duration` — average time per task poll. Healthy async code yields in microseconds. If `eth_call` is blocking a worker for 50ms, this spikes. Direct evidence of the problem.
- `total_local_queue_depth` — tasks waiting to be picked up. Grows when workers are occupied and can't accept new work. This is what causes the `eth_sendRawTransaction` delay.

After adding `spawn_blocking`, two more become relevant:
- `blocking_queue_depth` / `blocking_threads_count` — how hard the blocking pool is being hit, i.e. whether disk is the bottleneck.

Also — `tokio_unstable` doesn't change any runtime behaviour, only gates API visibility. The only real risk is a compile error if Tokio renames something in a minor version bump.

popzxc (dismissed):
No objections, but I guess if you're interested in checking particular logic, it would be better to instrument concrete flows with [`TaskMonitor`s](https://docs.rs/tokio-metrics/latest/tokio_metrics/struct.TaskMonitor.html), no? 

Probably it can also be exposed as metrics via labels per instrumented task (e.g. a generalized solution is possible).

hatemosphere (approved):
CI change LGTM

PR #1228, opened 2026-04-28
chore(deps): upgrade reth and alloy

itegulov-claude:
## Summary

- rebase the branch onto current `main`
- keep the alloy 2.x upgrade and replay the existing transaction / tracer adaptations on top of the rebased codebase
- switch the workspace `reth` pins to `itegulov/reth@ff8afdc5dbc253019df5b68f4b231f55daeb2e00` (`v2.2.0` plus the RLPx buffer fix)
- adapt `eth_simulateV1` to the current alloy 2 RPC types introduced on `main`

## Testing

- `cargo check --workspace --exclude zksync_os_server --exclude zksync_os_integration_tests -q`
- `cargo check -p zksync_os_server --no-default-features -q`

## Notes

- `cargo check --workspace -q` still hits the environment's missing `make` binary while building `tikv-jemalloc-sys` for the default `zksync_os_server` / integration-test path.


itegulov-claude:
Correction: pushed 45c51579 with a JS tracer test fix for rpc::debug::tracing::debug_trace_call_stack::current_to_l1. New workflow runs are in action_required because this PR comes from a fork, so a maintainer needs to approve the workflows before they will execute.

lib/network/src/service.rs
```diff
@@ -201,7 +201,12 @@ impl NetworkService {
         let boot_nodes = resolve_boot_nodes_with_retry(config.boot_nodes.clone()).await?;
         tracing::info!(?genesis, ?fork_id, "initializing p2p network service");
         let (protocol_tx, protocol_rx) = mpsc::unbounded_channel();
-        let cfg_builder = RethNetworkConfig::builder(config.secret_key)
+        let runtime = RuntimeBuilder::new(RuntimeConfig::default().with_tokio(
+            TokioConfig::existing_handle(tokio::runtime::Handle::current()),
+        ))
+        .build()
+        .map_err(|err| io::Error::other(format!("failed to initialize reth runtime: {err}")))?;
```
itegulov: We should pass runtime from `node/bin/src/lib.rs` instead

itegulov-claude: Fixed in 9d20ad5c. `NetworkService::new` now takes the existing reth runtime from `node/bin`, and both call sites pass `runtime.clone()` through instead of constructing a new runtime inside `lib/network`.

RomanBrodetski (dismissed):

itegulov (approved):

PR #1235, opened 2026-04-29
feat: track finalized executed batches and blocks

perekopskiy:
## Summary

Adds BlockBoundary enum, that is passed to `L1Watcher::new`, most of the watchers use BlockBoundary::Confirmed, but those that need finalized events use BlockBoundary::Finalized. Two versions of `L1ExecuteWatcher`s are run, one with Confirmed and one with Finalized, they set the corresponding finality fields. 
`last_finalized_executed` is used in RPC for resolving "finalized" block tag and in priority tree persistence.
Also, L1PersistBatchWatcher now runs in `BlockBoundary::Finalized` mode.

For persistance use-cases we use "finalized" so that we're sure we don't have to revert storages in some exceptional cases. For RPC we use `last_finalized_executed` because only for these blocks/batches true finalization is guaranteed + this way L2 to L1 log proofs are always available for finalized blocks.

### DIFF
--- a/docs/src/setup/local_run.md
+++ b/docs/src/setup/local_run.md
@@ -23,7 +23,7 @@ To run node locally, first decompress state and launch `anvil`:
 
 ```
 gzip -dfk ./local-chains/v30.2/l1-state.json.gz
-anvil --load-state ./local-chains/v30.2/l1-state.json --port 8545 --block-time 0.25 --mixed-mining
+anvil --load-state ./local-chains/v30.2/l1-state.json --port 8545 --block-time 0.25 --mixed-mining --slots-in-an-epoch 10
 ```
 
 then launch the server:
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -1200,6 +1200,8 @@ impl AnvilL1 {
         let locked_port = LockedPort::acquire_unused().await?;
         let address = format!("http://localhost:{}", locked_port.port);
 
+        // --slots-in-an-epoch defines what blocks are "finalized" in Anvil, last finalized block is `latest - 2 * slots_in_an_epoch`
+        // so we set block time to 0.25s and slots in epoch set to 10 and finalization delays is about 10*0.25s*2=5s which is reasonable for tests.
         let provider = ProviderBuilder::new().connect_anvil_with_wallet_and_config(|anvil| {
             anvil
                 .port(locked_port.port)
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -1209,6 +1209,8 @@ impl AnvilL1 {
                 .arg("--mixed-mining")
                 .arg("--load-state")
                 .arg(l1_state_path)
+                .arg("--slots-in-an-epoch")
+                .arg("10")
         })?;
 
         let wallet = provider.wallet().clone();
--- a/lib/contract_interface/src/l1_discovery.rs
+++ b/lib/contract_interface/src/l1_discovery.rs
@@ -289,6 +289,14 @@ async fn fetch_finalized_executed_batch(
         .context("failed to fetch finalized SL block number")?
         .context("failed to fetch finalized SL block number (`None` returned)")?;
 
+    if !zk_chain_sl
+        .code_exists_at_block(finalized_sl_block_number.into())
+        .await
+        .context("failed to check ZK chain contract code at finalized SL block")?
+    {
+        return Ok((finalized_sl_block_number, 0));
+    }
+
     let last_finalized_executed_batch = zk_chain_sl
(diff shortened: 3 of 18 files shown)


docs/src/setup/local_run.md
```diff
@@ -23,7 +23,7 @@ To run node locally, first decompress state and launch `anvil`:
 
 ```
 gzip -dfk ./local-chains/v30.2/l1-state.json.gz
-anvil --load-state ./local-chains/v30.2/l1-state.json --port 8545 --block-time 0.25 --mixed-mining
+anvil --load-state ./local-chains/v30.2/l1-state.json --port 8545 --block-time 0.25 --mixed-mining --slots-in-an-epoch 10
```
itegulov, line 26: This is slowly getting out of hand 😅

lib/contract_interface/src/l1_discovery.rs
```diff
@@ -268,6 +279,35 @@ impl L1State {
...
+    zk_chain_sl: &ZkChain<DynProvider>,
+) -> anyhow::Result<(u64, u64)> {
+    let finalized_sl_block_number = {
+        let Some(finalized_sl_block_number) = zk_chain_sl
+            .provider()
+            .get_block_number_by_id(BlockId::finalized())
+            .await
+            .context("failed to fetch finalized SL block number")?
+        else {
+            return Ok((0, 0));
```
itegulov: Do you need this fallback?

lib/contract_interface/src/l1_discovery.rs
```diff
@@ -268,6 +279,35 @@ impl L1State {
...
+        };
+        finalized_sl_block_number
```
itegulov: Nit: looks like you don't need this outer {} code block at all

lib/contract_interface/src/l1_discovery.rs
```diff
@@ -268,6 +279,35 @@ impl L1State {
...
+    };
+
+    if !zk_chain_sl
+        .code_exists_at_block(finalized_sl_block_number.into())
+        .await
+        .context("failed to check ZK chain contract code at finalized SL block")?
+    {
+        return Ok((finalized_sl_block_number, 0));
```
itegulov: Hmm, is this for cases where we deploy L1 contracts and then start L2 node immediately after?

lib/l1_watcher/src/execute_watcher.rs
```diff
@@ -71,6 +79,46 @@ impl<Finality: WriteFinality> L1ExecuteWatcher<Finality> {
 
         Ok(l1_watcher)
     }
+
+    pub async fn create_finalized_watcher(
```
itegulov, line 87: Hmm, maybe these should be two different watchers? Otherwise they'll share the same NAME which might be confusing in logs. There is not a lot of shared meaningful code in this file anyway.

integration-tests/src/lib.rs
```diff
@@ -1209,6 +1209,8 @@ impl AnvilL1 {
                 .arg("--mixed-mining")
                 .arg("--load-state")
                 .arg(l1_state_path)
+                .arg("--slots-in-an-epoch")
+                .arg("10")
```
itegulov, line 1215: I was hoping to find a comment why this is needed in one of the places where you added this but didn't. So -- why?

itegulov (commented):
LGTM overall, left a few comments

lib/contract_interface/src/l1_discovery.rs
perekopskiy: no, removed

lib/contract_interface/src/l1_discovery.rs
perekopskiy: yeah, on second thought, we shouldn't worry about it, removed this code

lib/l1_watcher/src/execute_watcher.rs
perekopskiy: split them

integration-tests/src/lib.rs
perekopskiy: Added comments, it's needed to speed up finalization so we wait less in tests that require finalization

perekopskiy: >This is slowly getting out of hand 😅

Agree, I'm not 100% sure we need it, default value is 32, which implies, 16 seconds finalization delay which is still okayish for tests/localhost

lib/contract_interface/src/l1_discovery.rs
perekopskiy: ok, it's needed for localhost, because on anvil startup finalized state doesn't have L1 contracts

itegulov (approved):

antonbaliasnikov (approved):

PR #1236, opened 2026-04-29
feat(l1-watcher): add SL-aware watcher

itegulov:
## Summary

Adds `SlAwareWatcher` that can traverse SL intervals from different chains (L1->GW->L1->...) and makes `L1PersistBatchWatcher` use it.


lib/l1_watcher/src/watcher.rs
```diff
@@ -1,69 +1,102 @@
...
 impl L1Watcher {
+    /// Polls for new events.
+    ///
+    /// For unbounded watchers (`end_block = None`) this never returns; for bounded watchers
+    /// it returns once the cursor passes `end_block`.
     pub async fn run(mut self) {
+        self.run_inner().await;
+    }
+
+    pub(crate) async fn run_inner(&mut self) {
```
perekopskiy, line 92: maybe add a comment about why `run_inner` is needed? IIUC correctly we need a method that doesn't consume `self` so that we can keep `processor`

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -23,68 +23,98 @@ use zksync_os_storage_api::{PersistedBatch, WriteBatch};
...
+            let first_batch = if segments.is_empty() {
+                if last_persisted_batch == 0 {
+                    0
+                } else {
+                    last_persisted_batch.max(interval.first_batch)
+                }
+            } else {
+                interval.first_batch
+            };
+            segments.push(SegmentSpec {
```
perekopskiy, line 89: nit: would also be nice to validate that we add first segment with `first_batch == last_persisted_batch`

perekopskiy (approved):

PR #1239, opened 2026-04-30
feat: add possibility to forcefully resubmit transactions

Artemka374:
## Summary

With current setup, we always wait for in-flight transactions indefinitely. This PR introduces an "emergency button" in case the inclusion is taking too long - so it is possible to force-replace transactions. This should not be enabled as default and should be used only if very needed - and disabled right after to avoid draining money


node/bin/src/config/mod.rs
```diff
@@ -767,6 +767,22 @@ pub struct L1SenderConfig {
     #[config(default_t = 2 * EtherUnit::Gwei)]
     pub max_fee_per_blob_gas: EtherAmount,
 
+    /// Multiplier applied to `max_fee_per_gas` when `force_transaction_resubmission` is enabled.
+    #[config(default_t = 1.0)]
+    pub max_fee_per_gas_replacement_multiplier: f64,
```
slowli: Nit: Could make sense to [validate](https://matter-labs.github.io/smart-config/smart_config/validation/index.html) these values – at least that they are positive (probably >= 1.0?).

node/bin/src/config/mod.rs
```diff
@@ -767,6 +767,22 @@ pub struct L1SenderConfig {
     #[config(default_t = 2 * EtherUnit::Gwei)]
     pub max_fee_per_blob_gas: EtherAmount,
 
+    /// Multiplier applied to `max_fee_per_gas` when `force_transaction_resubmission` is enabled.
```
slowli: Bikeshedding: If multipliers are dependent on `force_transaction_resubmission`, it could make sense to structure them underneath it:

```yaml
force_transaction_resubmission:
  enabled: true
  max_fee_per_gas_replacement_multiplier: 1.25
  # ...
```

node/bin/src/config/mod.rs
Artemka374: I think it's fine that it can be less than 1.0 in some cases as we compare to the cap

node/bin/src/config/mod.rs
```diff
@@ -807,6 +811,29 @@ pub struct L1SenderConfig {
...
+#[derive(Clone, Copy, Debug, DescribeConfig, DeserializeConfig, ConfigValidate)]
+#[config(derive(Default))]
+pub struct ForceTransactionResubmissionConfig {
+    /// Skips startup in-flight recovery and resubmits queued L1 transactions with replacement fee caps.
+    #[config(default_t = false)]
+    pub enabled: bool,
+
+    /// Multiplier applied to `max_fee_per_gas` when force transaction resubmission is enabled.
+    #[config(default_t = 1.0)]
+    #[config_validate(custom(|_: &Config, value: &f64| value.is_sign_positive(), "must be positive"))]
```
slowli: You can use a `config` attribute here and below:

```rust
pub struct ForceTransactionResubmissionConfig {
    #[config(default_t = 1.0, validate(is_positive_f64, "must be positive"))]
    pub max_fee_per_gas_replacement_multiplier: f64,
    // ...
}

fn is_positive_f64(&val: &f64) -> bool {
    val > 0.0 // I don't think `is_sign_positive()` is right; it's true for +0.0 and some NaNs
}
```

The benefit is that this would show up in `config help` docs and would provide more comprehensive error messages.

slowli (commented):
Looks good from the config perspective (other than validation; see the line comment). The changes to the main logic look fine too, but I'm not that familiar with the domain so I'd suggest to seek another review.

node/bin/src/config/mod.rs
```diff
@@ -807,6 +811,29 @@ pub struct L1SenderConfig {
...
 
+#[derive(Clone, Copy, Debug, DescribeConfig, DeserializeConfig, ConfigValidate)]
+#[config(derive(Default))]
+pub struct ForceTransactionResubmissionConfig {
+    /// Skips startup in-flight recovery and resubmits queued L1 transactions with replacement fee caps.
+    #[config(default_t = false)]
+    pub enabled: bool,
+
+    /// Multiplier applied to `max_fee_per_gas` when force transaction resubmission is enabled.
+    #[config(default_t = 1.0)]
```
itegulov: Should we set these to something reasonable by default? i.e. whatever reth expects in each field when trying to replace an existing tx

itegulov (dismissed):
LGTM, let's document the process somewhere (on-call handbook?)

node/bin/src/config/mod.rs
Artemka374: now it makes more sense as we use constant max_fee_per_gas, I'll change this

node/bin/src/config/mod.rs
Artemka374: I doubt that it's very possible to set +0.0 or NaN in config(otherwise I'd expect it to panic?), but we can do this way

itegulov (approved):

PR #1240, opened 2026-04-30
feat(rpc): Implement eth_simulateV1

Artemka374:
## Summary                                                                                                                                                                   
                                                                                                                                                                               
  Implements the `eth_simulateV1` RPC method ([spec](https://github.com/ethereum/execution-apis/pull/484)).
                                                                                                                                                                               
  Executes the requested blocks linearly against a base state, carrying an overlay of storage writes from each simulated block forward into subsequent ones. State overrides   
  per block are seeded into that overlay before execution, with actual VM writes taking precedence. When `validation=false`, the base fee is zeroed for execution while the    
  returned block header still reflects the requested/overridden fee.                                                                                                           
                                                                     
  **Not supported:**                                

  - `traceTransfers=true` — rejected with an error. ZKsync OS VM does not expose a transfer stream equivalent to reth's `TransferInspector`, so native value moves from        
  internal `CALL`/`CREATE`/`SELFDESTRUCT` cannot be captured.
  - `movePrecompileToAddress` (state override) — rejected with an error. The execution backend has no mechanism to remap the precompile table for a single simulated block.    
  - `difficulty` (block override) — silently ignored. ZKsync OS uses `mix_hash` for `PREVRANDAO`; the `random` block override maps to that correctly.                          
  - `beacon_root` (block override) — silently ignored. There is no corresponding field in `BlockContext`
  - `validation=false` - works partially since VM can still check explicitly assigned nonces and fail if they are stale, so nonce checks are still present in some way

### DIFF
--- a/integration-tests/tests/rpc/simulate.rs
+++ b/integration-tests/tests/rpc/simulate.rs
@@ -2,6 +2,8 @@ use alloy::primitives::U256;
 use alloy::providers::Provider;
 use alloy::rpc::types::TransactionRequest;
 use alloy::rpc::types::simulate::{SimBlock, SimulatePayload};
+use alloy::sol_types::{SolCall, SolEvent};
+use zksync_os_integration_tests::contracts::EventEmitter::TestEvent;
 use zksync_os_integration_tests::contracts::{Counter, EventEmitter};
 use zksync_os_integration_tests::{CURRENT_TO_L1, Tester, test_multisetup};
 
@@ -69,41 +71,63 @@ async fn simulate_multiple_txs_in_one_block(tester: Tester) -> anyhow::Result<()
         assert!(call.gas_used > 0, "call {i} gas used should be non-zero");
     }
 
+    // First two calls must have emitted exactly one TestEvent each, with numbers 1 and 2 in
+    // that order. This proves that logs (and their ordering) are reported per-call.
+    for (i, expected_number) in [1u64, 2u64].iter().enumerate() {
+        let logs = &block.calls[i].logs;
+        assert_eq!(logs.len(), 1, "call {i} should have one log");
+        let decoded = TestEvent::decode_log(&logs[0].inner)
+            .unwrap_or_else(|e| panic!("call {i} log is not a TestEvent: {e}"));
+        assert_eq!(decoded.number, U256::from(*expected_number));
+    }
+    // The plain ETH transfer should not have produced any logs.
+    assert!(
+        block.calls[2].logs.is_empty(),
+        "transfer should emit no logs"
+    );
+
     Ok(())
 }
 
-/// Simulate two blocks in sequence and verify that state changes from the first block carry over.
+/// Simulate three blocks in sequence and verify that state changes from earlier blocks carry over.
 #[test_multisetup([CURRENT_TO_L1])]
 async fn simulate_state_carries_across_blocks(tester: Tester) -> anyhow::Result<()> {
-    // Deploy a Counter and increment it in block 1. In block 2, read it via eth_call to confirm
-    // the simulated write is visible.
+    // Deploy a Counter, increment by 7 in block 1, by 3 in block 2, then read the counter in
+    // block 3 and confirm it equals 10. Without state carry-over, the read would return 0.
     let counter = Counter::deploy(tester.l2_provider.clone()).await?;
 
     let increment_call = counter.increment(U256::from(7)).into_transaction_request();
-    // A second increment in block 2 — if state did not carry over, this would fail or return wrong data.
     let increment_call_2 = counter.increment(U256::from(3)).into_transaction_request();
+    let read_call = counter.counter().into_transaction_request();
 
     let payload = SimulatePayload {
         block_state_calls: vec![
             SimBlock::default().call(increment_call),
             SimBlock::default().call(increment_call_2),
+            SimBlock::default().call(read_call),
         ],
         ..Default::default()
     };
 
     let results = tester.l2_provider.simulate(&payload).await?;
 
-    assert_eq!(results.len(), 2, "expected two simulated blocks");
+    assert_eq!(results.len(), 3, "expected three simulated blocks");
     assert_eq!(results[0].calls.len(), 1);
     assert_eq!(results[1].calls.len(), 1);
+    assert_eq!(results[2].calls.len(), 1);
 
     let first_call = &results[0].calls[0];
     let second_call = &results[1].calls[0];
+    let read_result = &results[2].calls[0];
 
     assert!(first_call.status, "first increment should succeed");
     assert!(second_call.status, "second increment should succeed");
+    assert!(read_result.status, "read should succeed");
     assert!(first_call.gas_used > 0);
     assert!(second_call.gas_used > 0);
+
+    let observed = Counter::counterCall::abi_decode_returns(&read_result.return_data)?;
+    assert_eq!(observed, U256::from(10), "counter should be 7 + 3 = 10");
 
     Ok(())
 }
--- a/integration-tests/tests/rpc/simulate.rs
+++ b/integration-tests/tests/rpc/simulate.rs
@@ -1,0 +1,109 @@
+use alloy::primitives::U256;
+use alloy::providers::Provider;
+use alloy::rpc::types::TransactionRequest;
+use alloy::rpc::types::simulate::{SimBlock, SimulatePayload};
+use zksync_os_integration_tests::contracts::{Counter, EventEmitter};
+use zksync_os_integration_tests::{CURRENT_TO_L1, Tester, test_multisetup};
+
+/// Simulate a single ETH transfer in one block and verify that gas was consumed.
+#[test_multisetup([CURRENT_TO_L1])]
+async fn simulate_eth_transfer_gas_used(tester: Tester) -> anyhow::Result<()> {
+    let sender = tester.l2_wallet.default_signer().address();
+    let recipient = alloy::primitives::address!("000000000000000000000000000000000000dead");
+    let value = U256::from(1_000u64);
+
+    let payload = SimulatePayload {
+        block_state_calls: vec![
+            SimBlock::default().call(
+                TransactionRequest::default()
+                    .from(sender)
+                    .to(recipient)
+                    .value(value),
+            ),
+        ],
+        ..Default::default()
+    };
+
+    let results = tester.l2_provider.simulate(&payload).await?;
+
+    assert_eq!(results.len(), 1, "expected one simulated block");
+    let block = &results[0];
+    assert_eq!(block.calls.len(), 1, "expected one call result");
+
+    let call = &block.calls[0];
+    assert!(call.status, "transfer should succeed");
+    assert!(call.gas_used > 0, "gas used should be non-zero");
+
+    Ok(())
+}
+
+/// Simulate three transactions in the same block and verify all succeed with non-zero gas.
+#[test_multisetup([CURRENT_TO_L1])]
+async fn simulate_multiple_txs_in_one_block(tester: Tester) -> anyhow::Result<()> {
+    let sender = tester.l2_wallet.default_signer().address();
+    let emitter = EventEmitter::deploy(tester.l2_provider.clone()).await?;
+    let recipient = alloy::primitives::address!("000000000000000000000000000000000000dead");
+
+    let payload = SimulatePayload {
+        block_state_calls: vec![
+            SimBlock::default()
+                .call(emitter.emitEvent(U256::from(1)).into_transaction_request())
+                .call(emitter.emitEvent(U256::from(2)).into_transaction_request())
+                .call(
+                    TransactionRequest::default()
+                        .from(sender)
+                        .to(recipient)
+                        .value(U256::from(1u64)),
+                ),
+        ],
+        ..Default::default()
+    };
+
+    let results = tester.l2_provider.simulate(&payload).await?;
+
+    assert_eq!(results.len(), 1);
+    let block = &results[0];
+    assert_eq!(block.calls.len(), 3, "expected three call results");
+    for (i, call) in block.calls.iter().enumerate() {
+        assert!(call.status, "call {i} should succeed");
+        assert!(call.gas_used > 0, "call {i} gas used should be non-zero");
+    }
+
+    Ok(())
+}
+
+/// Simulate two blocks in sequence and verify that state changes from the first block carry over.
+#[test_multisetup([CURRENT_TO_L1])]
+async fn simulate_state_carries_across_blocks(tester: Tester) -> anyhow::Result<()> {
+    // Deploy a Counter and increment it in block 1. In block 2, read it via eth_call to confirm
+    // the simulated write is visible.
+    let counter = Counter::deploy(tester.l2_provider.clone()).await?;
+
+    let increment_call = counter.increment(U256::from(7)).into_transaction_request();
+    // A second increment in block 2 — if state did not carry over, this would fail or return wrong data.
+    let increment_call_2 = counter.increment(U256::from(3)).into_transaction_request();
+
+    let payload = SimulatePayload {
+        block_state_calls: vec![
+            SimBlock::default().call(increment_call),
+            SimBlock::default().call(increment_call_2),
+        ],
+        ..Default::default()
+    };
+
+    let results = tester.l2_provider.simulate(&payload).await?;
+
+    assert_eq!(results.len(), 2, "expected two simulated blocks");
+    assert_eq!(results[0].calls.len(), 1);
+    assert_eq!(results[1].calls.len(), 1);
+
+    let first_call = &results[0].calls[0];
+    let second_call = &results[1].calls[0];
+
+    assert!(first_call.status, "first increment should succeed");
+    assert!(second_call.status, "second increment should succeed");
+    assert!(first_call.gas_used > 0);
+    assert!(second_call.gas_used > 0);
+
+    Ok(())
+}
--- a/lib/rpc/src/eth_call_handler.rs
+++ b/lib/rpc/src/eth_call_handler.rs
@@ -19,11 +19,14 @@ use std::sync::Arc;
 use std::time::{SystemTime, UNIX_EPOCH};
 use tokio::sync::watch;
 use zk_os_api::helpers::{get_balance, get_nonce};
+use zksync_os_interface::tracing::{NopTracer, NopValidator};
+use zksync_os_interface::traits::{NoopTxCallback, TxListSource};
 use zksync_os_interface::types::{BlockHashes, ExecutionOutput};
 use zksync_os_interface::{
     error::InvalidTransaction,
     types::{BlockContext, ExecutionResult},
 };
+use zksync_os_multivm::run_block;
 use zksync_os_rpc_api::types::ZkApiBlock;
 use zksync_os_storage_api::{
     BlockOverlay, RepositoryError, StateError, ViewState,
@@ -771,15 +774,14 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
 
         for sim_block in block_state_calls {
             block_context.mix_hash = U256::ZERO;
-            let header_overrides = match sim_block.block_overrides {
-                Some(block_overrides) => simulation_utils::apply_simulate_block_overrides(
+            if let Some(block_overrides) = sim_block.block_overrides {
+                simulation_utils::apply_simulate_block_overrides(
                     &mut block_context,
                     block_overrides,
                     previous_block_number,
                     previous_timestamp,
-                )?,
-                None => simulation_utils::SimulateHeaderOverrides::default(),
-            };
+                )?;
+            }
 
             // `validation=false` disables fee validation for execution, but the returned block
             // still reports the requested/header-overridden base fee.
@@ -806,12 +808,22 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
                 execution_context,
                 overridden_view.clone(),
             )?;
-            let block_output =
-                simulation_utils::run_simulation_block(execution_context, overridden_view, &txs)?;
+            let tx_source = TxListSource {
+                transactions: txs.iter().cloned().map(|tx| tx.encode()).collect(),
+            };
+            let block_output = run_block(
+                execution_context,
+                overridden_view.clone(),
+                overridden_view,
+                tx_source,
+                NoopTxCallback,
+                &mut NopTracer,
+                &mut NopValidator,
+            )
+            .map_err(EthCallError::ForwardSubsystemError)?;
 
             let simulated_block = simulation_utils::build_simulated_block_response(
                 response_context,
-                header_overrides,
                 txs,
                 block_output,
                 return_full_transactions,
--- a/lib/rpc/src/eth_call_handler.rs
+++ b/lib/rpc/src/eth_call_handler.rs
@@ -827,7 +827,7 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
             )
             .map_err(EthCallError::ForwardSubsystemError)?;
 
-            let simulated_block = simulation_utils::build_simulated_block_response(
+            let (simulated_block, block_overlay) = simulation_utils::build_simulated_block_response(
                 response_context,
                 txs,
                 block_output,
@@ -838,7 +838,7 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
             let mut overlay = state_overrides;
             // State overrides seed this simulated block; actual VM writes take precedence for
             // subsequent simulated blocks.
-            overlay.extend(simulated_block.overlay);
+            overlay.extend(block_overlay);
             Arc::get_mut(&mut overlays)
                 .ok_or_else(|| {
                     EthCallError::ForwardSubsystemError(anyhow::anyhow!(
@@ -848,10 +848,7 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
                 .insert(block_context.block_number, overlay);
             previous_block_number = block_context.block_number;
             previous_timestamp = block_context.timestamp;
-            simulated_blocks.push(SimulatedBlock {
-                inner: simulated_block.inner,
-                calls: simulated_block.calls,
-            });
+            simulated_blocks.push(simulated_block);
 
             block_context = simulation_utils::next_block_context(block_context, next_hash);
         }
--- a/lib/rpc/src/eth_call_handler.rs
+++ b/lib/rpc/src/eth_call_handler.rs
@@ -727,7 +727,7 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
     /// Spec limitations: this backend does not expose reth's `TransferInspector` equivalent, so
     /// `traceTransfers=true` is rejected instead of returning partial transfer logs. The backend
     /// also cannot remap precompiles for a single simulated block, so `movePrecompileToAddress` is
-    /// validated for self/duplicate references and then rejected as unsupported.
+    /// rejected as unsupported.
     pub fn simulate_v1_impl(
         &self,
         opts: SimulatePayload,
@@ -796,7 +796,12 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
 
             let state_overrides = match sim_block.state_overrides {
                 Some(state_overrides) => {
-                    simulation_utils::validate_state_overrides_for_simulate(&state_overrides)?;
+                    if state_overrides
+                        .values()
+                        .any(|account| account.move_precompile_to.is_some())
+                    {
+                        return Err(EthCallError::SimulateMovePrecompileNotSupported);
+                    }
                     build_state_override_maps(&simulation_view, state_overrides)
                 }
                 None => BlockOverlay::default(),
@@ -965,10 +970,6 @@ pub enum EthCallError {
     SimulateBlockTimestampInvalid { got: u64, parent: u64 },
     #[error("block gas limit exceeded by the block's transactions")]
     SimulateBlockGasLimitExceeded,
-    #[error("MovePrecompileToAddress referenced itself")]
-    SimulatePrecompileSelfReference,
-    #[error("multiple MovePrecompileToAddress values reference the same replacement address")]
-    SimulatePrecompileDuplicateAddress,
     #[error("movePrecompileToAddress is not supported by this execution backend")]
     SimulateMovePrecompileNotSupported,
     // todo(EIP-4844)
--- a/lib/rpc/src/eth_call_handler.rs
+++ b/lib/rpc/src/eth_call_handler.rs
@@ -779,6 +779,8 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
         let mut previous_timestamp = parent_timestamp;
 
         for sim_block in block_state_calls {
+            // Per the eth_simulateV1 spec, prevrandao defaults to zero for simulated blocks
+            // unless explicitly overridden via `blockOverrides.random`.
             block_context.mix_hash = U256::ZERO;
             if let Some(block_overrides) = sim_block.block_overrides {
                 simulation_utils::apply_simulate_block_overrides(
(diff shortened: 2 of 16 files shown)


lib/rpc/src/eth_call_handler.rs
```diff
@@ -420,6 +432,136 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+    /// Spec limitations: this backend does not expose reth's `TransferInspector` equivalent, so
+    /// `traceTransfers=true` is rejected instead of returning partial transfer logs. The backend
+    /// also cannot remap precompiles for a single simulated block, so `movePrecompileToAddress` is
+    /// validated for self/duplicate references and then rejected as unsupported.
+    pub fn simulate_v1_impl(
+        &self,
+        opts: SimulatePayload,
+        block: Option<BlockId>,
+    ) -> Result<Vec<SimulatedBlock<ZkApiBlock>>, EthCallError> {
+        if opts.block_state_calls.is_empty() {
```
slowli: Nit: Why not destructure `SimulatePayload` at the start, so that its fields can be accessed more directly in this and the following check?

lib/rpc/src/eth_call_handler.rs
```diff
@@ -420,6 +432,136 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+            // `validation=false` disables fee validation for execution, but the returned block
+            // still reports the requested/header-overridden base fee.
+            let response_context = block_context;
+            let mut execution_context = response_context;
+            if !validation {
+                execution_context.eip1559_basefee = U256::ZERO;
+            }
+
+            let simulation_view =
+                OverriddenStateView::new(base_state.clone(), Arc::clone(&overlays));
```
slowli: Nit: It'd be quite easy to implement `OverrideProvider` for `BTreeMap` directly, e.g., by refactoring the implementation for `Arc<_>` as a generic delegating impl. This would make `Arc` wrappers unnecessary.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -420,6 +432,136 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+            let next_hash = simulated_block.inner.header.hash;
+
+            let mut overlay = state_override_overlay;
+            // State overrides seed this simulated block; actual VM writes take precedence for
+            // subsequent simulated blocks.
+            overlay
+                .storage_writes
+                .extend(simulated_block.overlay.storage_writes);
+            overlay.preimages.extend(simulated_block.overlay.preimages);
+            Arc::make_mut(&mut overlays).insert(block_context.block_number, overlay);
```
slowli: `make_mut()` only works here as expected because there's a single strong `Arc` reference. To make this more explicit, it'd be better to use `Arc::get_mut()`.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -447,6 +589,103 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+                block_context: self.build_pending_block_context(),
+                base_state_block: parent_block,
+                parent_block_number: parent_block,
+                parent_timestamp: parent_context.timestamp,
+            });
+        }
+
+        let Some(parent_block) = self.storage.resolve_block_number(block)? else {
+            return Err(RpcStorageError::BlockNotFound(block).into());
+        };
```
slowli: Nit: It'd be possible to use `ok_or` here in the same way it's used below.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -447,6 +589,103 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+                    });
+                    request.set_nonce(nonce);
+                    let next_nonce =
+                        nonce
+                            .checked_add(1)
+                            .ok_or(EthCallError::SimulateInvalidParams(
+                                "nonce has max value".to_string(),
+                            ))?;
+                    next_nonces.insert(from, next_nonce);
+                } else if let Some(nonce) = request.nonce {
```
slowli: Bikeshedding: This and the preceding `if` condition are structured weirdly. Unless I'm missing something, they can be refactored as

```rust
if let Some(nonce) = request.nonce {
    // ...
} else {
    // ...
}
```

...which IMO would be much more understandable.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -447,6 +589,103 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+                    let state_nonce = next_nonces.entry(from).or_insert_with(|| {
+                        state_view
+                            .get_account(from)
+                            .as_ref()
+                            .map(get_nonce)
+                            .unwrap_or_default()
+                    });
```
slowli: Nit: Since `state_nonce` is used in both `if` and `else` blocks, it could be moved above `if` to avoid code duplication. (Also, it'd be possible to assign `*state_nonce = next_nonce` in the `if` block instead of `insert`ing into the map.)

lib/rpc/src/eth_call_handler.rs
```diff
@@ -447,6 +589,103 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+                    if nonce >= *state_nonce {
+                        *state_nonce = nonce.saturating_add(1);
```
slowli: Dumb question: Why is saturation OK here, but not above when computing `next_nonce`?

lib/rpc/src/eth_call_handler.rs
```diff
@@ -447,6 +589,103 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+
+        Ok(SimulationStartContext {
+            block_context,
+            base_state_block: parent_block,
+            parent_block_number: parent_block,
+            parent_timestamp: parent_context.timestamp,
+        })
+    }
+
+    fn create_simulation_txs<V: ViewState + Clone>(
```
slowli: Nit: The `Clone` requirement is unnecessary.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -447,6 +589,103 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+    ) -> Result<Vec<ZkTransaction>, EthCallError> {
+        let default_gas_limit = simulation_default_gas_limit(
+            &calls,
+            block_context.gas_limit,
+            self.config.eth_call_gas as u64,
+        )?;
+        let mut next_nonces = HashMap::new();
+
+        calls
+            .iter_mut()
```
slowli: Nit: It's possible to use `into_iter()` here instead (`calls` are passed by value as-is). This would make cloning a `request` for the `create_tx_from_request()` call unnecessary.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -662,6 +901,450 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+    header_overrides: SimulateHeaderOverrides,
+    txs: Vec<ZkTransaction>,
+    block_output: BlockOutput,
+    return_full_transactions: bool,
+) -> Result<SimulatedBlockResponse, EthCallError> {
+    let BlockOutput {
+        header: sealed_header,
+        tx_results,
+        storage_writes,
+        account_diffs: _,
```
slowli: Bikeshedding: It's possible to collect all unused fields to a single `..` at the end of the destructure, which IMO would be more readable (mentioning all fields *could* make sense if the goal is to ensure that their list isn't changed, which AFAIU isn't the goal here).

lib/rpc/src/eth_call_handler.rs
```diff
@@ -662,6 +901,450 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+    } = block_output;
+
+    let mut block_bloom = Bloom::default();
+    let mut number_of_logs_before_this_tx = 0;
+    let mut cumulative_gas_used = 0;
+    let mut receipts = Vec::with_capacity(tx_results.len());
+    let mut simulated_txs = Vec::with_capacity(tx_results.len());
+    let mut executed_tx_index = 0;
+
+    for (call_index, (tx, result)) in txs.into_iter().zip(tx_results.into_iter()).enumerate() {
```
slowli: Nit: `.zip(tx_results)` is equivalent.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -662,6 +901,450 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+
+    let mut header = sealed_header.unseal();
+    header.base_fee_per_gas = Some(block_context.eip1559_basefee.saturating_to());
+    header.logs_bloom = block_bloom;
+    header.gas_used = cumulative_gas_used;
+    header.transactions_root = calculate_transaction_root(
+        &simulated_txs
+            .iter()
+            .filter(|tx| tx.is_executed())
+            .map(|tx| tx.tx.envelope().clone())
```
slowli: Nit: No need to `clone()` envelopes; `Encodable2718` trait required by `calculate_transaction_root()` is auto-implemented for references.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -662,6 +901,450 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+    }
+
+    fn to_api_tx(
+        &self,
+        block_hash: B256,
+        block_context: BlockContext,
+    ) -> zksync_os_rpc_api::types::ZkApiTransaction {
+        build_api_tx(
+            self.tx.clone(),
+            Some(&self.tx_meta(block_hash, block_context, 0)),
```
slowli: Dumb question: Why is `gas_used == 0` here, as opposed to using a "real" value from `self.result` as in `api_logs()`?

lib/rpc/src/eth_call_handler.rs
```diff
@@ -447,6 +589,103 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+            let parent_block = self.storage.replay_storage().latest_record();
+            let parent_context = self
+                .storage
+                .replay_storage()
+                .get_context(parent_block)
+                .ok_or(RpcStorageError::BlockNotFound(block))?;
+            return Ok(SimulationStartContext {
+                block_context: self.build_pending_block_context(),
+                base_state_block: parent_block,
+                parent_block_number: parent_block,
```
slowli: AFAICT, `base_state_block` and `parent_block_number` are always assigned to the same value. Is there a particular reason to keep them as 2 distinct fields?

lib/rpc/src/eth_call_handler.rs
```diff
@@ -420,6 +432,136 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+                    validate_state_overrides_for_simulate(&state_overrides)?;
+                    let state_overrides =
+                        build_state_override_maps(&simulation_view, state_overrides);
+                    let (storage_writes, preimages) = state_overrides.clone().into_parts();
+                    (
+                        BlockOverlay {
+                            storage_writes,
+                            preimages,
+                        },
+                        state_overrides,
```
slowli: Off-topic (?): It looks weird that `BlockOverlay` and `OwnedOverrides` are identical (both in terms of contents and semantically, AFAICT). Maybe, it would make sense to reduce them to a single type?

lib/rpc/src/eth_call_handler.rs
```diff
@@ -420,6 +432,136 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+            )?;
+            let next_hash = simulated_block.inner.header.hash;
+
+            let mut overlay = state_override_overlay;
+            // State overrides seed this simulated block; actual VM writes take precedence for
+            // subsequent simulated blocks.
+            overlay
+                .storage_writes
+                .extend(simulated_block.overlay.storage_writes);
+            overlay.preimages.extend(simulated_block.overlay.preimages);
```
slowli: Nit: Could make sense to add the corresponding method to `BlockOverlay`: `fn extend(&mut self, changes: Self)`.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -420,6 +432,136 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+        if opts.block_state_calls.is_empty() {
+            return Err(EthCallError::SimulateInvalidParams(
+                "calls are empty".to_string(),
+            ));
+        }
+        if opts.block_state_calls.len() > MAX_SIMULATE_BLOCKS as usize {
+            return Err(EthCallError::SimulateInvalidParams(
+                "too many blocks".to_string(),
+            ));
+        }
```
slowli: Doesn't `MAX_SIMULATE_BLOCKS` include gaps between specified blocks? At least that's the impression I got from [geth docs](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth#eth-simulate-v1):

> The total number of blocks for a single request is limited to 256. This includes any gap blocks that client generated.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -420,6 +432,136 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+                .storage_writes
+                .extend(simulated_block.overlay.storage_writes);
+            overlay.preimages.extend(simulated_block.overlay.preimages);
+            Arc::make_mut(&mut overlays).insert(block_context.block_number, overlay);
+            previous_block_number = block_context.block_number;
+            previous_timestamp = block_context.timestamp;
+            simulated_blocks.push(SimulatedBlock {
+                inner: simulated_block.inner,
+                calls: simulated_block.calls,
+            });
```
slowli: Shouldn't skipped blocks be included into the response as well? At least, the [geth docs say](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth#eth-simulate-v1):

> Skipping numbers is allowed and skipped blocks are included in the response.

Although, it seems like reth does return only blocks with specified calls in them, and it does limit block number by the number of such blocks.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -32,8 +39,10 @@ use zksync_os_types::{
     ZkTransaction, ZkTxType,
 };
 
-const ESTIMATE_GAS_ERROR_RATIO: f64 = 0.015;
+#[path = "simulation_utils.rs"]
```
slowli: This attribute looks unnecessary.

lib/rpc/src/result.rs
```diff
@@ -112,6 +112,22 @@ impl<Ok> ToRpcResult<Ok, EthCallError> for Result<Ok, EthCallError> {
...
+            }
+            EthCallError::SimulateBlockNumberInvalid { .. } => {
+                rpc_error_with_code(-38020, err.to_string())
+            }
+            EthCallError::SimulateBlockTimestampInvalid { .. } => {
+                rpc_error_with_code(-38021, err.to_string())
+            }
+            EthCallError::SimulateBlockGasLimitExceeded => {
+                rpc_error_with_code(-38015, err.to_string())
+            }
```
slowli, line 128: Could make sense to comment that these error codes correspond to the `reth` implementation. (I find them slightly weird since they're `< i16::MIN`, but if we strive for compatibility, we should take them as-is, I guess 🤷‍♂️)

lib/rpc/src/simulation_utils.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+    if let Some(blob_base_fee) = overrides.blob_base_fee {
+        block_context.blob_fee = blob_base_fee;
+    }
+    // TODO: difficulty override is not propagated to BlockContext (ZKsync OS uses mix_hash
+    // for prevrandao), so it is silently ignored.
+    if let Some(block_hash_overrides) = overrides.block_hash {
+        for (block_number, block_hash) in block_hash_overrides {
+            if block_number >= block_context.block_number {
+                continue;
```
slowli: Nit: Since `block_hash_overrides` is an ordered map, it's possible to `break` here. Alternatively, iterate over a [range](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range) of `block_context.block_number - 256..block_context.block_number`.

lib/rpc/src/simulation_utils.rs
```diff
@@ -0,0 +1,459 @@
...
+        return Err(EthCallError::SimulateBlockGasLimitExceeded);
+    }
+
+    let calls_without_gas = calls.iter().filter(|call| call.gas.is_none()).count() as u64;
+    if calls_without_gas == 0 {
+        return Ok(0);
+    }
+
+    let gas_per_call = (block_gas_limit - total_specified_gas) / calls_without_gas;
+    Ok(if call_gas_limit == 0 {
```
slowli: AFAICT, `eth_call_gas` from config which is passed here doesn't have the special "0 means infinity" handling in other uses. Is it really necessary here?

lib/rpc/src/simulation_utils.rs
```diff
@@ -0,0 +1,459 @@
...
+impl SimulatedTx {
+    fn is_executed(&self) -> bool {
+        matches!(&self.result, SimulatedTxResult::Executed { .. })
+    }
+
+    fn to_call_result(&self, block_hash: B256, block_context: BlockContext) -> SimCallResult {
+        match &self.result {
+            SimulatedTxResult::Executed { output, receipt } => {
+                let logs = self.api_logs(block_hash, block_context, receipt);
+                let (return_data, status, error) = match &output.execution_result {
```
slowli: Nit: `status` can be derived from `error.is_none()`.

lib/rpc/src/simulation_utils.rs
```diff
@@ -0,0 +1,459 @@
...
+            return Err(EthCallError::SimulateBlockTimestampInvalid {
+                got: time,
+                parent: previous_timestamp,
+            });
+        }
+        block_context.timestamp = time;
+    }
+    if let Some(gas_limit) = overrides.gas_limit {
+        block_context.gas_limit = gas_limit;
+    }
```
slowli: Shouldn't there be a check that the user-provided `gas_limit` isn't too large, similar to [one in `reth`](https://github.com/paradigmxyz/reth/blob/a60c3ee3404cbb11be79f3fb250d354eccf7b489/crates/rpc/rpc-eth-api/src/helpers/call.rs#L157-L163)?

lib/rpc/src/eth_call_handler.rs
```diff
@@ -709,12 +718,273 @@ pub fn update_estimated_gas_range(
...
+            ));
+        }
+
+        let simulation_utils::SimulationStartContext {
+            mut block_context,
+            parent_block_number,
+            parent_timestamp,
+        } = self.resolve_simulation_start_context(block)?;
+        let base_state = self.storage.state_view_at(parent_block_number)?;
+        let mut overlays = Arc::new(BTreeMap::new());
```
slowli: Comment: I see now that using a plain `BTreeMap` here wouldn't work, and using an `Arc` is a sound idea; sorry for a misleading earlier comment. (Maybe, the only improvement is to use an `Rc`; AFAIU, there's no requirement for storage trait impls to be `Send`.)

lib/rpc/src/simulation_utils.rs
```diff
@@ -0,0 +1,459 @@
...
+            .enumerate()
+            .map(|(i, log)| build_api_log(tx_hash, log, meta.clone(), i as u64))
+            .collect()
+    }
+
+    fn tx_meta(
+        &self,
+        block_hash: B256,
+        block_context: BlockContext,
+        gas_used: u64,
```
slowli: Nit: Since both invocations set this to `self.gas_used()`, it's possible to get rid of this arg and set it inside the method.

lib/rpc/src/simulate.rs
```diff
@@ -0,0 +1,459 @@
...
+                }
+            }
+            SimulatedTxResult::Invalid(err) => SimCallResult {
+                return_data: Bytes::default(),
+                logs: vec![],
+                gas_used: 0,
+                status: false,
+                error: Some(SimulateError {
+                    code: -32015,
+                    message: format!("vm execution error: {err}"),
```
slowli, line 580: AFAICT, `Display` impl for `InvalidTransaction` delegates to `Debug`, saying

> the error should be passed to the sequencer, converted to an appropriate public error through zksync-error framework and then passed to the clients.

So, is it appropriate to leak impl details here?

Artemka374: I think it makes sense, then it would be possible to understand why the simulation fails, no?

lib/rpc/src/eth_call_handler.rs
Artemka374: I think it's fine to follow reth's implementation for now(also applies for the comment above). We need just a basic implementation for now, if additional checks will be needed, we can add them separately

integration-tests/tests/rpc/simulate.rs
```diff
@@ -0,0 +1,109 @@
...
+    let counter = Counter::deploy(tester.l2_provider.clone()).await?;
+
+    let increment_call = counter.increment(U256::from(7)).into_transaction_request();
+    // A second increment in block 2 — if state did not carry over, this would fail or return wrong data.
+    let increment_call_2 = counter.increment(U256::from(3)).into_transaction_request();
+
+    let payload = SimulatePayload {
+        block_state_calls: vec![
+            SimBlock::default().call(increment_call),
+            SimBlock::default().call(increment_call_2),
```
itegulov, line 106: I'd make this test more informative by adding another block with a read call and then validate that it read `10` from the counter. Otherwise we can't be sure the blocks actually got applied sequentially onto each other.

integration-tests/tests/rpc/simulate.rs
```diff
@@ -0,0 +1,109 @@
...
+async fn simulate_multiple_txs_in_one_block(tester: Tester) -> anyhow::Result<()> {
+    let sender = tester.l2_wallet.default_signer().address();
+    let emitter = EventEmitter::deploy(tester.l2_provider.clone()).await?;
+    let recipient = alloy::primitives::address!("000000000000000000000000000000000000dead");
+
+    let payload = SimulatePayload {
+        block_state_calls: vec![
+            SimBlock::default()
+                .call(emitter.emitEvent(U256::from(1)).into_transaction_request())
+                .call(emitter.emitEvent(U256::from(2)).into_transaction_request())
```
itegulov, line 53: Can we check that both events got emitted in the result?

lib/storage_api/src/overlay_buffer.rs
```diff
@@ -5,14 +5,10 @@ use alloy::primitives::{B256, BlockNumber};
...
-use crate::state_override_view::OverrideProvider;
+use crate::state_override_view::{OverrideProvider, OwnedOverrides};
 use crate::{OverriddenStateView, ReadStateHistory, ViewState};
 
-#[derive(Debug, Clone)]
-pub struct BlockOverlay {
-    pub storage_writes: HashMap<B256, B256>,
-    pub preimages: HashMap<B256, Vec<u8>>,
-}
+pub type BlockOverlay = OwnedOverrides;
```
itegulov: You added type alias  so this file doesn't have too many changes? If so I don't think alias needs to be public, doesn't look like it's used anywhere

lib/rpc/src/eth_call_handler.rs
```diff
@@ -709,12 +716,262 @@ pub fn update_estimated_gas_range(
...
+
+        Ok(simulated_blocks)
+    }
+
+    fn resolve_simulation_start_context(
+        &self,
+        block: Option<BlockId>,
+    ) -> Result<simulation_utils::SimulationStartContext, EthCallError> {
+        let block = block.unwrap_or_default();
+        if block.is_latest() || block.is_pending() {
```
itegulov: My first reaction was that it feels weird for `latest` to be handled in this branch. But I suppose otherwise default usage would resolve `latest` which in zksync-os can be a relatively old block (unlike Ethereum we don't produce blocks on a cadence). And that means outdated pubdata price, gas price etc. Is my understanding correct?

lib/rpc/src/eth_call_handler.rs
```diff
@@ -709,12 +716,262 @@ pub fn update_estimated_gas_range(
...
+            parent_timestamp,
+        } = self.resolve_simulation_start_context(block)?;
+        let base_state = self.storage.state_view_at(parent_block_number)?;
+        let mut overlays = Arc::new(BTreeMap::new());
+        let mut simulated_blocks = Vec::with_capacity(block_state_calls.len());
+        let mut previous_block_number = parent_block_number;
+        let mut previous_timestamp = parent_timestamp;
+
+        for sim_block in block_state_calls {
+            block_context.mix_hash = U256::ZERO;
```
itegulov: Why?

lib/rpc/src/eth_call_handler.rs
```diff
@@ -709,12 +716,262 @@ pub fn update_estimated_gas_range(
...
+            if let Some(block_overrides) = sim_block.block_overrides {
+                simulation_utils::apply_simulate_block_overrides(
+                    &mut block_context,
+                    block_overrides,
+                    previous_block_number,
+                    previous_timestamp,
+                    self.config.eth_call_gas as u64,
```
itegulov: Hmm, this reuses `eth_call_gas` which is per single transaction where this is per block. I think we should consider having a separate config or maybe rename that field to something else. At least change the doc for it

lib/rpc/src/simulation_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+}
+
+pub(crate) fn apply_simulate_block_overrides(
+    block_context: &mut BlockContext,
+    overrides: BlockOverrides,
+    previous_block_number: u64,
+    previous_timestamp: u64,
+    max_block_gas_limit: u64,
+) -> Result<(), EthCallError> {
+    if let Some(number) = overrides.number {
```
itegulov: I'd deconstruct `overrides` into all of its fields first just to make it clear we are handling everything. With separate branches for each field it's hard to check that it's the case.

lib/rpc/src/simulation_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+        }
+        block_context.gas_limit = gas_limit;
+    }
+    if let Some(coinbase) = overrides.coinbase {
+        block_context.coinbase = coinbase;
+    }
+    if let Some(random) = overrides.random {
+        block_context.mix_hash = U256::from_be_bytes(random.0);
+    } else {
+        block_context.mix_hash = U256::ZERO;
```
itegulov: Again: why?

lib/rpc/src/simulation_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+    }
+    if let Some(base_fee) = overrides.base_fee {
+        block_context.eip1559_basefee = base_fee;
+    }
+    if let Some(blob_base_fee) = overrides.blob_base_fee {
+        block_context.blob_fee = blob_base_fee;
+    }
+    // TODO: difficulty override is not propagated to BlockContext (ZKsync OS uses mix_hash
+    // for prevrandao), so it is silently ignored.
```
itegulov: Well, this is not so much a `TODO` as it is just a comment. I don't think there is anything actionable here, is there?

itegulov (commented):
Partial review, will finish in a bit

lib/storage_api/src/overlay_buffer.rs
Artemka374: Not really, these types are identical, so I just unified them in a way

lib/rpc/src/eth_call_handler.rs
Artemka374: This is the default by ethereum spec: https://github.com/ethereum/execution-apis/blob/main/docs-api/docs/ethsimulatev1-notes.mdx. It also matches the code from reth & geth

lib/rpc/src/simulation_utils.rs
Artemka374: I don't think there is, I'll change it to regular comment

lib/storage_api/src/overlay_buffer.rs
itegulov: No I get that but there must've been a reason you didn't just start using `OwnedOverrides` everywhere no?

lib/rpc/src/simulation_utils.rs
itegulov: Probably redundant to have it here too since you set it to zero in that other place then

lib/rpc/src/simulation_utils.rs
```diff
@@ -0,0 +1,454 @@
...
+    if total_specified_gas > block_gas_limit {
+        return Err(EthCallError::SimulateBlockGasLimitExceeded);
+    }
+
+    let calls_without_gas = calls.iter().filter(|call| call.gas.is_none()).count() as u64;
+    if calls_without_gas == 0 {
+        return Ok(0);
+    }
+
+    Ok((block_gas_limit - total_specified_gas) / calls_without_gas)
```
itegulov: Do we need to `min` with `config.eth_call_gas` here?

lib/rpc/src/eth_call_handler.rs
itegulov, line 1: Consider moving implementation to `simulate.rs` along with code in `simulation_utils.rs`. The logic is very interconnected so it feels weird having these two split apart like that. I know you use some `self` methods here but from a brief glance they are general enough to actually be shared rather than being in `eth_call_handler` specifically.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -709,12 +716,262 @@ pub fn update_estimated_gas_range(
...
+                OverriddenStateView::new(simulation_view, state_overrides.clone());
+            let txs = self.create_simulation_txs(
+                sim_block.calls,
+                execution_context,
+                overridden_view.clone(),
+            )?;
+            let tx_source = TxListSource {
+                transactions: txs.iter().cloned().map(|tx| tx.encode()).collect(),
+            };
+            let block_output = run_block(
```
itegulov: Note that for `V5` this will use version of zksync-os that does not support estimation for zero-balance accounts. See comment around `RunBlockForwardV5Running`. Not sure how much of a big deal this is, likely not very given that `V6` is coming soon. But flagging just in case

itegulov (commented):
Finished review, LGTM overall

lib/storage_api/src/overlay_buffer.rs
Artemka374: ah yes, this was mostly to reduce the diff

lib/rpc/src/eth_call_handler.rs
Artemka374: it was like that initially, but I thought it made sense to leave the method itself in eth_call_handler. No strong opinion though

lib/rpc/src/simulate.rs
slowli: Yes, makes sense; for some reason, I incorrectly thought that `InvalidTransaction` corresponds to internal VM errors. There *is* an issue of UX (debug representation is somewhat difficult to parse for non-developers), but it's clearly outside the scope of this PR.

lib/rpc/src/simulate.rs
```diff
@@ -0,0 +1,722 @@
...
+};
+use zksync_os_multivm::run_block;
+use zksync_os_rpc_api::types::ZkApiBlock;
+use zksync_os_storage_api::ViewState;
+use zksync_os_storage_api::state_override_view::{
+    OverriddenStateView, OwnedOverrides, build_state_override_maps,
+};
+use zksync_os_types::{ZkReceipt, ZkReceiptEnvelope, ZkTransaction, ZksyncOsEncode};
+
+impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
```
itegulov, line 31: I actually meant to have a whole separate struct and errors here. `create_tx_from_request` and `build_pending_block_context` seem general enough that they might belong to `util` or maybe even an extension method to `RpcStorage` itself.

This is already better though

itegulov (approved):
One more comment but looks mergeable to me already

lib/rpc/src/eth_call_handler.rs
Artemka374: agreed, I can put everything else in a follow-up, I'd like to merge the PR

lib/rpc/src/simulate.rs
Artemka374: Agree, I'd put everything else in a follow-up

PR #1241, opened 2026-04-30
perf(merkle-tree): Optimize passing tree data through component pipeline

slowli:
## Summary

Refactors passing tree data through components.

Previously, the tree handle itself was passed through components, starting from the tree manager. This resulted in high I/O overhead in PIG, which needs to re-read storage data and proofs via tree using pointwise operations. The new approach uses the `MerkleTree::extend_with_proof()` method, which was designed with this specific use case in mind and uses batched I/O.

## Discussion and future improvements

- Refactoring requires recording (unique) reads during block execution in the sequencer, which is straightforward to implement and shouldn't produce much overhead. These could theoretically be used to extend the `revm` consistency checker.
- `MerkleTree::extend_with_proof()` is slower than `extend()` since it needs to take reads into account. On ENs (or, more abstractly, nodes that don't run a PIG), we can switch off `extend_with_proof()`, or read collection in the sequencer altogether.
- Alternatively, it should be possible to use `extend_with_proof()` w/o reads (this should have zero I/O overhead compared to `extend()`) and then augment the proof with reads separately in the PIG itself. This would use batch I/O, so is likely to still be significantly more efficient than the current approach.
- The implemented approach uses a fallback reading from `MerkleTree` just in case (equivalent to `MerkleTreeVersion`, which is removed as low-quality). Normally, the fallback should never be triggered, so if we observe that it works properly, it could be removed in the future.


lib/sequencer/src/execution/vm_wrapper.rs
```diff
@@ -39,15 +41,21 @@ impl VmWrapper {
...
                 tx_source,
                 tx_callback,
                 &mut tracer,
                 &mut validator,
-            )
+            )?;
+
+            Ok(BlockOutputWithReads::new(
+                block_output,
+                recording_handle.into_read_keys(),
```
RomanBrodetski, line 57: Now that we have it, would be good to also log the number of accessed slots [here](https://github.com/matter-labs/zksync-os-server/blob/c5c339f345fcdae447fcc02a5a29b13045184758/lib/sequencer/src/execution/execute_block_in_vm.rs#L390) (in case it's possible without making code awkward)

node/bin/src/prover_input_generator/mod.rs
```diff
@@ -288,7 +295,7 @@ fn compute_prover_input(
                     last_block_timestamp: replay_record.previous_block_timestamp,
                 },
                 da_commitment_scheme,
-                tree_view,
+                TreeOutputAdapter::new(tree_view).with_fallback(versioned_tree),
```
RomanBrodetski, line 306: nice way to safely rollout, thanks!

RomanBrodetski (dismissed):

RomanBrodetski (approved):
Looks good!

let's make sure we track the effect on performance when we roll this out (a good env to see is genlayer-testnet). 
@slowli please track it - that is, just make a note when we release this PR.

PR #1242, opened 2026-05-01
fix: Reduce logging verbosity

EmilLuta:
This PR gets rid of printing the entire block command & moves tx logging to debug. External Nodes on high traffic could generate up to 50GB of logs for a full sync.


perekopskiy (approved):
It was a matter of time before we would have to reduce verbosity, so no objections

PR #1244, opened 2026-05-01
perf: Default to jemalloc for mem allocation

EmilLuta:
Enables jemalloc as the default allocator on Unix targets. It is feature gated (by default on), but one can opt out via `--no-default-features`.

All releases are running under Linux machines. As such the likely default choice is glibc. The problem with glibc is that it is lazy to release memory and under heavy pressure, it hogs memory for "later use". Another issue is that memory gets fragmented and glibc does an "average" job at reclaiming it properly (at least in use cases I've observed). Jemalloc is a better fit for long running, high concurrency processes.

For this repo, we have plenty of such cases (i.e. a mainnode under heavy load or a syncing EN).

I have tested EN sync scenario and it reduced overall potential memory usage from ~20GB (it might go higher than this, but didn't wait for a full sync, the data I got in a couple of runs in ~2h was enough to justify the change) to max ~11GB. I've ran multiple tests to compare EN memory and they are all positive.

I have high confidence that this will be operationally better for all nodes (including existing mainnodes with high load), albeit, didn't manage to test a a main node locally (i.e reproduce high load from an existing a main node locally).

### DIFF
--- a/node/bin/src/main.rs
+++ b/node/bin/src/main.rs
@@ -18,6 +18,10 @@ use zksync_os_server::{INTERNAL_CONFIG_FILE_NAME, run};
 use zksync_os_state::StateHandle;
 use zksync_os_state_full_diffs::FullDiffsState;
 
+#[cfg(all(feature = "jemalloc", target_family = "unix"))]
+#[global_allocator]
+static GLOBAL_ALLOCATOR: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
+
 const IMMEDIATE_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(1);
 const GRACEFUL_SHUTDOWN_TIMEOUT: Duration = Duration::from_secs(10);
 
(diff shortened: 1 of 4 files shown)


perekopskiy (approved):

node/bin/src/main.rs
```diff
@@ -18,6 +18,10 @@ use zksync_os_server::{INTERNAL_CONFIG_FILE_NAME, run};
 use zksync_os_state::StateHandle;
 use zksync_os_state_full_diffs::FullDiffsState;
 
+#[cfg(all(feature = "jemalloc", target_family = "unix"))]
+#[global_allocator]
+static GLOBAL_ALLOCATOR: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
```
LogvinovLeon, line 23: Super nit: Is there any logic behind using fully-qualified name here?
Why not import and:
```suggestion
static GLOBAL_ALLOCATOR: Jemalloc = Jemalloc;
```

PR #1254, opened 2026-05-05
fix(prover-api): potential race conditions in SnarkJobManager

perekopskiy:
## Summary

Fixes SNARK prover job handling when fake and real provers race on the same batches.

The fake SNARK path previously selected a consecutive range, but completed jobs one by one. If a real prover completed an overlapping range during that loop, the fake path could emit a partial proof command, breaking downstream gapless ordering and potentially leaving the proof pipeline stuck.

This PR:

  - Completes fake SNARK ranges atomically with complete_many_jobs.
  - Skips fake proof emission if another prover already completed part of the selected range.
  - Reserves downstream SNARK channel capacity before removing jobs from the retryable job map, matching the FRI manager behavior with try_reserve.
  - Validates SNARK proving version / VK before consuming jobs.

### DIFF
--- a/node/bin/src/prover_api/snark_job_manager.rs
+++ b/node/bin/src/prover_api/snark_job_manager.rs
@@ -261,10 +261,13 @@ impl FakeSnarkProver {
     pub async fn run(self) {
         loop {
             tokio::time::sleep(self.polling_interval).await;
-            self.job_manager
+            if let Err(err) = self
+                .job_manager
                 .process_pending_fake_or_timed_out_fri_proofs(Some(self.max_batch_age))
                 .await
-                .expect("snark prover failed");
+            {
+                tracing::info!("`FakeSnarkProver` iteration failed: {err}");
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


node/bin/src/prover_api/snark_job_manager.rs
```diff
@@ -186,38 +191,51 @@ impl SnarkJobManager {
...
-                    .jobs
-                    .complete_job(job.batch_number, ProverType::Fake, "fake_prover")
-                    .await
-                {
-                    completed.push(envelope);
-                }
-            }
+            let batch_from = assigned.first().unwrap().0.batch_number;
+            let batch_to = assigned.last().unwrap().0.batch_number;
+            let permit = self.try_reserve_permit_downstream()?;
```
EmilLuta, line 196: Wouldn't this break the fake proving pipeline? Previously, we'd block & wait on this call. Nowe we send an error here, that ends up bubbling up and panicing with `.expect("snark prover failed");`.

In practice, I'm not 100% sure this will happen. But we have a single channel so if there are delays between `complete_many_jobs()` and `try_reserve`, this could happen, right?

I could be wrong, I don't know enough of the codebase.

perekopskiy: Good catch! Indeed we shouldn't panic, just retry in this case, fixed

RomanBrodetski (approved):

PR #1256, opened 2026-05-06
fix(rocksdb): retry repeated stalled writes

sidhujag:
## Summary

- Fixes the local RocksDB wrapper so repeated stalled-write errors keep using the configured retry intervals instead of returning after the second stall.
- Keeps stalled-write metrics reporting to one increment per stall incident, with a `SYSCOIN` tag marking the fork delta against upstream code.

## Test plan

- `cargo fmt -p zksync_os_rocksdb -- --check`
- `ReadLints` on `lib/rocksdb/src/db.rs`
- `cargo test -p zksync_os_rocksdb` was started locally but interrupted after ~69s per request to avoid local validation.
- Remote `cargo build` on `ubuntu@88.198.25.188` was attempted, but the default `/usr/bin/cargo` is 1.75 and cannot parse `resolver = "3"`; follow-up remote build was interrupted when switching to PR creation.


sidhujag:
Opened against the wrong upstream repository by mistake; this fork-specific fix will be proposed against syscoin/zksync-os-server instead.

PR #1257, opened 2026-05-06
feat(provider)!: Add metrics & retry config

kostia244:
## Summary

Add provider metrics (L1 and Gateway have their own):
- Request Latency labeled by method name (end to end, including retries).
- Retry Count

Add ProviderConfig:
- max_retries (new)
- retry_count (new)
- poll_interval (**moved out of general config**)
- rpc_url (**moved out of general config**)
ProviderConfig is used for both L1 and gateway. Now there are 4 config variables for the provider so it makes sense to group them. Gateway provider is constructed the same way, so the same config type can be used.
Retry config will help with EN sync by reducing number of crashes.

Config module changes:
- Add a layer for tracking latency.
- Add a retry layer based on the existing retry policy.

## Breaking changes
The following config variables:
`general_l1_rpc_url, general_gateway_rpc_url, general_l1_poll_interval, general_gateway_poll_interval` have been replaced by:
`l1_provider_rpc_url, gateway_provider_rpc_url, l1_provider_poll_interval, gateway_provider_poll_interval`
All configs are affected.
You should rename these variables for the upgrade.


node/bin/src/config/mod.rs
```diff
@@ -451,6 +441,38 @@ pub struct GeneralConfig {
     pub ephemeral_state: Option<PathBuf>,
 }
 
+/// Config for L1 or Gateway provider
+#[derive(Clone, Debug, DescribeConfig, DeserializeConfig, ConfigValidate)]
+#[config(derive(Default))]
+pub struct ProviderConfig {
+    /// JSON RPC API URL.
+    #[config(default_t = "http://localhost:8545".into())]
+    pub rpc_url: String,
```
kostia244, line 450: I think putting this here makes a lot of sense. But it might be a painful because it breaks old configs in the current state.

node/bin/src/config/mod.rs
```diff
@@ -451,6 +441,38 @@ pub struct GeneralConfig {
...
+    #[config(default_t = 7 * TimeUnit::Seconds)]
+    pub rpc_poll_interval: Duration,
+
+    /// Maximum number of retry attempts, excluding the initial attempt.
+    #[config(default_t = 2)]
+    pub max_retries: u32,
+
+    /// Backoff used between retry attempts.
+    #[config(default_t = Duration::from_millis(200))]
+    pub retry_backoff: Duration,
```
itegulov, line 463: We try to keep defaults reasonable for real (production) testnet environments. With that said, should this and `max_retries` have better defaults?

node/bin/src/provider/retry.rs
```diff
@@ -0,0 +1,104 @@
...
+use alloy::transports::layers::{RateLimitRetryPolicy, RetryPolicy};
+use alloy::transports::{TransportError, TransportErrorKind, TransportFut};
+use std::task::{Context, Poll};
+use std::time::Duration;
+use tokio::time::sleep;
+use tower::Service;
+
+/// Retry RPC requests & track retry count metric
+#[derive(Debug, Clone)]
+pub(super) struct RetryService<S> {
```
itegulov, line 13: What's the reason for writing a custom service over `RetryBackoffLayer`? Exposing `retry_count` metric?

itegulov (commented):
LGTM overall, just a couple of comments

node/bin/src/config/mod.rs
kostia244: I personally use 30 retries and 2000ms. But that's because I use a cheap RPC. I don't know what are reasonable defaults for our infra.

node/bin/src/provider/retry.rs
kostia244: Yes. 
I also think it can be useful if want to improve error handling or have more detailed metrics.

perekopskiy (dismissed):
LGTM

node/bin/src/config/mod.rs
itegulov: I talked to DevOps, they told me this should be good enough: retries=5, backoff=1s

itegulov (approved):

EmilLuta:
Uh-oh, this rename caused a lot of pain.

PR #1260, opened 2026-05-07
feat(l1-sender): Refresh balance & nonce more often, refactor for simpler method signatures

kostia244:
## Summary

Right now operator balance is updated when submitting, which could be once an hour. But prometheus exports metrics every 30 seconds anyway. This can lead to some confusion if balance changes. Especially when looking at plots. I added a loop that updates the values every minute.

I also moved all the sender functions into L1Sender struct. They were passing around a lot of the same data(run_l1_sender literally took all the member fields of L1Sender as arguments) and each of them had the same generic boilerplate. So now it should be simpler to work with.

### DIFF
--- a/lib/l1_sender/src/pipeline_component.rs
+++ b/lib/l1_sender/src/pipeline_component.rs
@@ -1,6 +1,5 @@
 use crate::commands::{L1SenderCommand, SendToL1};
 use crate::config::L1SenderConfig;
-use crate::run_l1_sender;
 use alloy::network::{Ethereum, EthereumWallet};
 use alloy::primitives::Address;
 use alloy::providers::fillers::{FillProvider, TxFiller};
@@ -42,17 +41,6 @@ where
         output: mpsc::Sender<Self::Output>,
         state_reporter: ComponentStateReporter,
     ) -> anyhow::Result<()> {
-        run_l1_sender(
-            input,
-            output,
-            self.to_address,
-            self.provider,
-            self.config,
-            self.gateway,
-            state_reporter,
-            self.commit_submitted_tx,
-            self.sl_block_number,
-        )
-        .await
+        self.run_l1_sender(input, output, state_reporter).await
     }
 }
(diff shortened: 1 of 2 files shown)


kostia244:
I moved a lot of things from one file to another.
Here's the diff before the move, should be much more readable
https://github.com/matter-labs/zksync-os-server/pull/1260/changes/BASE..6391b7d8b59e1b92889facdc29d87f19c12c62c7

lib/l1_sender/src/pipeline_component.rs
```diff
@@ -38,20 +72,688 @@ where
...
+    ///     Otherwise, there will be a nonce conflict and a failed L1 transaction
+    ///     (recoverable on restart)
+    ///
+    /// Known issues:
+    ///   * Crashes when there is a gap in incoming L1 blocks (happens periodically with Infura provider)
+    ///
+    /// Note: we pass `to_address` - L1 contract address to send transactions to.
+    /// It differs between commit/prove/execute (e.g., timelock vs diamond proxy)
+    pub async fn run_l1_sender(
+        &self,
```
Artemka374: hmm, I wonder if it can be avaliable via just `self`, so the function is consuming

Artemka374 (dismissed):
Overall LGTM

lib/l1_sender/src/pipeline_component.rs
kostia244: Is there any benefit to doing this? Right now both sender and metrics loops borrow it, so we'd have to change something.

kostia244:
I had to re-do a bunch of things due to merge conflicts.
Please use this to review:
1. [refactor part, w=1 makes diff 3x smaller](https://github.com/matter-labs/zksync-os-server/pull/1260/changes/BASE..7b1d2b0a0384901b7ba83442c6bbcf4811f83743?w=1)
2. [update metrics part](https://github.com/matter-labs/zksync-os-server/pull/1260/changes/7b1d2b0a0384901b7ba83442c6bbcf4811f83743..4bf03cc000e250eda2568ffb833e1325e62999db)

@Artemka374

Artemka374 (approved):

PR #1261, opened 2026-05-07
feat(rpc): track gas used per eth_call

LogvinovLeon:
## Summary

- Adds `api_call_gas_used{method="eth_call"}` histogram with round gas buckets (20k, 50k, 100k, 200k, 500k, 1M, 2M, 5M, 10M, 20M, 50M, 100M, +Inf) to the existing `Api` metrics struct
- Records gas from `TxOutput.gas_used` after every `eth_call` execution, including reverts — only skipped on pre-execution failures (`InvalidTransaction`) and VM crashes (`ForwardSubsystemError`) where no `TxOutput` is available
- Consolidates the `state_overrides` match into a single `OverriddenStateView` construction (empty overrides are a no-op pass-through, so no behaviour change)

Motivation: when clients spam heavy `eth_call`s the existing `api_response_time_seconds` histogram shows latency but not why. `api_call_gas_used` lets you correlate slow calls with actual EVM work, alert on p99 gas, and compute total gas throughput per second.

Example queries:
```promql
# p99 gas per eth_call
histogram_quantile(0.99, rate(api_call_gas_used_bucket{method="eth_call"}[5m]))

# total gas load from eth_call per second
rate(api_call_gas_used_sum{method="eth_call"}[5m])
```

## Test plan

- [ ] Build passes (`cargo build -p zksync_os_rpc`)
- [ ] Existing RPC tests pass (`cargo nextest run -p zksync_os_rpc`)
- [ ] Verify metric appears in Prometheus scrape after an `eth_call`

### DIFF
--- a/lib/rpc/src/eth_call_handler.rs
+++ b/lib/rpc/src/eth_call_handler.rs
@@ -1,6 +1,7 @@
 use crate::call_fees::{CallFees, CallFeesError};
 use crate::config::RpcConfig;
 use crate::js_tracer;
+use crate::metrics::API_METRICS;
 use crate::result::RevertError;
 use crate::rpc_storage::{ReadRpcStorage, RpcStorageError};
 use crate::sandbox::{call_trace_simulate, execute};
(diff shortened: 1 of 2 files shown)


lib/rpc/src/metrics.rs
```diff
@@ -92,6 +106,10 @@ pub struct Api {
     pub cancelled: LabeledFamily<String, Counter>,
     #[metrics(labels = ["method"])]
     pub panicked: LabeledFamily<String, Counter>,
+    /// Gas consumed by methods whose execution cost scales with gas (eth_call, eth_estimateGas,
+    /// debug_traceTransaction, debug_traceCall). Always recorded, including on revert.
+    #[metrics(labels = ["method"], buckets = GAS_BUCKETS)]
+    pub call_gas_used: LabeledFamily<String, Histogram<u64>>,
```
kostia244, line 112: Would it be hard to implement tracking for other methods in the same PR?

lib/rpc/src/eth_call_handler.rs
```diff
@@ -302,26 +303,26 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
         block_overrides: Option<Box<BlockOverrides>>,
     ) -> Result<Bytes, EthCallError> {
         let mut execution_env = self.prepare_execution_env(request, block, block_overrides)?;
+        execution_env.block_context.eip1559_basefee = U256::from(0);
```
kostia244, line 306: I am not an expert, afaiu this shouldn't change the execution in any way, right?

lib/rpc/src/metrics.rs
kostia244: Also, how do you feel about adding a plot with "total gas / minute". Similar to how you have token cost histograms.

Image

kostia244 (commented):
Thank you! Overall looks good, left some comments.

lib/rpc/src/eth_call_handler.rs
LogvinovLeon: This line was there before. It was just moved

lib/rpc/src/metrics.rs
LogvinovLeon: It will be hard. Example:
eth_estimateGas is a nice offender. But it doesn't execute transaction once. It does binary search.
Another offender is debug_traceBlock. But there we do mutliple tx executions and some of them are allowed to fail. So decided to go 80/20 here with eth_call as it's big chunk of the traffic.

I'm working on the next PR that will refactor eth_estimateGas because it's frequently slow but it's also so complex that I don't understand it

LogvinovLeon: As for the plot - can do it for eth_call

kostia244 (approved):

PR #1265, opened 2026-05-08
feat(observability): semi-automatic export of config values as metrics 

kostia244:
## Summary

It's actually pretty easy to iterate over the configs. The only problem is that the main config does not derive DescribeConfig/DeserializeConfig.

If we're fine with extra line per member this would work.
image

### DIFF
--- a/node/bin/src/config/util.rs
+++ b/node/bin/src/config/util.rs
@@ -1,42 +1,12 @@
-use super::metrics::CONFIG_METRICS;
 use alloy::primitives::B256;
 use alloy::signers::k256::ecdsa::SigningKey;
 use serde::Deserialize;
 use serde_json::Value;
+use smart_config::ErrorWithOrigin;
 use smart_config::de::{DeserializeContext, DeserializeParam};
 use smart_config::metadata::{BasicTypes, ParamMetadata};
-use smart_config::{DescribeConfig, ErrorWithOrigin, SerializerOptions};
 use zksync_os_network::SecretKey;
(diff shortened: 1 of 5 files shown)


node/bin/src/main.rs
```diff
@@ -168,6 +169,8 @@ pub async fn main() {
     load_internal_config(&mut config);
     config.validate().await.expect("invalid config");
 
+    report_flat_config_metrics(&config.prover_api_config, "prover_api");
```
kostia244: We'd have to do this for all members of config, or change the config itself. But I thikn it's ready much nicer than doing it by hand.

node/bin/src/config/util.rs
```diff
@@ -1,13 +1,43 @@
...
+    }
+}
+
+fn flat_config_metric_entries<C: DescribeConfig>(
+    config: &C,
+    prefix: &str,
+) -> Vec<(String, String)> {
+    SerializerOptions::default()
+        .flat(true)
+        .serialize(config)
```
kostia244: I suppose this is a stable feature?

slowli: Yes, it is.

slowli (commented):
You can also use [metrics reporting in Era](https://github.com/matter-labs/zksync-era/blob/main/core/lib/config/src/observability_ext/metrics.rs) as a reference; it essentially implements the same concept (with additional labels for signaling whether a config param has a default value, and whether it is secret).

cytadela8 (approved):

PR #1267, opened 2026-05-08
feat(revm-checker): adopt new zksync-os-revm version

0xVolosnikov:
## Summary

Bumps `zksync-os-revm` to the **v0.1.0** release tag, which adds **AtlasV3** spec support — the spec corresponding to **ZKsync OS v0.3.0**. The consistency checker is adapted to the new tx/block envelope, mirroring zksync-os' `tests/revm_runner`.

AtlasV3-specific features now available to the checker: v3 hook precompiles (L1 messenger, set-bytecode-on-address, mint-base-token), treasury-based L1→L2 accounting via `BASE_TOKEN_HOLDER_ADDRESS`, spec-aware P256 warming, and corrected base-fee-zero gas-price handling. Note that the server's `zk_spec_version` mapping still routes V4–V6 to `AtlasV2`; wiring a new `ExecutionVersion` variant to `AtlasV3` is a one-liner follow-up that should land alongside the v0.3.0 protocol bump.

Required infrastructure changes:
- Switches the crate from `reth-revm` (revm 34.0.0) to `revm =38.0.0` directly — the new `zksync-os-revm` pins revm 38.0.0 and the EVM context, `CacheDB`, and `DatabaseRef` must come from the same revm version as the executor.
- Plumbs `settlement_layer_chain_id` (read from `SYSTEM_CONTEXT_ADDRESS` slot 0), blob fields, authorization list through the tx builder; sets `block.prevrandao` from `mix_hash` and configures `blob_excess_gas_and_price`.
- Adds support for system transactions (type `0x7d`); the new revm activates the service-tx code path purely from the tx type.
- Bytecode hashing delegates to `zk_os_api::set_properties_code`, removing the local blake2/jump-table reimplementation.

Operational hardening:
- `eip1559_basefee` and `blob_fee` use `saturating_to::<u64>()` so unexpected fees don't tear down the pipeline.
- New `revm_blocks_skipped` push metric for visibility into silent skips (e.g. tx-conversion failures).

## Test plan

- [x] `cargo build -p zksync_os_server`
- [x] `cargo clippy -p zksync_os_revm_consistency_checker --all-targets -- -D warnings`
- [x] `cargo test -p zksync_os_revm_consistency_checker --lib` (4/4 incl. ported blob-fee inversion tests)
- [x] `cargo fmt --check`
- [ ] Sanity-check against a node replaying real blocks (no divergences)


0xVolosnikov:
@itegulov should I rebase it on top of https://github.com/matter-labs/zksync-os-server/pull/1228 ?

Cargo.toml
```diff
@@ -254,7 +254,8 @@ reth-provider = { git = "https://github.com/matter-labs/reth.git", branch = "fix
 reth-rpc-eth-types = { git = "https://github.com/matter-labs/reth.git", branch = "fix/discv5-shutdown-v1.11.1" }
 reth-tasks = { git = "https://github.com/matter-labs/reth.git", branch = "fix/discv5-shutdown-v1.11.1" }
 
-zksync-os-revm = { git = "https://github.com/matter-labs/zksync-os-revm", tag = "v0.0.4" }
+zksync-os-revm = { git = "https://github.com/matter-labs/zksync-os-revm", rev = "6777d01e63aa7bbe4b0478e496f0be04488ffc52" }
```
0xVolosnikov: Should be pinned to a new tag before merging

itegulov (commented):
I think you can keep as-is for now. The only downside is extra duplicate dependencies to compile which is not a huge deal short-term.

We had to patch reth again yesterday so looks like we'll need a custom fork for a little while longer while we wait for patch to be upstreamed.

lib/revm_consistency_checker/src/node.rs
```diff
@@ -165,37 +190,55 @@ where
...
                         block.timestamp = U256::from(replay_record.block_context.timestamp);
                         block.beneficiary = replay_record.block_context.coinbase;
-                        block.basefee = replay_record.block_context.eip1559_basefee.saturating_to();
+                        block.basefee = block_basefee;
                         block.gas_limit = replay_record.block_context.gas_limit;
-                        // `replay_record.block_context` holds an incorrect `prevrandao` value.
-                        // We use the actual value that ZKsync OS uses instead.
-                        block.prevrandao = Some(U256::ONE.into());
+                        block.prevrandao = Some(replay_record.block_context.mix_hash.into());
+                        block.blob_excess_gas_and_price = Some(blob_excess_gas_and_price);
```
perekopskiy: Won't this break check for old blocks?

perekopskiy (commented):
> wiring a new ExecutionVersion variant to AtlasV3 is a one-liner follow-up that should land alongside the v0.3.0 protocol bump.

Why should it be a follow-up?

0xVolosnikov:
> Why should it be a follow-up?

I noticed that we don't have any wiring for the new version few days ago, prob I looked in the wrong place, can do it, np

lib/revm_consistency_checker/src/node.rs
0xVolosnikov: Thanks, let's me check these things!

0xVolosnikov: Fixed

perekopskiy (approved):
Looks good, cc @itegulov do you think we should enable revm checker on localhost with revert by default? (in local_dev.yaml)

itegulov:
> do you think we should enable revm checker on localhost with revert by default? (in local_dev.yaml)

I think that makes sense. It will essentially crash and write the batch that needs to be reverted in the local config, correct? i.e. will not be deleted automatically before you rerun the node again

perekopskiy:
yes, but anyway it's more important for tests where it will just fail and that's enough to investigate

PR #1268, opened 2026-05-08
refactor(rpc): refactor estimate_gas_with_view for clarity

LogvinovLeon:
## Summary

Refactors `estimate_gas_with_view` for readability and correctness.

**Stats:** the method was 207 lines, now split into `build_estimate_tx` (24 lines) + `estimate_gas_with_view` (64 lines).

- Extract `max_gas_from_balance` and `build_estimate_tx` helpers to separate tx setup from estimation logic
- Introduce `GasRange` struct to encapsulate binary search bounds with invariant: tx fails at `lowest`, succeeds at `highest`
- Introduce `Probe` enum (`Highest`/`Optimistic`/`Midpoint`) to carry gas limit + kind, enabling centralized logging and `apply_probe` dispatch
- Fix bug: OOG errors in the optimistic check were propagating as hard errors instead of narrowing the range (matching geth/reth behavior)
- Remove dead code: `CallerGasLimitMoreThanBlock` arm in binary search (impossible since all probes ≤ block gas limit)
- Simplify `biased_midpoint` to use `range.lowest * 3` instead of tracking `gas_used` separately
- Rename `try_at` closure to `run_at`; add `ViewState::balance` convenience method

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,7 +33,7 @@ dependencies = [
  "bitflags 2.11.0",
  "bytes",
  "bytestring",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "encoding_rs",
  "foldhash 0.1.5",
  "futures-core",
@@ -131,7 +131,7 @@ dependencies = [
  "bytes",
  "bytestring",
  "cfg-if",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "encoding_rs",
  "foldhash 0.1.5",
  "futures-core",
@@ -340,7 +340,7 @@ dependencies = [
  "auto_impl",
  "borsh",
  "c-kzg",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "either",
  "k256",
  "once_cell",
@@ -368,7 +368,7 @@ dependencies = [
  "auto_impl",
  "borsh",
  "c-kzg",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "either",
  "k256",
  "once_cell",
@@ -442,7 +442,7 @@ dependencies = [
  "alloy-primitives",
  "alloy-sol-type-parser",
  "alloy-sol-types",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "itoa",
  "serde",
  "serde_json",
@@ -526,7 +526,7 @@ dependencies = [
  "auto_impl",
  "borsh",
  "c-kzg",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "either",
  "serde",
  "serde_with",
@@ -551,7 +551,7 @@ dependencies = [
  "auto_impl",
  "borsh",
  "c-kzg",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "either",
  "serde",
  "serde_with",
@@ -572,7 +572,7 @@ dependencies = [
  "alloy-rpc-types-eth",
  "alloy-sol-types",
  "auto_impl",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "revm",
  "thiserror 2.0.17",
  "tracing",
@@ -666,7 +666,7 @@ dependencies = [
  "alloy-sol-types",
  "async-trait",
  "auto_impl",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "futures-utils-wasm",
  "serde",
  "serde_json",
@@ -719,7 +719,7 @@ dependencies = [
  "bytes",
  "cfg-if",
  "const-hex",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "foldhash 0.2.0",
  "getrandom 0.4.2",
  "hashbrown 0.17.1",
@@ -918,7 +918,7 @@ checksum = "48b9ad6eee93dd35a9ec0a6c1c6b180892a900ee17a6ed6500921552dd71e846"
 dependencies = [
  "alloy-primitives",
  "alloy-rlp",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "serde",
  "serde_with",
 ]
@@ -934,7 +934,7 @@ dependencies = [
  "alloy-primitives",
  "alloy-rlp",
  "alloy-serde 2.0.4",
- "derive_more 2.0.1",
+ "derive_more 2.1.1",
  "rand 0.8.5",
  "serde",
  "strum",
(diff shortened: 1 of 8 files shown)


lib/rpc/src/eth_call_handler.rs
```diff
@@ -451,218 +449,171 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
         // Execute the transaction with the highest possible gas limit.
-        let mut res = execute(tx.clone(), block_context, storage_view.clone())
-            .map_err(EthCallError::ForwardSubsystemError)?
-            .map_err(EthCallError::InvalidTransaction)?;
+        let res = run_at(tx.gas_limit())?.map_err(EthCallError::InvalidTransaction)?;
         tracing::trace!(
-            "Executed tx in estimate_gas with highest gas limit {}, result {res:?}",
-            tx.gas_limit(),
+            "Executed tx in eth_estimateGas with gas limit: {}, result {res:?}",
+            Probe::Highest(tx.gas_limit())
```
itegulov, line 517: Doesn't look like `Probe::Highest` falls nicely in the proposed abstraction layer. Any way we can have this logic go through the common flow in `apply_probe`? Otherwise I'd get rid of it

itegulov (commented):
Just one suggestion but refactoring looks good to me overall

lib/rpc/src/eth_call_handler.rs
```diff
@@ -451,218 +449,171 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
-        Ok(U256::from(highest_gas_limit))
+        Ok(U256::from(range.highest))
     }
 
     pub fn last_constructed_block_context(&self) -> Option<BlockContext> {
         *self.last_constructed_block_context.borrow()
     }
 }
 
+#[derive(Debug, Deref, Display)]
```
slowli: Bikeshedding: both `Display` and `Deref` derives look unnecessary. The `Display` one can be removed, and `Debug` (`{:?}`) specifier used instead of it in the relevant logs; `Deref` may be replaced with an inherent method:

```rust
impl Probe {
    fn value(self) -> u64 {
        match self {
            Self::Midpoint(value) | Self::Highest(value) | Self::Optimistic(value) => value,
        }
    }
}
```

lib/rpc/src/eth_call_handler.rs
```diff
@@ -451,218 +449,171 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
-            // search is small enough (less than 1.5% of the highest gas limit)
-            // <https://github.com/ethereum/go-ethereum/blob/a5a4fa7032bb248f5a7c40f4e8df2b131c4186a4/eth/gasestimator/gasestimator.go#L152
-            if (highest_gas_limit - lowest_gas_limit) as f64 / (highest_gas_limit as f64)
-                < ESTIMATE_GAS_ERROR_RATIO
-            {
+        // <https://github.com/ethereum/go-ethereum/blob/a5a4fa7032bb248f5a7c40f4e8df2b131c4186a4/eth/gasestimator/gasestimator.go#L152>
+        let mut mid = range.biased_midpoint();
+        while !range.is_empty() {
+            // Stop once error is within 1.5% of the highest gas limit.
+            if range.is_narrow_enough() {
```
slowli: Dumb question: Can't this be moved in the `while` condition above? (Or maybe replace it, since it seems strictly more lax than `is_empty()`)

lib/rpc/src/eth_call_handler.rs
```diff
@@ -451,218 +449,171 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+    fn is_empty(&self) -> bool {
+        self.lowest + 1 >= self.highest
+    }
+
+    fn is_narrow_enough(&self) -> bool {
+        (self.highest - self.lowest) as f64 / (self.highest as f64) < ESTIMATE_GAS_ERROR_RATIO
+    }
+
+    fn midpoint(&self) -> u64 {
+        ((self.highest as u128 + self.lowest as u128) / 2) as u64
```
slowli: Nit: AFAIU, it's possible to use [`u64::midpoint()`](https://doc.rust-lang.org/std/primitive.u64.html#method.midpoint) here.

lib/rpc/src/eth_call_handler.rs
```diff
@@ -451,218 +449,171 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
-            // An estimation error is allowed once the current gas limit range used in the binary
-            // search is small enough (less than 1.5% of the highest gas limit)
-            // <https://github.com/ethereum/go-ethereum/blob/a5a4fa7032bb248f5a7c40f4e8df2b131c4186a4/eth/gasestimator/gasestimator.go#L152
-            if (highest_gas_limit - lowest_gas_limit) as f64 / (highest_gas_limit as f64)
-                < ESTIMATE_GAS_ERROR_RATIO
-            {
+        // <https://github.com/ethereum/go-ethereum/blob/a5a4fa7032bb248f5a7c40f4e8df2b131c4186a4/eth/gasestimator/gasestimator.go#L152>
+        let mut mid = range.biased_midpoint();
+        while !range.is_empty() {
+            // Stop once error is within 1.5% of the highest gas limit.
```
slowli: Off-topic (?): In Era, gas estimation accuracy was configurable; is it deemed unnecessary? AFAIU, it'd be relatively easy to add the corresponding config param.

lib/rpc/src/eth_call_handler.rs
LogvinovLeon: We can't use apply_probe here as the logic is different. If the tx fails for OOG in other probes - that means that the gas limit is wrong. If it fails with highest - that means 4xx - invalid tx.

As for getting rid of it - I kinda like having all three probes we use defined in one place. I think it brings clarity. But no strong opinions here so can remove.

lib/rpc/src/eth_call_handler.rs
LogvinovLeon: Why would I add 7 lines of code and some tokens at the call site if deriving Deref already does the same thing?
The only reason I can think of is if Deref usage is confusing but `self.highest = *probe` seems clear to me.

lib/rpc/src/eth_call_handler.rs
LogvinovLeon: It would be easy but let's wait till someone actually requests it

lib/rpc/src/eth_call_handler.rs
LogvinovLeon: Display removed

slowli (dismissed):
LGTM. Please wait for a positive review from @itegulov though.

Cargo.lock
itegulov, line 1: Looks like you bumped a lot dependency versions here. Overall I don't mind but I'd prefer if we did it in a separate PR so it's easily revertable on its own.

itegulov (dismissed):
LGTM, just one nit

itegulov (approved):

PR #1269, opened 2026-05-08
feat(status): add pipeline status endpoint

antonbaliasnikov:
## Summary

Adds a pipeline diagnostics endpoint to the status server, on both the main node and external nodes.

**`GET /status/pipeline`**
Per-component pipeline state snapshot. Returns an ordered array of components, each with its current state (`idle`/`active` plus the component-specific state), last picked/processed block and batch coordinates, and lag to its upstream neighbor (`block_diff`, `batch_diff`).

Intended for diagnostics and dashboards to observe pipeline health and backpressure in detail.

Example element:

```json
{
  "name": "l1_sender_execute",
  "state": "active",
  "specific_state": "waiting_l1_inclusion",
  "picked": { "block_number": 2, "timestamp": 1783436859, "batch_number": 2 },
  "processed": { "block_number": 1, "timestamp": 1783436859, "batch_number": 1 },
  "lag": { "block_diff": 1, "batch_diff": 1 }
}
```

## Implementation notes

- Core types (`ComponentId`, `GenericComponentState`, `ComponentState`, `TrackingCoordinates`, `AdjacentSnapshot`) derive `Serialize` so they own their JSON representation; the endpoint only assembles them.
- Both pipelines now expose their `PipelineTracker` snapshot watch through `PipelineHandles`, so the endpoint works on external nodes too.
- `BatchSink` now reports the last block timestamp so the final stage carries time coordinates like every other stage.
- The existing `/status` and `/status/health` endpoints are unchanged.

Smoke-tested against a local chain (`run_local.sh`): the endpoint returns all 19 pipeline components with live state and lag.

### DIFF
--- a/lib/observability/src/generic_component_state.rs
+++ b/lib/observability/src/generic_component_state.rs
@@ -1,7 +1,9 @@
+use serde::Serialize;
 use vise::EncodeLabelValue;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelValue)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelValue, Serialize)]
 #[metrics(label = "state", rename_all = "snake_case")]
+#[serde(rename_all = "snake_case")]
 pub enum GenericComponentState {
     /// No work available — waiting for upstream.
     Idle,
--- a/lib/status/src/lib.rs
+++ b/lib/status/src/lib.rs
@@ -1,35 +1,44 @@
+//! Status HTTP endpoints.
+//!
+//! - `GET /status` — general node status, including consensus (Raft) state.
+//! - `GET /status/health` — liveness endpoint. Always 200 while the process is up.
+//! - `GET /status/pipeline` — per-component backpressure and lag snapshot for
+//!   diagnostics and dashboards.
+
 mod health;
+mod pipeline;
 mod status;
 
 use crate::health::health;
+use crate::pipeline::pipeline;
 use crate::status::status;
 use axum::{Router, routing::get};
 use reth_tasks::shutdown::GracefulShutdown;
 use tokio::{net::TcpListener, sync::watch};
+use zksync_os_backpressure::PipelineSnapshot;
 use zksync_os_raft::RaftConsensusStatus;
 
 pub use status::{ConsensusStatus, StatusResponse};
 
 #[derive(Clone)]
-struct AppState {
-    consensus_raft_status_rx: Option<watch::Receiver<Option<RaftConsensusStatus>>>,
+pub struct StatusServerState {
+    pub pipeline_snapshot: watch::Receiver<PipelineSnapshot>,
+    pub consensus_raft_status_rx: Option<watch::Receiver<Option<RaftConsensusStatus>>>,
 }
 
-// todo: handle graceful shutdown in a meaningful manner:
-//       we should start a timer for status server's lifetime, report healthy=false and only shutdown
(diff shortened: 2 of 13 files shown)


lib/status/src/lib.rs
```diff
@@ -1,35 +1,44 @@
+//! Status HTTP endpoints.
+//!
+//! - `GET /status` — general node status, including consensus (Raft) state.
```
LogvinovLeon, line 3: Let's think about the hierarchy here. Top-level status should either include the status of all components or just be the simple endpoint returning very simple overview while nested ones give more details.
I understand that changing it now is a breaking change for consumers but what we can do is:
* Return consensus status in `/status/consensus`
* Deprecate main endpoint
* Remove usages of top-level endpoint
* Remove/change top-level endpoint

Or if we don't have consumers of this endpoint's status key - we can just keep one bool there and move consensus data to /status/consensus in one change

lib/observability/src/generic_component_state.rs
```diff
@@ -1,7 +1,9 @@
+use serde::Serialize;
 use vise::EncodeLabelValue;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelValue)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EncodeLabelValue, Serialize)]
 #[metrics(label = "state", rename_all = "snake_case")]
+#[serde(rename_all = "lowercase")]
```
LogvinovLeon: `#[serde(rename_all = "lowercase")]` here, but the metrics label uses `snake_case` and the sibling `ComponentId` enum uses `snake_case` for serde too. Single-word variants (`Idle`/`Active`) hide it today, but a future multi-word variant would serialize differently in JSON vs metrics. Suggest `rename_all = "snake_case"` for parity.

node/bin/src/lib.rs
```diff
@@ -1439,7 +1450,10 @@ async fn run_en_pipeline(
     chain_id: u64,
     verify_batch_rx: tokio::sync::mpsc::Receiver<PeerVerifyBatch>,
     outgoing_verify_results: tokio::sync::broadcast::Sender<PeerVerifyBatchResult>,
-) -> watch::Receiver<TransactionAcceptanceState> {
+) -> (
+    watch::Receiver<TransactionAcceptanceState>,
+    watch::Receiver<PipelineSnapshot>,
+) {
```
LogvinovLeon: The MN path returns a named `PipelineHandles`, but `run_en_pipeline` now returns a bare `(watch::Receiver<TransactionAcceptanceState>, watch::Receiver<PipelineSnapshot>)`. A named struct would be more self-documenting and symmetric with the MN side.

lib/observability/src/generic_component_state.rs
antonbaliasnikov: Changed to `rename_all = "snake_case"` — no JSON change today (both variants are single-word), but it now matches the metrics label and `ComponentId`.

node/bin/src/lib.rs
antonbaliasnikov: `run_en_pipeline` now returns `PipelineHandles` like the MN path, with `prover_api_port: None` set inside; the call site no longer destructures and re-wraps.

LogvinovLeon (approved):

lib/status/src/lib.rs
antonbaliasnikov: It makes sense to me, but let's do this separately.

PR #1270, opened 2026-05-09
fix(gateway): validate funding wallet set

sidhujag:
## Summary
- Validate every wallet file selected for funding as a safe file inside the active `GATEWAY_DIR`.
- Fund root, chain, and explicit in-tree wallet files through one aggregate plan so funder balance is checked before any transaction is broadcast.
- Deduplicate recipients by address and use the highest required target across all wallet files.

## Test plan
- `bash -n scripts/gateway-launch/fund-wallets.sh scripts/gateway-launch/_common.sh`
- `git diff --check -- scripts/gateway-launch/fund-wallets.sh scripts/gateway-launch/_common.sh`
- Local mock-`cast` harness: aggregate insufficient balance fails before send; explicit wallet path outside `GATEWAY_DIR` is rejected.


sidhujag:
Closed: this branch is intended for syscoin/zksync-os-server main.

PR #1274, opened 2026-05-11
fix(l1-watcher): do not panic on non-existent migration 

itegulov:
## Summary

* Removes anvil-related hacks in all L1 watchers
* Makes `find_block_by_migration_number` not panic when next migration has not been activated yet on L1


kostia244:
Did something change on anvil side? Why can we do this now?

kostia244 (approved):

itegulov:
> Did something change on anvil side? Why can we do this now?

We use local state with historical states preserved. It's a lot heavier but gives us ability to finally get rid of it. This is partially why we wanted to have that JSON compression built-in.

PR #1277, opened 2026-05-11
feat: separate gateway sender config

itegulov:
## Summary

Adds new `GwSenderConfig` that is used when current SL interval is on Gateway instead of L1. This also unties weird assumption "gateway RPC provided -> we are settling on GW" that we used to rely on before this PR.


node/bin/src/lib.rs
```diff
@@ -222,25 +222,35 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
     }
 
-    if node_role.is_main() {
-        let pubdata_mode = config
-            .l1_sender_config
-            .pubdata_mode
-            .expect("l1_sender_pubdata_mode must be set on the Main Node");
+    // Effective pubdata mode used by all block-producing components: read from config only when
+    // the chain settles on L1. When settling on Gateway, it is always
+    // `PubdataMode::RelayedL2Calldata` regardless of what is configured.
```
perekopskiy: I think `PubdataMode::Validium` can also be used when settling on Gateway

perekopskiy (commented):
Overall looks good, left one comment. I assume "gateway RPC provided -> we are settling on GW" assumption for zks_getL2ToL1LogProof will be fixed in separate PR, right?

node/bin/src/config/mod.rs
```diff
@@ -865,6 +946,59 @@ fn is_positive_f64(&val: &f64) -> bool {
...
+/// Gateway sender configuration. Used by the L1Sender pipeline components when the chain is
+/// currently settling on a Gateway (as discovered from the L1 settlement layer interval at
+/// startup). When the chain is settling on L1 directly, this config is ignored and
+/// [`L1SenderConfig`] is used instead.
+///
+/// Operator keys here must be funded on the Gateway chain. They are optional at config-validation
+/// time; their presence is enforced at runtime once the settlement layer is discovered.
+#[derive(Clone, Debug, DescribeConfig, DeserializeConfig, ConfigValidate)]
+#[config(derive(Default))]
+pub struct GwSenderConfig {
```
slowli: - Bikeshedding: Subjectively, "gw" contraction looks superficial; it'd be acceptable to name the type `GatewaySenderConfig` and use the `gateway_sender` prefix for it.
- Just to be clear: Does this struct largely copy fields from `L1SenderConfig` as opposed to reusing them via extracting in a separate struct and using `#[config(flatten)]` because of differing documentation? Not that it creates much code duplication in this case, just want to understand the motivation better.

node/bin/src/lib.rs
```diff
@@ -1198,6 +1198,29 @@ async fn run_main_node_pipeline(
...
+            config.l1_sender_config.clone().into()
+        };
+    let prove_sender_config: zksync_os_l1_sender::config::L1SenderConfig<ProofCommand> =
+        if settles_on_gateway {
+            config.gw_sender_config.clone().into()
+        } else {
+            config.l1_sender_config.clone().into()
+        };
+    let execute_sender_config: zksync_os_l1_sender::config::L1SenderConfig<
+        zksync_os_l1_sender::commands::execute::ExecuteCommand,
```
slowli: Micro-nit: Why is this type fully qualified, whereas `CommitCommand` / `ProofCommand` above are imported? IMO, importing `ExecuteCommand` as well would make code more readable.

node/bin/src/config/mod.rs
itegulov: Yes, mostly differing documentation plus I did not want to create a common struct prematurely. There is no clean split there unfortunately as for example 2/3 of fee configuration is common but blob-related field only makes sense for L1 sender.

Not a strong opinion either way, having these extra `From`s and copy-pasted operator validation logic is far from perfect too.

node/bin/src/lib.rs
itegulov: Hmm, good point. I asked protocol to confirm for sure if this ever makes sense, will adjust accordingly

itegulov:
> I assume "gateway RPC provided -> we are settling on GW" assumption for zks_getL2ToL1LogProof will be fixed in separate PR, right?

@perekopskiy yes, will handle in a separate PR. This needs some changes to `SettlementLayerIntervals` that I'd rather not put into this PR.

node/bin/src/lib.rs
itegulov: Added pubdata mode detection using GW's DA input mode. It's unambiguous in Gateway's case

node/bin/src/config/mod.rs
```diff
@@ -64,6 +64,8 @@ pub struct Config {
     pub sequencer_config: SequencerConfig,
     #[config_validate(async_validate(Self::validate_operator_signers))]
     pub l1_sender_config: L1SenderConfig,
+    #[config_validate(async_validate(Self::validate_gw_operator_signers))]
+    pub gateway_sender_config: GatewaySenderConfig,
```
perekopskiy, line 68: worth adding to `report_static_config_metrics`

perekopskiy (dismissed):

perekopskiy (approved):

PR #1279, opened 2026-05-12
feat: replay record archive

perekopskiy:
## Summary

Adds replay record cold storage support so main node can archive replay records outside local RocksDB and later recover the node replay WAL from that archive.

This PR includes:
- A new zksync_os_replay_archive crate with append-only archive traits.
- Filesystem replay archive storage and reader implementations.
- Optional age X25519 encryption for archived replay records.
- A bounded replay archive component that accepts records from BlockApplier without blocking on storage writes unless the queue is full.
- Node config for selecting replay archive backend, defaulting to no-op.
- Replay archive gate before batch commit, ensuring replay records are present in cold storage before committing.
- Recovery tooling to download, decrypt, verify, and rebuild block_replay_wal from archived replay records using an anchor block number/hash.
- An end-to-end integration test covering encrypted archive, storage deletion, recovery, restart, and continued batch finalization.


Nexus2k (approved):
LGTM in terms of use and handling of encryption

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,655 @@
...
+/// The output layout is:
+///
+/// ```text
+/// <output_root>/<block_number>/<block_hash>/<session>
+/// ```
+///
+/// This keeps all session copies for the same replay record next to each other.
+pub async fn download_all_replay_archive_objects<Reader>(
+    reader: &Reader,
+    output_root: impl AsRef<Path>,
```
slowli: Bikeshedding: Using `impl AsRef<_>` looks like over-engineering (which potentially slows down compilation because of the necessity to retain generics and monomorphize them cross-crate). A simple `&Path` would work fine. Ditto for other functions.

lib/replay_archive/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+
+/// Session-bound byte storage using the session/block/hash layout.
+///
+/// Implementations must be append-only. Creating storage must create or mark exactly one session
+/// and must fail if that session already exists. Appending an object must fail if data already
+/// exists at `<session>/<block_number>/<block_hash>`, even if the stored value is byte-for-byte
+/// identical. This is intentionally stricter than idempotent object-store writes so that bugs do
+/// not silently replace archived replay data.
+#[async_trait]
+pub trait ReplayArchiveStorage: Sized {
```
slowli: - Bikeshedding: IMO, it makes sense to add `Send` requirement for all async traits (and `Sync` for ones with methods with a `&self` receiver) since otherwise the implementations would be unusable in Tokio.
- Are there plans to store anything else in the storage beside replay records? Otherwise, having separate low-level `ReplayArchiveStorage` and high-level `ReplayArchiver` traits looks like an overkill; one would suffice.

lib/replay_archive/src/reader.rs
```diff
@@ -0,0 +1,17 @@
...
+
+pub type ReplayArchiveObjectStream = BoxStream<'static, anyhow::Result<ReplayArchiveKey>>;
+
+/// Read-side access to replay archive objects.
+///
+/// Implementations should hide backend-specific path parsing and return normalized archive keys.
+#[async_trait]
+pub trait ReplayArchiveStorageReader {
+    /// Lists all stored replay archive objects.
+    async fn list_objects(&self) -> anyhow::Result<ReplayArchiveObjectStream>;
```
slowli: Bikeshedding: This signature looks weird; an "outer" error can be always rolled into the stream as its first element. (Further, the `FileSystemReplayArchiveReader` implementation never returns outer errors.)

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,655 @@
...
+        output_root = %output_root.display(),
+        "Starting replay archive object download"
+    );
+    let mut keys = reader.list_objects().await?;
+    let mut downloaded = 0;
+
+    while let Some(key) = keys.next().await {
+        let key = key?;
+        let object = reader
+            .read_object(&key)
```
slowli: Since this is the only use of `read_object`, could it make sense to unite both `ReplayArchiveStorageReader` methods so that e.g. `list_objects` returns a stream of *objects*, rather than object keys? OTOH, having two separate methods enables parallelizing object processing (which AFAIU is not implemented right now, but could be implemented with a reasonable effort).

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,655 @@
...
+
+    tracing::info!(downloaded, "Finished replay archive object download");
+    Ok(downloaded)
+}
+
+/// Decrypts downloaded replay archive objects into a separate local output directory.
+///
+/// The input is expected to use the layout created by [`download_all_replay_archive_objects`].
+/// The same relative layout is preserved under `output_root`.
+pub async fn decrypt_downloaded_replay_archive_objects(
```
slowli: Can't archive objects be decrypted on the fly (e.g., in `recover_replay_records_to_rocksdb` below)? Having a separate decryption step looks wasteful at least in terms of disk space (and maybe performance / time as well, especially if object processing is parallelized). For the same reason, a separate decryption CLI command looks redundant IMO; I'd suggest to have an optional `--key-path` arg for the `RecoverRocksdb` command instead.

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,655 @@
...
+    for (block_number, block_hash) in canonical_chain {
+        let replay_record = read_verified_replay_record(input_root, block_number, block_hash)
+            .await
+            .with_context(|| {
+                format!("failed to read replay record #{block_number}, {block_hash} for writing")
+            })?;
+        anyhow::ensure!(
+            replay_storage.write(Sealed::new_unchecked(replay_record, block_hash), false),
+            "replay record #{block_number} already exists in recovered RocksDB"
+        );
```
slowli, line 161: Won't this approach create problems with fault tolerance (i.e., if recovery is stopped / crashes for whatever reason in the middle, it won't be able to resume)? AFAIU, it'd be reasonably easy to add some degree of fault tolerance by extending the chain to the next block from one persisted in RocksDB.

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,655 @@
...
+/// Downloads every archive object into a local layout grouped by block number and block hash.
+///
+/// The output layout is:
+///
+/// ```text
+/// <output_root>/<block_number>/<block_hash>/<session>
+/// ```
+///
+/// This keeps all session copies for the same replay record next to each other.
+pub async fn download_all_replay_archive_objects<Reader>(
```
slowli, line 19: Maybe, I'm not getting the full picture (e.g., future non-local `ReplayArchiveStorageReader` extensions), but right now it looks quite weird that this function essentially just rearranges some files to create a directory index over (block number, block hash). There are other ways to create such an index – obvious ones for an in-memory index, but a persistent one could be reasonably easy to implement as well (e.g., an embedded relational DB like SQLite – which could store replay records as well, btw).

lib/replay_archive/src/replay_record.rs
```diff
@@ -0,0 +1,53 @@
...
+        block_number: BlockNumber,
+        block_hash: BlockHash,
+    ) -> anyhow::Result<bool> {
+        self.storage.contains_object(block_number, block_hash).await
+    }
+}
+
+pub(crate) fn encode_replay_record(replay_record: &ReplayRecord) -> anyhow::Result<Vec<u8>> {
+    serde_json::to_vec(replay_record).context("failed to encode replay record")
+}
```
slowli: Nit: Returning a `Result` looks unnecessarily complicated in this case; `serde` serialization only fails in fringe cases (e.g., on recursive data structures).

lib/replay_archive/src/init.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use crate::{
+    AgeEncryptedReplayArchiver, FileSystemReplayArchiveStorage, FileSystemReplayArchiver,
+    ReplayArchiveComponent, ReplayArchiveSender, ReplayArchiveSession, ReplayArchiveStorage,
+    ReplayArchiver,
+};
+
+#[derive(Debug, Clone)]
+pub enum ReplayArchiveConfig {
+    Noop {},
```
slowli: Nit: Braces look redundant; ditto below in `ReplayArchiveEncryptionConfig`.

lib/replay_archive/src/age_encrypted.rs
```diff
@@ -0,0 +1,83 @@
...
+        Ok(Self::new(storage, recipient))
+    }
+
+    pub fn storage(&self) -> &Storage {
+        &self.storage
+    }
+
+    pub fn recipient(&self) -> &age::x25519::Recipient {
+        &self.recipient
+    }
```
slowli: These methods are unused.

lib/replay_archive/src/age_encrypted.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    pub fn encrypt_replay_record(&self, replay_record: &ReplayRecord) -> anyhow::Result<Vec<u8>> {
```
slowli: Nit: This doesn't need to be public.

lib/replay_archive/src/age_encrypted.rs
```diff
@@ -0,0 +1,83 @@
...
+        let encrypted = age::encrypt(&self.recipient, encoded.as_slice())
+            .context("failed to encrypt replay record with age X25519")?;
+        let elapsed = started_at.elapsed();
+        REPLAY_ARCHIVE_METRICS.encryption_time.observe(elapsed);
+        if encoded_len > 0 {
+            REPLAY_ARCHIVE_METRICS
+                .encryption_time_per_megabyte
+                .observe(elapsed.as_secs_f64() * BYTES_PER_MEGABYTE / encoded_len as f64);
+        }
+        REPLAY_ARCHIVE_METRICS.object_bytes[&"stored"].observe(encrypted.len());
```
slowli, line 49: Wouldn't this metric be reported by the `Storage` that this archiver delegates to?

lib/replay_archive/src/age_encrypted.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    pub fn encrypt_replay_record(&self, replay_record: &ReplayRecord) -> anyhow::Result<Vec<u8>> {
+        let encoded = encode_replay_record(replay_record)?;
+        let encoded_len = encoded.len();
+        REPLAY_ARCHIVE_METRICS.object_bytes[&"plaintext"].observe(encoded_len);
+
+        let started_at = Instant::now();
+        let encrypted = age::encrypt(&self.recipient, encoded.as_slice())
+            .context("failed to encrypt replay record with age X25519")?;
```
slowli, line 41: Don't quite get when this call can fail (it seems that calls used internally in `age::encrypt` only emit I/O errors, which are impossible with the `&mut Vec<u8>` writer). Strictly speaking, this is an issue with the library though 🤷‍♂️

lib/batch_types/src/batcher_model.rs
```diff
@@ -35,6 +35,7 @@ pub struct BatchMetadata {
     pub blob_sidecar: Option<BlobTransactionSidecar>,
     pub first_block_number: u64,
     pub last_block_number: u64,
+    pub last_block_hash: B256,
```
slowli: Isn't this a breaking change that the unit test below guard against? 🙃 As this value is only read in `ReplayArchiveGateComponent`, isn't there another way to get a block hash by its number there?

lib/replay_archive/src/filesystem/storage.rs
```diff
@@ -0,0 +1,185 @@
...
+        self.inner.storage()
+    }
+
+    pub fn root_path(&self) -> &Path {
+        self.storage().root_path()
+    }
+
+    pub fn session(&self) -> &ReplayArchiveSession {
+        self.storage().session()
+    }
```
slowli: Nit: AFAICT, these getters are unused.

lib/sequencer/src/execution/block_applier.rs
```diff
@@ -64,10 +69,27 @@ where
             state_reporter.enter_state(BlockApplierState::AddingToStorage);
             tracing::info!(block_number, "Persisting block {block_number}");
             self.replay.write(
-                Sealed::new_unchecked(executed_replay.clone(), block_output.header.hash()),
+                Sealed::new_unchecked(executed_replay.clone(), block_hash),
                 override_allowed,
             );
 
+            if let Some(replay_archive_sender) = &self.replay_archive_sender {
```
slowli: Maybe I'm missing something obvious about the codebase, but I'd at least try to fold this functionality into a `WriteReplay` implementation (e.g., as a decorator).

lib/batch_types/src/batcher_model.rs
perekopskiy: Indeed, for some reason I thought that we no longer save full batch envelope but just proofs, but it's not the case. I'd rather keep optional `last_block_hash` to not introduce dependency on repository to read hash from there

lib/replay_archive/src/age_encrypted.rs
perekopskiy: agree, anyway it's not a big deal at all

lib/replay_archive/src/recovery.rs
perekopskiy: agree, I thought it would be easier to follow when steps are split, but looking at it now, it makes no sense

lib/replay_archive/src/recovery.rs
perekopskiy: You're right, although I'd like to keep fault tolerance out of the scope for this PR. I will keep this line as it kinda states that fault tolerance is not supported at the moment and will remove it separately

lib/replay_archive/src/recovery.rs
perekopskiy: My reasoning behind `download_all_replay_archive_objects` is that it's a simple solution + that it doesn't set a lower bound for RAM needed (which in-memory index does). IMO it's okay to keep it as is since we don't really care much about disk space requirement for a one-time recovery operation, moreover it will be negligible compared to external storage pricing.

lib/replay_archive/src/lib.rs
perekopskiy: `ReplayArchiveStorage`, `ReplayArchiver` split is needed so I can implement `ReplayArchiver` for `AgeEncryptedReplayArchiver` over generic storage

lib/replay_archive/src/age_encrypted.rs
perekopskiy: no, metric is reported by `AgeEncryptedReplayArchiver` and `ReplayRecordArchiver` which are disjoint -- basically these are archivers with encryption and without encryption

lib/sequencer/src/execution/block_applier.rs
perekopskiy: `WriteReplay` is sync so it's a bit problematic, we'll need `tokio::task::block_in_place` which I think we should avoid

perekopskiy: actually, I think it's worth making it async for this case, I changed this part

lib/replay_archive/src/age_encrypted.rs
slowli: With an infallible`age::encrypt`, this method could be made infallible as well, which logically makes sense. Not a big deal, as you've said.

lib/replay_archive/src/component.rs
```diff
@@ -0,0 +1,42 @@
...
+{
+    pub fn new(archive: Archive) -> (ReplayArchiveSender, Self) {
+        let (sender, records) = mpsc::channel(REPLAY_ARCHIVE_QUEUE_SIZE);
+        (sender, Self { archive, records })
+    }
+
+    pub async fn run(mut self) -> anyhow::Result<()> {
+        while let Some((block_hash, replay_record)) = self.records.recv().await {
+            let block_number = replay_record.block_context.block_number;
+            tracing::info!("Archiving replay record for block #{block_number}, {block_hash}");
```
slowli, line 33: Nit: Does this really merit `INFO` log level? Wouldn't it risk spamming logs?

lib/replay_archive/src/metrics.rs
```diff
@@ -0,0 +1,30 @@
+use std::time::Duration;
+use vise::{Buckets, Gauge, Histogram, LabeledFamily, Metrics, Unit};
+
+const LATENCIES: Buckets = Buckets::exponential(0.00001..=10.0, 10.0);
+const BYTES: Buckets = Buckets::exponential(1.0..=128.0 * 1024.0 * 1024.0, 2.0);
+const SECONDS_PER_MEGABYTE: Buckets = Buckets::exponential(0.0001..=1000.0, 10.0);
+
+#[derive(Debug, Metrics)]
+#[metrics(prefix = "replay_archive")]
+pub(crate) struct ReplayArchiveMetrics {
```
slowli, line 10: Just to understand better: Is it correct that a gauge with the last archived block is unnecessary because it's reported via pipeline metrics?

lib/replay_archive/src/recovery.rs
slowli: Just in case: an in-memory index doesn't necessarily need to store file *contents* in RAM, just a `(block number, block hash) -> file path` multi-map. But you're probably right that it's not an immediate concern.

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,778 @@
...
+
+    tracing::info!(downloaded, "Finished replay archive object download");
+    Ok(downloaded)
+}
+
+/// Decrypts downloaded replay archive objects into a separate local output directory.
+///
+/// The input is expected to use the layout created by [`download_all_replay_archive_objects`].
+/// The same relative layout is preserved under `output_root`.
+pub async fn decrypt_downloaded_replay_archive_objects(
```
slowli: IIUC, this function can be removed now.

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,778 @@
...
+            format!(
+                "failed to read replay archive record file {}",
+                entry.path().display()
+            )
+        })?;
+        let record = decoder.decode(record_bytes, &entry.path())?;
+        if let Some(canonical_record) = &canonical_record {
+            anyhow::ensure!(
+                canonical_record == &record,
+                "replay archive record {} differs from another session copy for block #{block_number}, {block_hash}",
```
slowli: Nit: Could make sense to stash the "canonical" file path (or at least session ID) and display it in this error message for easier investigation.

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,778 @@
...
+        format!(
+            "failed to flush decrypted replay archive object {}",
+            output_path.display()
+        )
+    })?;
+    Ok(())
+}
+
+async fn read_age_x25519_identity(identity_file: &Path) -> anyhow::Result<age::x25519::Identity> {
+    let contents = tokio::fs::read_to_string(identity_file)
```
slowli: Nit: This has the potential to read a lot of unnecessary data / take some time if an incorrect path is specified by mistake. As an alternative, it could be possible to use [`AsyncBufReadExt::lines()`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncBufReadExt.html#method.lines).

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,778 @@
...
+            )
+        })?;
+    for line in contents.lines() {
+        let line = line.trim();
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+        return line
+            .parse()
+            .map_err(|err| anyhow::anyhow!("failed to parse age X25519 identity: {err}"));
```
slowli: Is this OK that this logic will ignore any possible garbage after the first non-empty / non-comment line? I guess the probability of a file accidentally containing a valid identity secret is extraordinarily low, so this isn't something to worry about in practice.

lib/replay_archive/src/bin/replay_archive_recovery.rs
```diff
@@ -0,0 +1,79 @@
...
+        replay_db_path: PathBuf,
+        /// Anchor block number to recover from.
+        #[arg(long)]
+        anchor_block_number: u64,
+        /// Canonical anchor block hash.
+        #[arg(long)]
+        anchor_block_hash: alloy::primitives::BlockHash,
+        /// age identity file containing AGE-SECRET-KEY. If provided, records are decrypted in memory.
+        #[arg(long)]
+        identity_file: Option<PathBuf>,
```
slowli, line 42: Out-of-context question: Is there a reason to allow only file-based identity secrets in particular (as opposed to reading them directly, e.g. from an env var via `#[arg(long, env = _)]`)?

lib/replay_archive/src/write_replay.rs
```diff
@@ -0,0 +1,79 @@
...
+
+impl<Replay> WriteReplay for ReplayArchivingWriteReplay<Replay>
+where
+    Replay: WriteReplay,
+{
+    async fn write(&self, record: Sealed<ReplayRecord>, override_allowed: bool) -> bool {
+        let (replay_record, block_hash) = record.clone().split();
+        let written = self.replay.write(record, override_allowed).await;
+
+        if let Some(archive_sender) = &self.archive_sender {
```
slowli, line 63: Should `written` be factored into this condition (`if written && let ..`)? Or is it intentional that a record may be archived multiple times (across different sessions, I guess)?

lib/replay_archive/src/write_replay.rs
```diff
@@ -0,0 +1,79 @@
...
+                .set(replay_archive_queue_depth(archive_sender));
+            let started_at = Instant::now();
+            let send_result = archive_sender.send((block_hash, replay_record)).await;
+            REPLAY_ARCHIVE_METRICS
+                .enqueue_latency
+                .observe(started_at.elapsed());
+            REPLAY_ARCHIVE_METRICS
+                .queue_depth
+                .set(replay_archive_queue_depth(archive_sender));
+            send_result.expect("replay archive component stopped before accepting replay record");
```
slowli: Can't the archive component stop during shutdown? If so, it really doesn't make sense to spuriously panic here (but I'm not sure how this case can be checked). Further, even if the component does fail (e.g., the node runs out of disk space), does it make sense to treat it as a critical error? AFAIU, the gate component would eventually halt block production in this case; maybe, this (+ error logs / metrics) would be enough?

lib/storage_api/src/replay.rs
```diff
@@ -178,5 +178,9 @@ pub trait WriteReplay: ReadReplay {
     ///   all [`ReadReplay`] methods should reflect its existence appropriately
     /// * MUST be atomic and always leave storage in a valid state (that satisfies all requirements
     ///   here and in [`ReadReplay`]) regardless of the method's outcome (including panic)
-    fn write(&self, record: Sealed<ReplayRecord>, override_allowed: bool) -> bool;
+    fn write(
+        &self,
+        record: Sealed<ReplayRecord>,
+        override_allowed: bool,
+    ) -> impl Future<Output = bool> + Send;
```
slowli: Future work: Using async here makes the "main" RocksDB implementation sloppy because it uses blocking I/O despite being async (before, it *was* still used in async contexts, but at least the API surface level itself was sound). AFAIU, it'd be relatively easy to wrap blocking I/O in `spawn_blocking()`.

lib/replay_archive/src/gate_component.rs
```diff
@@ -0,0 +1,118 @@
...
+    ReplayStorage: ReadReplay,
+{
+    pub fn new(archive: Archive, replay_storage: ReplayStorage) -> Self {
+        Self {
+            archive,
+            replay_storage,
+        }
+    }
+
+    pub async fn wait_for_archive_to_contain_block(
```
slowli: Nit: This method doesn't need to be public.

lib/storage/src/db/replay.rs
```diff
@@ -477,11 +492,23 @@ impl ReadReplay for BlockReplayStorage {
...
+                block_context.block_number, 0,
+                "tried to append first replay record with non-zero block number: {}",
+                block_context.block_number
+            );
+            self.write_replay_unchecked(sealed_record, true);
+            latency_observer.observe();
+            return true;
+        }
+
+        let current_latest_record = current_latest_record.expect("checked above");
```
slowli: Nit: It'd be more idiomatic to use let-else here, moving the `if current_latest_record.is_none()` block above into `else`.

lib/replay_archive/src/recovery.rs
perekopskiy: yes, it's ok

lib/replay_archive/src/write_replay.rs
perekopskiy: It's intentional, we archive every processed replay record for redundancy

lib/replay_archive/src/write_replay.rs
perekopskiy: it is a critical error, but indeed we shouldn't panic but gracefully stop in this case

lib/storage_api/src/replay.rs
perekopskiy: we should be careful with it, spawn_blocking can add non-negligible overhead

lib/replay_archive/src/component.rs
perekopskiy: no, we are fine with short per-block logs

lib/replay_archive/src/metrics.rs
perekopskiy: I think no because archiver is not part of pipeline (but not sure if gauge would be reported if archiver was in pipeline); I added metric

lib/replay_archive/src/bin/replay_archive_recovery.rs
perekopskiy: not really, added a way to read key directly

lib/replay_archive/src/write_replay.rs
```diff
@@ -0,0 +1,86 @@
...
+where
+    Replay: WriteReplay,
+{
+    async fn write(
+        &self,
+        record: Sealed<ReplayRecord>,
+        override_allowed: bool,
+    ) -> anyhow::Result<bool> {
+        let (replay_record, block_hash) = record.clone().split();
+        let written = self.replay.write(record, override_allowed).await;
```
slowli: Should errors be propagated early here? I understand that this isn't a concern now (the only other implementation of `WriteReplay` never errors – it panics on I/O errors and the like), but still, what should the behavior be if it would propagate errors?

lib/replay_archive/src/recovery.rs
```diff
@@ -0,0 +1,624 @@
...
+) -> anyhow::Result<usize> {
+    tracing::info!(
+        input_root = %input_root.display(),
+        replay_db_path = %replay_db_path.display(),
+        anchor_block_number,
+        %anchor_block_hash,
+        "Starting replay archive RocksDB recovery"
+    );
+    if identity.is_some() {
+        tracing::info!("Replay archive RocksDB recovery will decrypt objects in memory");
```
slowli: Nit: Could make sense to log `identity.to_public()` here.

slowli (dismissed):

lib/replay_archive/src/write_replay.rs
perekopskiy: Indeed, there is no reason not to propagate early

RomanBrodetski (approved):

PR #1281, opened 2026-05-13
feat(watcher): shared l1/gw polling

kostia244:
## Summary

Currently we do a lot of redundant calls to `eth_blockNumber` and `eth_getBlockByNumber`/`eth_getHeaderByNumber`.
Every watcher gets a handle with latest & last finalized blocks. This way they don't need poll the same values.

This PR would reduce the count at least 3x.
This is helpful for all nodes, but is mostly directed at employee ENs. We need to reduce RPC cost fast. And this addresses the biggest waste of tokens.


lib/l1_watcher/src/watcher_cache.rs
```diff
@@ -0,0 +1,128 @@
...
+pub enum BlockBoundary {
+    Confirmed { confirmations: BlockNumber },
+    Finalized,
+}
+
+/// Used to track changes & notify watchers.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct ChainHead {
+    pub latest_block: BlockNumber,
+    pub finalized_block: Option<BlockNumber>,
```
perekopskiy: let's get rid of option for finalized, get_block_number_by_id(BlockId::finalized()) should not return None

kostia244: Good suggestion, done.

perekopskiy (dismissed):

lib/l1_watcher/src/watcher_cache.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct ChainHead {
+    pub latest_block: BlockNumber,
+    pub finalized_block: BlockNumber,
+}
+
+/// Used for reading L1 data which might be used by more than one watcher.
+/// Currently only block numbers.
+#[derive(Clone, Debug)]
+pub struct WatcherCache {
```
EmilLuta: Isn't this name misleading? You don't really cache anything. Maybe I'd call this `SharedL1Watcher` or something similar? In my head, in a cache I'd expect a hashmap that gives me access to it. (not suggesting we have a hashmap, just saying that `Cache` seems a inappropriate)

lib/l1_watcher/src/watcher_cache.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    async fn poll(&self) -> alloy::transports::TransportResult<()> {
+        let latest_block = self.provider.get_block_number().await?;
+        let finalized_block = self
+            .provider
+            .get_block_number_by_id(BlockId::finalized())
+            .await?
+            .expect("no finalized blocks yet");
```
EmilLuta: Correct me if I'm wrong, but doesn't this mean that all watchers now expect finalized blocks? This changes the semantics of the previous code behavior (i.e. we do a refactor, but we also change logic). I believe a bunch of watchers don't care that much about finalized blocks (although, I could be wrong).

EmilLuta (commented):
I think this PR changes the way watcher fails. Before, confirmed watcher only needed `eth_getBlockNumber`.  Now they also need to be finalized. Is that on purpose? The `.expect` there is a bit scary, as it will most likely panic (maybe it's not a problem, I don't know the surrounding code good enough).

On the design front, it feels like `WatcherCache` (don't like the name, but treat it as nit) does too many things. (i.e. does it make sense for the Watcher to have access to the sender; they don't need send capability, looks a bit like a code smell)

I would maybe split this into 2 parts -- tracker (can keep provider and Sender) & Handle (should have only Receiver).

And then the L1Watcher could contian the Prover and Handler (i.e. receiver). I didn't try to implement it, hope it makes sense (there might be reasons why this is not feasible in this codebase, again, I don't know the codebase very well).

Sidenote -- I love the shared polling idea!!!

lib/l1_watcher/src/watcher_cache.rs
kostia244: This call does not wait for anything to be finalized, it just gets the last finalized block's number.
The watchers can choose to read it or not, this behavior is not modified.
Two things that DO change:
1. The number of empty polls.
- Previously the watchers would wake up every second - 90% of time no new blocks are there.
- Now if a finalized block updates, a regular watcher will wake up. But it will see there are no new blocks and stop.
2. Point of failure(we will do this even if nobody asked for finalized block info):
- This specific L1 RPC call+argument combination was added together with PoS fork, the only reason it fails is broken infra.
- When settling on gateway: This can only fail if we deploy a chain before the gateway has an executed block, which means that the execute transaction for genesis block has not yet landed on gateway. 

I think the gateway case might be worth addressing, maybe someone will need this for testing, otherwise I think it's fine. Thanks for bringing this up.

kostia244: @perekopskiy You suggested to get rid of Option here, how should we handle the case where gateway has no finalized blocks, but no watcher needs it?

kostia244:
I wanted to have caching for events and polling for block numbers in one place eventually, hence watchercache. I thought about it and it doesn't bring us much benefit to do that. So I simplified it to pass just the channel to watchers.

lib/l1_watcher/src/watcher_cache.rs
kostia244: AFAIU, this does not matter, because when we start main node we fetch **finalized** state from SL anyway. So if there are no finalized blocks at this point. It means something went wrong. So it should be fine this way.

perekopskiy: in reth and our node finalized is always Some, so getting None for get_block_number_by_id(BlockId::finalized()) is really unexpected, None is generally returned if you ask for block by hash and it's not there and also pending can be None, but not finalized

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -78,6 +83,10 @@ impl<BatchStorage: WriteBatch> L1PersistBatchWatcher<BatchStorage> {
             }
 
             let zk_chain = &interval.proxy;
+            let block_updates = match &interval.settlement_layer {
+                IntervalSettlementLayer::L1 => l1_block_updates.clone(),
+                IntervalSettlementLayer::Gateway(_) => sl_block_updates.clone(),
```
perekopskiy: This looks incorrect, current SL can already be L1, we need explicitly gw_block_updates here

kostia244: Thanks for catching this. I think I fixed this as well as similar bug in interop_watcher.
How hard is it to test this case?

perekopskiy: it's hard, we need gateway->l1 migration, which we didn't test yet

perekopskiy (dismissed):

lib/l1_watcher/src/persist_batch_watcher.rs
itegulov: There is an integration test for migration in the new repo: https://github.com/matter-labs/zksync-os-integration-tests

perekopskiy (approved):

PR #1284, opened 2026-05-14
feat(observability): jemalloc metrics

kostia244:
## Summary

Add basic jemalloc metrics to make troubleshooting in the future easier. Note: metrics are only reported on unix based systems.
Do you think this is enough or should we get more detailed information?


EmilLuta (approved):

EmilLuta:
I believe I have all I need to troubleshoot memory on this PR. Thanks!

PR #1285, opened 2026-05-14
feat(l1-watcher): traverse historical GW intervals

itegulov:
## Summary

Makes `InteropWatcher` start even if current SL is not gateway. It then traverses historical intervals and fetches interop roots from GW.

### DIFF
--- a/lib/l1_watcher/src/interop_watcher.rs
+++ b/lib/l1_watcher/src/interop_watcher.rs
@@ -2,61 +2,131 @@ use alloy::primitives::ruint::FromUintError;
 use alloy::providers::DynProvider;
 use alloy::rpc::types::{Log, Topic};
 use alloy::sol_types::SolEvent;
+use anyhow::Context;
 use std::collections::HashMap;
 use zksync_os_contract_interface::Bridgehub;
 use zksync_os_contract_interface::IMessageRoot::NewInteropRoot;
 use zksync_os_contract_interface::InteropRoot;
+use zksync_os_contract_interface::l1_discovery::L2_BRIDGEHUB_ADDRESS;
+use zksync_os_contract_interface::settlement_layer_intervals::{
+    IntervalSettlementLayer, SettlementLayerIntervals,
+};
 use zksync_os_mempool::subpools::interop_roots::InteropRootsSubpool;
 use zksync_os_types::IndexedInteropRoot;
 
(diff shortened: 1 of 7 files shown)


lib/l1_watcher/src/interop_watcher.rs
```diff
@@ -2,61 +2,131 @@ use alloy::primitives::ruint::FromUintError;
...
+/// To support a chain that has migrated GW → L1 (or GW → L1 → GW → …), the watcher walks every
+/// Gateway interval — historical and active — via [`SlAwareL1Watcher`]. Each historical Gateway
+/// segment is bounded by the L1/SL block where the last included interop root was emitted.
 pub struct InteropWatcher {
     starting_interop_root_id: u64,
     interop_roots_subpool: InteropRootsSubpool,
 }
 
 impl InteropWatcher {
+    /// Builds the watcher iff the chain has at least one Gateway interval. Returns `Ok(None)`
```
Artemka374:
```suggestion
    /// Builds the watcher if the chain has at least one Gateway interval. Returns `Ok(None)`
```

lib/l1_watcher/src/interop_watcher.rs
```diff
@@ -2,61 +2,131 @@ use alloy::primitives::ruint::FromUintError;
...
+            // Empty intervals are possible when a migration closes without committing anything
+            // (`first_batch > last_batch`). Nothing to scan.
+            if interval
+                .last_batch
+                .is_some_and(|lb| interval.first_batch > lb)
+            {
+                continue;
+            }
+
+            let gw_zk_chain = intervals.resolve_proxy(interval.first_batch)?.clone();
```
Artemka374: bikeshedding: maybe it makes sense to make intervals to return a tuple which contains a proxy? or put a proxy inside of this interval type, so we don't do this call separately

Artemka374 (dismissed):
LGTM, one question - AFAIU sl aware watcher will only work if gw and l1 rpc URLs are not changing the whole time the chain is run - is it not expected at all that chain might connect to different gateways over time?

itegulov:
> is it not expected at all that chain might connect to different gateways over time?

I also thought about this. Overall we can extend the system to support that but right now V31 limits number of migration to 2, i.e. you can only do round trip L1->GW->L1 once

Artemka374 (approved):

PR #1288, opened 2026-05-14
feat(rpc): add eth_fillTransaction

coffee-the-dev:
## Summary

Adds support for `eth_fillTransaction` to the ZKsync OS server RPC API.

`eth_fillTransaction` takes an unsigned transaction request, fills missing execution defaults, and returns both the EIP-2718 encoded raw transaction bytes and the filled transaction object. This is useful for wallet, relayer, and client flows that want the node to apply the same nonce, gas, fee, and chain defaults used by adjacent RPCs before the transaction is signed and submitted elsewhere.

The implementation follows the existing RPC stack instead of adding a parallel transaction-building path:

- Adds the `eth_fillTransaction` method to the `EthApi` RPC trait.
- Reuses `transaction_count_impl(..., pending)` for the default nonce so fill behavior matches `eth_getTransactionCount` with pending state.
- Reuses the existing ZKsync OS `gas_price_impl()` value for fee defaults so the method stays aligned with runtime `eth_gasPrice` semantics.
- Reuses the existing call handler transaction preparation path after filling request defaults.
- Estimates missing gas against the pending block, matching how clients expect fill/defaulting APIs to account for current pending state.

This improves parity with established Ethereum JSON-RPC implementations.

## Behavior Notes

The method mirrors reth more closely than geth where the implementations differ:

- `chainId` is set to the local chain ID during fill, matching reth's behavior of applying the local chain ID as a default. Geth rejects a mismatched caller-provided `chainId`; this PR does not add that stricter geth-only validation.
- The request input is normalized through Alloy's transaction request handling. This keeps compatibility with the legacy `data` field while preferring `input` when present, without adding geth's explicit mismatched `data`/`input` rejection.
- Missing `value` defaults to zero.
- Missing `nonce` defaults from pending transaction count.
- Missing `gas` is filled from pending gas estimation.
- Missing dynamic fee fields are filled using the ZKsync OS scaled `eth_gasPrice` path rather than reth's L1 base-fee lookup, because ZKsync OS fee semantics should remain consistent with this node's runtime RPC behavior.

Unsupported transaction families continue to use the existing server behavior:

- EIP-4844/blob transaction fields are rejected because they are not supported by this API path.
- EIP-7702 authorization lists are rejected because they are not supported by this API path.

## Works Cited

- reth `eth_fillTransaction` implementation: fills value, nonce, chain ID, gas, fee defaults, and returns `FillTransaction { raw, tx }`.
  https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-eth-api/src/helpers/transaction.rs

- geth `TransactionAPI.FillTransaction`: exposes the equivalent RPC and delegates defaulting to `TransactionArgs.setDefaults`.
  https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go

- geth `TransactionArgs.setDefaults`: documents defaulting behavior for nonce, gas, value, chain ID, fee fields, and legacy `data` / `input` handling.
  https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/transaction_args.go

## Validation

- `cargo check -p zksync_os_rpc_api -p zksync_os_rpc`

### DIFF
--- a/integration-tests/tests/rpc/fill_transaction.rs
+++ b/integration-tests/tests/rpc/fill_transaction.rs
@@ -1,0 +1,204 @@
+use alloy::consensus::{BlobTransactionSidecar, BlobTransactionSidecarVariant, Transaction};
+use alloy::eips::Encodable2718;
+use alloy::network::TransactionBuilder;
+use alloy::primitives::{Bytes, U128, U256, address};
+use alloy::providers::Provider;
+use alloy::rpc::types::{FillTransaction, TransactionRequest};
+use alloy::transports::TransportResult;
+use zksync_os_integration_tests::{CURRENT_TO_L1, TestEnvironment, Tester, test_multisetup};
+use zksync_os_server::config::FeeConfig;
+use zksync_os_types::ZkEnvelope;
+
+const FIXED_BASE_FEE: u128 = 100_000_000;
+const FILL_GAS_PRICE_SCALE_FACTOR: f64 = 2.0;
+const EXPECTED_FILL_GAS_PRICE: u128 = 200_000_000;
+
+async fn launch_with_fixed_gas_price(env: TestEnvironment) -> anyhow::Result<Tester> {
+    let mut config = env.default_config().await?;
+    config.fee_config = FeeConfig {
+        native_price_usd: 3e-9,
+        base_fee_override: Some(U128::from(FIXED_BASE_FEE)),
+        native_per_gas: 100,
+        pubdata_price_override: Some(U128::from(1_000_000u64)),
+        native_price_override: Some(U128::from(1_000_000u64)),
+        pubdata_price_cap: None,
+    };
+    config.rpc_config.gas_price_scale_factor = FILL_GAS_PRICE_SCALE_FACTOR;
+    env.launch(config).await
+}
+
+async fn fill_transaction(
+    tester: &Tester,
+    request: TransactionRequest,
+) -> TransportResult<FillTransaction<ZkEnvelope>> {
+    tester
+        .l2_provider
+        .client()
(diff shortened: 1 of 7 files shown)


coffee-the-dev:
Making a change to align with reth (still return filled transaction if insufficient funds for gas) and to add test coverage for the RPC method

lib/rpc/src/eth_call_handler.rs
```diff
@@ -293,6 +293,71 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
+                request.max_fee_per_gas = Some(max_fee_per_gas);
+            }
+            if request.max_fee_per_gas.unwrap_or_default() < tip {
+                return Err(CallFeesError::TipAboveFeeCap.into());
+            }
+        }
+
+        let max_fee_per_gas = request.max_fee_per_gas;
+        let mut tx = self
+            .prepare_execution_env(request, Some(BlockId::pending()), None)?
```
itegulov: There is a race condition between pending block here and the one before this (line 326). So you could use two different blocks for the same transactions, not sure if it meaningfully impacts the flow.

coffee-the-dev: I don't think this would matter in this case as the goal of the method is to take a partial transaction body (ie from, to, data) and fill it with missing values that gives us a full body that can be signed for submission and inclusion in the next block.

coffee-the-dev:
`pub(crate)` changes to `eth_call_handler` from #1220 made it possible to break this logic out to a separate handler instead of putting it inside eth_call_handler which feels cleaner implementation wise; rebased to fix the conflicts and extract this to its own handler

itegulov (approved):

PR #1289, opened 2026-05-14
fix(l1-watcher): adapt upgrade watcher to ServerNotifier semantics

zkzoomer:
## Summary

[L1UpgradeTxWatcher](https://github.com/matter-labs/zksync-os-server/blob/main/lib/l1_watcher/src/upgrade_tx_watcher.rs) treats `UpdateUpgradeTimestamp` as immediately actionable, but this can crash the node if no matching `NewUpgradeCutData` exists. This can happen as `ChainAdmin`'s `setUpgradeTimestamp` does **not** validate that the upgrade data exists when being called with the new protocol version.

To fix this issue, in v31 we have modified the `ServerNotifier`'s `setUpgradeTimestamp` to instead take the shape:
```solidity
function setUpgradeTimestamp(uint256 _chainId, uint256 _upgradeTimestamp) external onlyChainAdmin(_chainId) {
    if (_upgradeTimestamp == 0) {
        revert ZeroUpgradeTimestamp();
    }
    uint256 _oldProtocolVersion = chainTypeManager.getProtocolVersion(_chainId);
    if (chainTypeManager.upgradeCutHash(_oldProtocolVersion) == bytes32(0)) {
        revert CutDataForProtocolVersionNotAvailable(_oldProtocolVersion);
    }
    protocolVersionToUpgradeTimestamp[_chainId][_oldProtocolVersion] = _upgradeTimestamp;
    emit UpgradeTimestampUpdated(_chainId, _oldProtocolVersion, _upgradeTimestamp);
}
```
 Validating that at the time of it being called, there exists some cut data to upgrade from the current protocol version.

This PR fixes this issue by instead listening to `ServerNotifier`'s `UpgradeTimestampUpdated`.


lib/l1_watcher/src/upgrade_tx_watcher.rs
```diff
@@ -275,24 +300,56 @@ impl L1UpgradeTxWatcher {
...
+    /// a pre-V31 upgrade.
+    async fn find_new_protocol_version(
+        &self,
+        raw_old_protocol_version: U256,
+        start_block: u64,
+        end_block: u64,
+    ) -> anyhow::Result<U256> {
+        let filter = Filter::new()
+            .from_block(start_block)
+            .to_block(end_block)
```
itegulov: You are scanning 2.5M blocks here, correct? This will not fit the filter on any reasonable provider. You need to scan in chunks and respect `self.max_blocks_to_process`

zkzoomer: oopsie, resolved in 83ec2358e6dccfca9bd3396f342a2d1942ee4d8f

itegulov (approved):

PR #1291, opened 2026-05-15
feat(l1_watcher): timeout + heartbeat metric for silent stalls 

dimazhornyk:
## Summary

Adds three defenses against the L1 watcher silently wedging (priority_tx ingest stuck for on a leader that was alive and producing L2 blocks; recovered only on process restart):

- poll_iterations counter per event type, so a flat rate immediately identifies a wedged task vs. a healthy task with nothing to scan.
- poll_iteration_timeout wrapping each poll() call; on overflow the task panics and the critical-task supervisor recycles the process with fresh provider connections. Catches half-dead TCP / hung RPC calls of any kind. Default 600s.
- sl_wait_timeout bounding the previously-unbounded SL-visibility loop in L1TxWatcher::process_event. Default 300s.


lib/l1_watcher/src/watcher.rs
```diff
@@ -87,10 +91,27 @@ impl L1Watcher {
...
+                        event_name,
+                        timeout_secs = self.poll_iteration_timeout.as_secs(),
+                        next_block = self.next_block,
+                        "l1 watcher poll timed out, panicking to get restarted"
+                    );
+                    panic!(
+                        "l1 watcher {event_name} poll iteration timed out after {:?}",
+                        self.poll_iteration_timeout
+                    );
+                }
```
kostia244: Have you tested EN sync scenario? For example, the first poll of persist_batch_watcher will do all the blocks, and it will not finish in 10minutes I believe.

kostia244: Another thing to consider, in upgrade watcher:
```
        // Wait until the timestamp before sending the upgrade tx, so that it's immediately executable.
        // TODO: this will block the watcher, so if e.g. a timestamp is set far in the future, and then an event
        // to override it is emitted, we will not be able to process it.
        self.wait_until_timestamp(request.timestamp).await;
```

I can't make claims here, but this looks like it could _intentionally_ take more than 10 minutes.

lib/l1_watcher/src/metrics.rs
```diff
@@ -7,6 +7,9 @@ pub struct L1Metrics {
     pub most_recently_scanned_l1_block: LabeledFamily<&'static str, Gauge<BlockNumber>>,
     #[metrics(labels = ["event"])]
     pub events_loaded: LabeledFamily<&'static str, Counter>,
+    /// Heartbeat: incremented once per `poll()` call. A flat counter means the watcher is stuck.
+    #[metrics(labels = ["event"])]
+    pub poll_iterations: LabeledFamily<&'static str, Counter>,
```
kostia244, line 12: > Heartbeat: incremented once per `poll()` call. A flat counter means the watcher is stuck.

Is this not redundant with respect to `most_recently_scanned_l1_block`?
That one provides a very similar signal but with higher frequency? 
Do you see a situation where `most_recently_scanned_l1_block` is moving but we are genuinely stuck?

lib/l1_watcher/src/watcher.rs
itegulov: Agree with @kostia244 the semantics here do not allow for timeouts, maybe a better place would be in mempool where we block and wait for transaction to appear in mempool?

lib/l1_watcher/src/metrics.rs
dimazhornyk: My reasoning here was to clearly distinguish between a hanging request and a stale L1 node:
most_recently_scanned_l1_block allows you to see that the server didn't receive the new block, but it doesn't help understand if it is a connection-level problem or an L1 node problem

This info can be derived from the dashboard of the L1 node, but poll_iterations just provides that directly.
No strong opinion here, so we can remove it if you guys think it adds more noise than benefit

lib/l1_watcher/src/watcher.rs
dimazhornyk: Makes sense, thanks for pointing this out. 
I changed the timeout to be per iteration within `poll` rather than the `poll` itself. 

As for the upgrade path - the watcher inserts the upgrade tx in the `Waiting` state, which has the timeout built-in.

PR #1296, opened 2026-05-18
fix(raft): Back off unreachable peer retries

RomanBrodetski:
## Summary
- add a capped OpenRaft network backoff for unreachable Raft peers
- reduce repeated RPCError logs when a configured consensus peer is offline while preserving periodic retries

## Testing
- cargo fmt --all -- --check
- git diff --check

Note: cargo check -p zksync_os_raft --lib is currently blocked by an existing jsonrpsee macro compile error in lib/rpc_api/src/pubsub.rs before reaching the raft crate.

### DIFF
--- a/lib/raft/src/network.rs
+++ b/lib/raft/src/network.rs
@@ -1,7 +1,9 @@
 use async_trait::async_trait;
 use openraft::Config;
 use openraft::error::{Fatal, RPCError, RaftError, ReplicationClosed, StreamingError, Unreachable};
-use openraft::network::{RPCOption, RaftNetwork, RaftNetworkFactory as OpenraftNetworkFactory};
+use openraft::network::{
+    Backoff, RPCOption, RaftNetwork, RaftNetworkFactory as OpenraftNetworkFactory,
+};
 use openraft::raft::{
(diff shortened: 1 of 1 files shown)


lib/raft/src/network.rs
```diff
@@ -16,6 +18,14 @@ use zksync_os_consensus_types::{RaftNode, RaftTypeConfig, debug_display_raft_ent
 use zksync_os_network::raft::protocol::{RaftRequestHandler, RaftRouter};
 use zksync_os_network::raft::wire::{RaftRequest, RaftResponse};
 
+const UNREACHABLE_BACKOFF_STEPS: [Duration; 4] = [
+    Duration::from_secs(1),
+    Duration::from_secs(2),
+    Duration::from_secs(5),
+    Duration::from_secs(10),
+];
+const UNREACHABLE_BACKOFF_MAX: Duration = Duration::from_secs(10);
```
perekopskiy: nit: any reason not to drop it and use `UNREACHABLE_BACKOFF_STEPS.last().unwrap()` instead?

RomanBrodetski: Done in a26e8b63.

perekopskiy (approved):

PR #1297, opened 2026-05-18
feat(rpc): per-method and global JSON-RPC rate limiting

LogvinovLeon:
## Summary

- Adds a `RateLimiting<S>` RPC middleware layer (token bucket via `governor`) that sits inside `Monitoring` so rate-limited requests are counted in error metrics
- Config: `rpc.rate_limits` is a map of method name → max req/s; use `"*"` for a global cap applied before per-method limits; empty map (the default) means no limiting
- Rejected requests return EIP-1474 error `-32005` with `{"retryAfterMs": N}` in the error data field
- `Monitoring` generalised from `Monitoring<RpcService>` to `Monitoring<S>` to allow middleware composition

**Config example (YAML):**
```yaml
rpc:
  rate_limits:
    "*": 1000       # global cap across all methods
    eth_call: 100   # stricter limit for eth_call
    eth_getLogs: 50
```

**Config example (env var):**
```
RPC_RATE_LIMITS='*=1000,eth_call=100,eth_getLogs=50'
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -168,8 +168,8 @@ futures = "0.3"
 vise = "0.3.0"
 vise-exporter = "0.3.0"
 bincode = { version = "2.0.1", features = ["serde"] }
-smart-config = { git = "https://github.com/matter-labs/smart-config" }
-smart-config-commands = { git = "https://github.com/matter-labs/smart-config" }
+smart-config = { version = "0.4.0-pre.3" }
+smart-config-commands = { version = "0.4.0-pre.3" }
 thiserror = "2.0.12"
 tikv-jemallocator = { version = "0.6.1", features = [
     "override_allocator_on_supported_platforms",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,7 +75,7 @@ serde = { version = "1", features = ["derive"] }
 serde_with = "3.14.0"
 serde_json = "1"
 sentry = "0.43.0"
-regex = "1.11.3"
+regex = "1.12.3"
 roaring = "0.11.3"
 
 #zksync_os_interface = { path = "../zksync-os-interface/crates/interface" }
@@ -168,8 +168,8 @@ futures = "0.3"
 vise = "0.3.0"
 vise-exporter = "0.3.0"
 bincode = { version = "2.0.1", features = ["serde"] }
-smart-config = { version = "0.4.0-pre.2" }
-smart-config-commands = { version = "0.4.0-pre.2" }
+smart-config = { git = "https://github.com/matter-labs/smart-config" }
+smart-config-commands = { git = "https://github.com/matter-labs/smart-config" }
 thiserror = "2.0.12"
 tikv-jemallocator = { version = "0.6.1", features = [
     "override_allocator_on_supported_platforms",
@@ -197,6 +197,7 @@ criterion = "0.7.0"
 sha2 = "0.10.9"
 tower = "0.5.2"
 tower-http = "0.6.6"
+governor = "0.6"
 hyper = "1.6.0"
 semver = { version = "1", features = ["serde"] }
 num_cpus = "1.17.0"
--- a/lib/rpc/src/config.rs
+++ b/lib/rpc/src/config.rs
@@ -1,6 +1,25 @@
 use alloy::primitives::Address;
 use std::collections::HashSet;
+use std::num::NonZeroU32;
 use std::time::Duration;
+
+/// A per-method rate limit entry.
+#[derive(Clone, Debug)]
+pub struct RpcRateLimit {
+    /// Exact RPC method name, e.g. `"eth_call"`.
+    pub method: String,
+    /// Maximum number of requests per second across all callers combined.
+    pub requests_per_second: NonZeroU32,
+}
+
+impl From<(String, NonZeroU32)> for RpcRateLimit {
+    fn from((method, requests_per_second): (String, NonZeroU32)) -> Self {
+        Self {
+            method,
+            requests_per_second,
+        }
+    }
+}
 
 #[derive(Clone, Debug)]
 pub struct RpcConfig {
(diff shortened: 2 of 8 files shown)


RomanBrodetski:
Nice!
Will we have sufficient metrics? We'll get this in error metrics "for free", right? I think we'll want a spearaet chart in grafana that shows rate limiting, but maybe it's possible to do nicely with what we have..

RomanBrodetski:
and separately.. in some cases different requests to the same method can result in vastly differnt load (100x+). Is it possible to somehow limit by total CPU seconds spent? Or will it be a hell to track/enforce?

LogvinovLeon:
### Metrics
Currently those will be tracked as RPC errors. We track those by code.
We already have panels that show the rate of RPC errors by code. After it's merged - will add a separate for only those -32005 so that it's visible at a glance.

### Load differences
Agree but this should be handled in a different PR
In an ideal world we have:
* Global RPS limit
* Per-method RPS limits for some heavy methods (debug_traceBlock as 1/s)
* Global timeout (1s)
* Some method-specific limits (eth_getLogs when unconstrained - no more than 100 blocks)

node/bin/src/config/mod.rs
```diff
@@ -984,6 +984,12 @@ pub struct RpcConfig {
     /// because pubdata price increases or native price decreases in-between estimation and sequencing.
     #[config(default_t = 2.0)]
     pub estimate_gas_pubdata_price_factor: f64,
+
+    /// Per-method rate limits: map from RPC method name to max requests per second (across all
+    /// callers).  Use `"*"` for a global limit applied before per-method limits.  Methods absent
+    /// from the map are unrestricted; a value of 0 disables the limit.
+    #[config(default, with = Serde![*])]
```
slowli: I'd suggest to use a [`DelimitedEntries`](https://matter-labs.github.io/smart-config/smart_config/de/struct.DelimitedEntries.html) deserializer here so that the configuration may be specified more easily via env vars:

```shell
export RPC_RATE_LIMITS='* = 1000, eth_call = 100'
```

lib/rpc/src/config.rs
```diff
@@ -2,6 +2,24 @@ use alloy::primitives::Address;
 use std::collections::HashSet;
 use std::time::Duration;
 
+/// A per-method rate limit entry.
+#[derive(Clone, Debug)]
+pub struct RpcRateLimit {
+    /// Exact RPC method name, e.g. `"eth_call"`.
+    pub method: String,
+    /// Maximum number of requests per second across all callers combined.  Zero disables the limit.
```
slowli: Dumb question: Where is disabling the limit actually implemented? AFAIU, `build_limiters()` makes it so that `RpcRateLimit` structs with `requests_per_second == 0` are filtered out, so that they are effectively ignored (e.g., `* = 1000, eth_call = 0` spec is the same as `* = 1000`). IMO, this does not mean "disabling the limit"; I'd expect "disabling the limit" for `eth_call` mean that `eth_call`s are not subject to rate limiting *at all*.

As a related point, as it is implemented right now, a per-method limit cannot be effectively higher than the global one. So e.g. `* = 100, eth_call = 1000` doesn't work as written. Not sure whether this is a bug or a feature, but AFAIU, it *is* possible to implement limits in such a way that per-method limiting takes precedence over global limiting (i.e., global limiting is actually *fallback* limiting across all methods with unspecified limits).

Also, as a bikeshedding point, I'm against using 0 as a magic value; I'd rather use `null` or another dedicated string (`Inf`?) to make it more explicit; `* = 1000, eth_call = Inf` is more easily interpreted. (This could require a custom deserializer for values though.)

lib/rpc/src/rate_limit_middleware.rs
```diff
@@ -0,0 +1,126 @@
...
+                .as_millis()
+                .try_into()
+                .unwrap_or(u64::MAX)
+        });
+        let inner = self.inner.clone();
+        async move {
+            if let Some(ms) = retry_after_ms {
+                return MethodResponse::error(
+                    request.id.clone().into_owned(),
+                    rate_limited_err(ms),
```
slowli, line 105: Nit: If clients would actually adhere to the provided timeout, it could make sense to randomize it a bit in order to prevent a stampede of retries.

lib/rpc/src/config.rs
LogvinovLeon: This is intentional. Global limit should take precedence.
The way I intend this feature to be used is that we set some high global limit. And then go over some heavy methods and restrict those further. I.e.:
* GLOBAL - 5_000/s
* eth_getLogs - 100/s
* debug_traceBlock - 1/s

Added validation that local limits are less than the global limit

LogvinovLeon: Removed using zero as a null. Now the absence of value is unlimited (but the global limit still applies if set)

lib/rpc/src/rate_limit_middleware.rs
LogvinovLeon: Yes in theory and no in practice.
The window is 1 second. If the client receives the message - please wait 800ms - by the time it reaches the client (Sequencer -> LB -> Client machine) - there is usually enough natural randomness accumulated from the network.

node/bin/src/config/mod.rs
```diff
@@ -923,7 +926,7 @@ pub struct PolicyServiceConfig {
     pub auth_token: Option<SecretString>,
 }
 
-#[derive(Clone, Debug, DescribeConfig, DeserializeConfig)]
+#[derive(Clone, Debug, DescribeConfig, DeserializeConfig, ConfigValidate)]
```
slowli: FWIW, config validation is wired via `DescribeConfig` / `DeserializeConfig` and doesn't require `ConfigValidate` derivation.

Cargo.toml
```diff
@@ -168,8 +168,8 @@ futures = "0.3"
 vise = "0.3.0"
 vise-exporter = "0.3.0"
 bincode = { version = "2.0.1", features = ["serde"] }
-smart-config = { version = "0.4.0-pre.2" }
-smart-config-commands = { version = "0.4.0-pre.2" }
+smart-config = { git = "https://github.com/matter-labs/smart-config" }
+smart-config-commands = { git = "https://github.com/matter-labs/smart-config" }
```
slowli: Please pin a specific revision so that this doesn't break unexpectedly in the nearest future. Could make sense to cut a new `smart-config` release, maybe.

slowli (approved):

PR #1298, opened 2026-05-18
test: add integration tests for block rebuild with L1 revert

vasyl-ivanchuk:
## Summary

Adds two integration tests to `integration-tests/tests/node/rebuild.rs`.

**`rebuild_panics_if_from_block_is_already_committed`** — verifies that the node panics on startup when `block_rebuild.from_block` points to a block already committed on L1. Starts a node with the batcher enabled, waits for a committed batch, then restarts with `from_block = 1` and asserts the panic message contains `"rebuild_from_block must be > last_l1_committed_block"`.

**`rebuild_after_l1_revert_starts_successfully`** — verifies that after reverting committed L1 batches the node can restart in rebuild mode and process new transactions. Starts a node, waits for a committed but not executed batch, reverts it on L1, then restarts the same node with `block_rebuild.from_block = 1` (now valid since `last_l1_committed_block` is 0 after the revert) and confirms it accepts a new L2 transaction.

### DIFF
--- a/integration-tests/tests/node/rebuild.rs
+++ b/integration-tests/tests/node/rebuild.rs
@@ -347,6 +347,61 @@ async fn rebuild_after_emptying_historical_block_preserves_unrelated_l2_txs(
     Ok(())
 }
 
+/// Verifies that the node panics on startup when `block_rebuild.from_block` points to a block
+/// that is already committed on L1 (i.e. `from_block <= last_l1_committed_block`).
+///
+/// Scenario:
+///   1. Start a node with the batcher enabled and mine a few blocks until at least one batch
+///      is committed to L1.
+///   2. Restart with `block_rebuild.from_block = 1`, which is guaranteed to be within the
+///      already-committed range.
+///   3. Expect a fatal error containing "rebuild_from_block must be > last_l1_committed_block".
+#[test_multisetup([CURRENT_TO_L1])]
+#[test_runtime(flavor = "multi_thread")]
+async fn rebuild_panics_if_from_block_is_already_committed(
+    env: TestEnvironment,
+) -> anyhow::Result<()> {
+    let mut config = env.default_config().await?;
+    config.sequencer_config.block_time = Duration::from_millis(50);
+    let tester = env.launch(config).await?;
+
+    // Mine transactions until at least one batch is committed on L1.
+    wait_for_l1_state(&tester, "at least one batch committed on L1", |state| {
+        state.last_committed_batch >= 1
+    })
+    .await?;
+
+    // Block 1 is always within the committed range once any batch has been committed.
+    let mut restarted_config = tester.config().clone();
+    restarted_config.sequencer_config.block_rebuild = Some(RebuildBlocksConfig {
+        from_block: 1,
+        blocks_to_empty: vec![],
+        reset_timestamps: false,
+    });
+
+    // The assert! fires synchronously during node startup (before any background tasks are
+    // spawned), so it panics through `start_with_config`. Isolate it in a spawned task so
+    // the JoinError captures the panic instead of unwinding the test thread.
+    let stopped = tester.stop().await?;
+    let join_result =
+        tokio::task::spawn(async move { stopped.start_with_config(restarted_config).await }).await;
+
+    let join_err = join_result.expect_err("expected node startup to panic");
+    assert!(join_err.is_panic(), "expected a panic, got a cancellation");
+    let payload = join_err.into_panic();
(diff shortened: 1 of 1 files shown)


integration-tests/tests/node/rebuild.rs
```diff
@@ -222,3 +346,125 @@ async fn rebuild_after_emptying_historical_block_preserves_unrelated_l2_txs(
...
+    env: TestEnvironment,
+) -> anyhow::Result<()> {
+    let mut config = env.default_config().await?;
+    config.sequencer_config.block_time = Duration::from_millis(50);
+    let tester = env.launch(config).await?;
+
+    // Mine transactions until at least one batch is committed on L1.
+    wait_for_l1_state(&tester, "at least one batch committed on L1", |state| {
+        state.last_committed_batch >= 1
+    })
```
RomanBrodetski, line 371: the alternative would be to wait for the l2 block to be marked as committed on l1 on l2 RPC. Not 100% sure what's cleaner.
@itegulov if you have opinion, please raise - otherwise we can just laave as is

integration-tests/tests/node/rebuild.rs
```diff
@@ -222,3 +346,125 @@ async fn rebuild_after_emptying_historical_block_preserves_unrelated_l2_txs(
...
+                .with_value(U256::from(1u64)),
+        )
+        .await?
+        .expect_successful_receipt()
+        .await?;
+
+    let committed_state = wait_for_l1_state(
+        &tester,
+        "a committed but not yet executed batch on L1",
+        |state| state.last_committed_batch >= 1 && state.last_executed_batch == 0,
```
RomanBrodetski, line 445: nit: why do we have `&& state.last_executed_batch == 0,` here but not in the other test?

integration-tests/tests/node/rebuild.rs
```diff
@@ -222,3 +346,125 @@ async fn rebuild_after_emptying_historical_block_preserves_unrelated_l2_txs(
...
+    tester
+        .l2_provider
+        .send_transaction(
+            TransactionRequest::default()
+                .with_to(Address::random())
+                .with_value(U256::from(1u64)),
+        )
+        .await?
+        .expect_successful_receipt()
+        .await?;
```
RomanBrodetski, line 438: nit: why do we do a test transaction here but not in the prev. test?

integration-tests/tests/node/rebuild.rs
```diff
@@ -222,3 +346,125 @@ async fn rebuild_after_emptying_historical_block_preserves_unrelated_l2_txs(
...
+    restarted
+        .l2_provider
+        .send_transaction(
+            TransactionRequest::default()
+                .with_to(Address::random())
+                .with_value(U256::from(1u64)),
+        )
+        .await?
+        .expect_successful_receipt()
+        .await?;
```
RomanBrodetski, line 495: It would be good to also check that server is able to commit a (new, updated) batch on L1. 

Right now we only check that transactions is accepted. In fact, we don't even validate that the l2 block was reverted (albeit admittedly other tests ensure that `block_rebuild` config is respected).

I would make sure that we are able to send a new transaction - and that server commits a different batch with the same number as the batch that was earlier reverted on chain.

integration-tests/tests/node/rebuild.rs
vasyl-ivanchuk: In `rebuild_after_emptying_historical_block_preserves_unrelated_l2_txs`, the batcher is disabled and L1 batches are not reverted, just the L2 blocks, so waiting for the L1 state is not relevant.

In `rebuild_panics_if_from_block_is_already_committed`, we want the opposite: we want batches to be executed to test that the node panics.

In this test, it's a safety measure: since batch execution is disabled, it should always satisfy.

integration-tests/tests/node/rebuild.rs
vasyl-ivanchuk: this test doesn't set a fast `block_time`, so a transaction is needed to give the batcher content to commit quickly.

integration-tests/tests/node/rebuild.rs
vasyl-ivanchuk: Good point, added two additional checks:
1. that the number of committed batches is 0 right after the revert.
2. the number of committed batches in the end of the test is >= that the number of committed batches before the revert.

vasyl-ivanchuk:
Closing in favor of this PR: https://github.com/matter-labs/zksync-os-server/pull/1328

PR #1304, opened 2026-05-19
feat(tree_manager): combine tree writes with a single RocksDB flush

antonbaliasnikov:
## Summary

Instead of writing one block at a time, drain the entire inbound channel with `recv_many`, apply all blocks to a `Patched` (in-memory) tree, and flush to RocksDB in one `apply_patch` call.

Metrics and downstream sends are still emitted per block; per_block_time is split evenly from total elapsed.

Here are some statistics on how many blocks are available in the buffer when EN is synchronizing from scratch very quickly:

```
  - blocks=1: 15,772 (58.2%)
  - blocks=2: 6,925 (25.6%)
  - blocks=3: 2,518 (9.3%)
  - blocks=4: 1,121 (4.1%)
  - blocks=5+: 749 (2.8%)
  - max ever seen: 14
  - p99: 5, p95: 4, p90: 3
```

Under high load, we can combine multiple blocks and get some speed up here.

### DIFF
--- a/node/bin/src/tree_manager.rs
+++ b/node/bin/src/tree_manager.rs
@@ -17,6 +17,8 @@ use zksync_os_rocksdb::{RocksDB, RocksDBOptions, StalledWritesRetries};
 use zksync_os_sequencer::model::blocks::AppliedBlock;
 use zksync_os_storage_api::TreeBlock;
 
+const MAX_BLOCKS_PER_ITERATION: usize = 32;
+
 pub(crate) struct TreeManager {
     pub tree: MerkleTree<RocksDBWrapper>,
 }
@@ -45,7 +47,7 @@ impl PipelineComponent for TreeManager {
             state_reporter.enter_state(GenericComponentState::Idle);
 
             let mut blocks: Vec<AppliedBlock> = vec![];
-            let received = input.recv_many(&mut blocks, usize::MAX).await;
+            let received = input.recv_many(&mut blocks, MAX_BLOCKS_PER_ITERATION).await;
             if received == 0 {
                 tracing::info!("inbound channel closed");
                 return Ok(());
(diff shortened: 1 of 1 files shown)


node/bin/src/tree_manager.rs
```diff
@@ -44,85 +44,116 @@ impl PipelineComponent for TreeManager {
         loop {
             state_reporter.enter_state(GenericComponentState::Idle);
 
-            let Some(AppliedBlock {
-                output: block_output,
-                record: replay_record,
-            }) = input.recv_and_record_picked(&state_reporter).await
-            else {
+            let mut blocks: Vec<AppliedBlock> = vec![];
+            let received = input.recv_many(&mut blocks, usize::MAX).await;
```
perekopskiy: I think we should put some realistic limit here, e.g. 10, if somehow tree is far behind we can see OOM

antonbaliasnikov: Good thing to do. I put 32.

perekopskiy (approved):

kostia244:
If this improved the sync time, it's very nice!

However, I will test this on our highest load network. I suspect the computational cost of tree there might be too high compared to other components.

antonbaliasnikov:
> If this improved the sync time, it's very nice!
> 
> However, I will test this on our highest load network before merging. I suspect the computational cost of tree there might be too high compared to other components.

I've tested it during the EN sync and confirmed it fixed the issue with backpressure.

I did not measure completely computational VS disk load though, just discussed with Claude that disk operations are a bottleneck here, and computations should be relatively fast in comparison to RocksDB writes.

kostia244:
I am talking about genlayer specifically. I can confirm that it works on low tps/gas netowrks.

PR #1306, opened 2026-05-20
fix: Revert change that returned back base fee estimation

Artemka374:
## Summary

The change that was setting `max_fee_per_gas` to constant value was rollbacked by mistake. This PR returns it.

Add a test that verifies that we submit transactions at the same value as in config.


itegulov (dismissed):

LogvinovLeon:
Do we really need an integration test here? We're testing the logic and the logic should be tested with a unit-test.
Can we maybe refactor the code slightly to make it more testable so that we can test pure functions without needing to setup all this machinery to test one if?

LogvinovLeon:
I propose extracting:
`fn apply_fee_caps(configured: FeeParams, estimated: Eip1559Estimation) -> FeeParams {`

Artemka374:
tbh I feel more comfortable doing it with integration test. Functions can change, their usage as well, there's no guarantee that someone won't introduce new logic that will overwrite this one

LogvinovLeon:
Then we should find another testing point that is higher up and that we are sure about.
Maybe something like `estimate_tx_fees` (I'm not saying this point exists now, just that there should be a reasonable one). But spinning up the node to test simple logic seems very wasteful in terms of the CI time and also the complexity of the test itself. It's hard to see actual logic being tested buried within all this setup code.

If this is OK here - then @itegulov can accept.
It just seemed to me that if we spend some time separating logic from data fetching here - we can get to a unit-testable state.

Artemka374:
alright, I still think integration test might be better in that case, but I won't fight for it. Maybe we can consider adding it later on(or have the check inside of other test?)

LogvinovLeon:
I've accidentally pushed my other commit to this branch but it's force-pushed now to the original state

LogvinovLeon (approved):

PR #1308, opened 2026-05-20
feat: use simulateV1 for gas limits

Artemka374:
## Summary

  Replace the L1 sender's hardcoded `gas_limit = 15_000_000` with per-tx estimation via `eth_simulateV1`, and fix the underlying simulate impl so `gas_used` is accurate under `validation=false`.
 
  - **L1 sender**: each command is simulated in its own `SimBlock` (state from earlier blocks visible to later ones per spec). Per-tx `gas_limit = 2 * gas_used`. Falls back to `L1_GAS_LIMIT_FALLBACK = 15M` per tx on any simulate failure with a warning log.
  - **`simulate_v1_impl`**: stop zeroing basefee under `validation=false` (that drove the V31 bootloader's `native_per_gas` to 0 and under-counted pubdata). Instead pass `relax_fee_validation = !validation` to `create_tx_from_request` so under-priced requests are accepted at the request layer.
   Bootloader sees real basefee, reported `gas_used` is directly usable.


lib/rpc/src/simulate.rs
```diff
@@ -28,27 +28,21 @@ use zksync_os_storage_api::state_override_view::{
...
+    /// - `traceTransfers=true`: rejected. ZKsync OS has no transfer-tracing inspector.
+    /// - `movePrecompileToAddress`: rejected. Precompile remapping is not supported.
+    /// - `blockOverrides.difficulty`: silently ignored. Use `blockOverrides.random`
+    ///   (prevrandao) instead.
+    /// - `blockOverrides.parentBeaconBlockRoot`: silently ignored.
+    /// - `validation=false` nonce relaxation: partial. Missing nonces are auto-filled from
+    ///   state, but an explicitly supplied stale nonce is still rejected by the VM.
+    /// - `validation=false` fee relaxation: under-priced requests are accepted, but the
+    ///   real basefee is preserved during execution so the bootloader's gas accounting
+    ///   matches a future real submission.
```
itegulov, line 45: AI's favourite thing -- slightly changing perfectly fine parts of existing comments 😅

lib/rpc/src/simulate.rs
```diff
@@ -550,6 +551,22 @@ fn simulation_default_gas_limit(
...
+fn clamp_request_fees_to_basefee(call: &mut TransactionRequest, basefee: u128) {
+    if let Some(gas_price) = call.gas_price {
+        call.gas_price = Some(gas_price.max(basefee));
+    } else if call.max_fee_per_gas.is_some() || call.max_priority_fee_per_gas.is_some() {
+        call.max_fee_per_gas = Some(call.max_fee_per_gas.unwrap_or(0).max(basefee));
+        if call.max_priority_fee_per_gas.is_none() {
+            call.max_priority_fee_per_gas = Some(0);
+        }
+    } else {
+        call.gas_price = Some(basefee);
```
itegulov, line 566: Note: this fallback silently promotes a no-fee call to a legacy tx shape. Should be harmless in practice but maybe leave a comment about this.

lib/rpc/src/simulate.rs
```diff
@@ -103,39 +97,45 @@ impl<RpcStorage: ReadRpcStorage> EthCallHandler<RpcStorage> {
...
-                        .values()
-                        .any(|account| account.move_precompile_to.is_some())
-                    {
-                        return Err(EthCallError::SimulateMovePrecompileNotSupported);
+            let mut user_state_overrides = sim_block.state_overrides.unwrap_or_default();
+            if !validation {
+                // RPC-layer `ensure_fees` only suppresses the early `FeeCapTooLow`; the
+                // bootloader still enforces `gas_price >= basefee` and
+                // `balance >= gas_limit * gas_price + value`. Clamp fees up to basefee and
+                // bump unfunded senders to `U256::MAX`. User-supplied overrides win.
```
itegulov, line 108: I believe this is because we are using `run_block` interface where fee validation is enforced. Ideally we need to rework multivm interface to support `simulate_block` with similar semantics instead of only `simulate_tx`

Artemka374: Agree, although should be good enough for now

itegulov (approved):
LGTM

PR #1311, opened 2026-05-20
feat(multivm): use v0.1.3 interface

itegulov:
## Summary

Migrates us to v0.1.3 interface, see https://github.com/matter-labs/zksync-os-interface/pull/71 for list of changes. Overall this should improve maintenance of old versions going forward.


lib/storage_api/src/model.rs
```diff
@@ -165,24 +166,92 @@ pub struct BlockContext {
...
+        self.mix_hash
+    }
+
+    fn blob_fee(&self) -> U256 {
+        self.blob_fee
+    }
+
+    fn is_gateway(&self) -> bool {
+        // todo: source from a new optional field?
+        false
```
itegulov, line 255: This is the easy option, more involved one is to get is from chain-specific sidecar and not store it unnecessarily on disk (like we do with `chain_id` which is the same for all blocks and can never change).

perekopskiy (approved):

PR #1315, opened 2026-05-21
fix(gateway-launch): Harden wallet and prover API artifacts

sidhujag:
## Summary
- Restore fail-closed validation for wallet files used by funding, while still hardening zkstack-generated wallet artifacts immediately after creation.
- Avoid generating or leaving public prover API nginx artifacts when no-proofs configs intentionally omit Basic Auth for loopback-only use.

## Test plan
- `bash -n scripts/gateway-launch/fund-wallets.sh scripts/gateway-launch/gateway-ecosystem-create.sh scripts/gateway-launch/generate-os-server-configs.sh`
- Compiled the embedded Python block in `scripts/gateway-launch/generate-os-server-configs.sh` with `python3`.


sidhujag:
Opened against upstream by mistake; superseded by the Syscoin fork PR.

PR #1317, opened 2026-05-22
feat(l1_watcher,node,sequencer): EN L1 consistency check + drop EN cross-check watchers

Artemka374-claude:
## Summary

- `L1PersistBatchWatcher` now verifies both `batchHash` (state_commitment) and `commitment` (public-input hash) from each `BlockExecution` event against locally-computed values. Two distinct panics on divergence.
- The locally-computed values come from `ExtendedCommitBatchInfo::build` (the same function MN's batcher uses) run over cached `TreeBlock`s. A new lightweight `TreeBlockCacher` pipeline component sits between `TreeManager` and the downstream sink, mirroring `BatchVerificationResponder`'s caching pattern. Cache is evicted per batch after persisting.
- `L1TxWatcher`, `L1UpgradeTxWatcher`, `GatewayMigrationWatcher`, and `InteropWatcher` no longer spawn on External Nodes.
- `BlockContextProvider` skips the `Pool` drain on EN (the four L1-fed subpools are empty without their watchers; the next `Replay` command supplies fresh `starting_cursors`).

`pubdata_mode` is derived from the L1 commit calldata's `DACommitmentScheme` so EN doesn't need to know its chain's mode from config. Safe because the contract enforces `daCommitmentScheme == s.l2DACommitmentScheme` at commit time.

## Note on attestation semantics

If batch verification is enabled, EN signatures previously attested to "EN's replay matches MN's claim AND L1 watcher data matches MN's claim." With the watcher removal they attest only to the first half. L1 contract checks (priorityOperationsHash via priorityTree merkle inclusion, dependencyRootsRollingHash via messageRoot lookup, slChainId, systemContractsUpgradeTxHash via batch_output_hash) prevent a malicious MN from ever landing a bad batch on L1, so chain safety is unchanged — only the per-EN attestation is weaker.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features --workspace --exclude zksync_os_integration_tests -- -D warnings`
- [x] `cargo nextest run --workspace --exclude zksync_os_integration_tests` — 295 passed
- [x] `cargo nextest run -p zksync_os_integration_tests --profile no-pig` — 110 passed


lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -256,7 +256,26 @@ impl<BatchStorage: WriteBatch> ProcessRawEvents for L1PersistBatchWatcher<BatchS
                 let batch_number = execute.batchNumber.to::<u64>();
                 if batch_number > self.last_persisted_batch_on_start {
                     let batch_hash = execute.batchHash;
-                    if let Some(committed_batch) = self.committed_batches.remove(&batch_number) {
+                    let commitment = execute.commitment;
+                    if let Some(committed_batch) = self.committed_batches.get(&batch_number) {
+                        // Panic on divergence rather than persist mismatched data.
+                        let local = &committed_batch.batch_info;
```
kostia244: `batch_hash = execute.bashHash`. `execute` comes from L1
```
let execute = BlockExecution::decode_log(&log.inner)?.data;
```
-----------------
`local = committed_batch.batch_info` comes from `self.committed_batches`
```
let committed_batch = self.parse_committed_batch(provider, report, log).await?;

            self.committed_batches.insert(batch_number, committed_batch);
```

Both come from L1 no?

node/bin/src/lib.rs
```diff
@@ -879,25 +904,7 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
-            node_startup_state.l1_state.diamond_proxy_l1.clone(),
-            node_startup_state.l1_state.diamond_proxy_sl.clone(),
-            bytecode_supplier_address,
-            current_protocol_version.clone(),
-            upgrade_subpool,
-        )
-        .await
-        .expect("failed to start L1 upgrade transaction watcher")
-        .run(),
+        node_role.is_external(),
```
kostia244, line 909: nit: Why not pass node_role directly?

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -256,7 +256,26 @@ impl<BatchStorage: WriteBatch> ProcessRawEvents for L1PersistBatchWatcher<BatchS
                 let batch_number = execute.batchNumber.to::<u64>();
                 if batch_number > self.last_persisted_batch_on_start {
                     let batch_hash = execute.batchHash;
-                    if let Some(committed_batch) = self.committed_batches.remove(&batch_number) {
+                    let commitment = execute.commitment;
+                    if let Some(committed_batch) = self.committed_batches.get(&batch_number) {
+                        // Panic on divergence rather than persist mismatched data.
```
kostia244: If I remember correctly, BlockApplier/BlockExecutor don't wait for this watcher. So it's possible that we persist the block before we get here?

kostia244:
Old behavior: first check then apply
New behavior: async verify, after execution. (correct me if I'm wrong here)

Right now `BatchVerificationResponder` is at the end of the pipeline. Which means we  have to do async/delayed verify(after we already went through BlockApplier etc.)
Otherwise it would be a deadlock: we cannot commit without signatures, cannot get to verification without L1 commit.

`BatchVerificationResponder` only verifies that the execution is correct given the transactions, not L1 consistency.

So we will be signing batches without verifying consistency with L1.

We either accept this, or find another way to verify L1 transactions(maybe keep these watches for 2 FA nodes?)

PR #1318, opened 2026-05-22
fix: tracing multi-root calls

Raid5594:
## Summary

Fixes RPC tracing for L1->L2 transactions that produce multiple top-level execution frames.

Some L1->L2 transactions can now be traced as several root calls, for example asset-tracker hooks around the actual transaction root. The call tracer previously kept only one completed root frame at transaction finalization, which could hide the actual transaction subtree and omit frames such as the `CREATE2` used to deploy bridged token contracts.

This PR preserves all completed root frames by wrapping multi-root traces in a synthetic top-level frame, while keeping single-root traces unchanged. It also updates the JS tracer transaction result context to account for the same multi-root shape.


lib/rpc/src/js_tracer/tracer.rs
```diff
@@ -24,6 +26,61 @@ use zksync_os_types::{ZkTransaction, ZksyncOsEncode};
...
 const MAX_JS_TRACER_PAYLOAD_BYTES: usize = 512 * 1024;
 
+fn multi_root_context(
+    input_transactions: &[ZkTransaction],
+    result_index: usize,
+    roots: Vec<TxContext>,
+) -> TxContext {
+    debug_assert!(roots.len() > 1);
+
+    let tx = input_transactions.get(result_index);
```
0xVolosnikov: Can't we just pass a reference to the tx as a param of this function?

lib/rpc/src/sandbox.rs
```diff
@@ -224,6 +232,83 @@ impl CallTracer {
...
+            },
+        }
+    }
+
+    fn without_asset_tracker_root_frames(roots: Vec<CallFrame>) -> Vec<CallFrame> {
+        if roots.len() <= 1 {
+            return roots;
+        }
+
+        let mut kept = Vec::with_capacity(roots.len());
```
0xVolosnikov: nit: it will always have 1 frame now, not `roots.len`

lib/rpc/src/sandbox.rs
```diff
@@ -224,6 +232,83 @@ impl CallTracer {
...
+        let mut ignored = Vec::new();
+        for root in roots {
```
0xVolosnikov: you can use `roots.iter().filter(...)...`

lib/rpc/src/sandbox.rs
```diff
@@ -224,6 +232,83 @@ impl CallTracer {
...
+            typ: if tx.is_some_and(|tx| tx.to().is_none()) {
+                "CREATE".to_string()
+            } else {
+                "CALL".to_string()
+            },
+        }
+    }
+
+    fn without_asset_tracker_root_frames(roots: Vec<CallFrame>) -> Vec<CallFrame> {
+        if roots.len() <= 1 {
```
0xVolosnikov: I don't think that we should do this check - it's not going to increase performance, but adds some implied assumptions

lib/rpc/src/sandbox.rs
```diff
@@ -18,6 +18,14 @@ pub const STACK_SIZE: usize = 1024;
 /// zksync-os ergs per gas.
 pub const ERGS_PER_GAS: u64 = 256;
 
+pub(crate) const ASSET_TRACKER_ADDRESS: Address =
+    address!("0x000000000000000000000000000000000001000f");
+pub(crate) const ASSET_TRACKER_ROOT_SELECTOR: [u8; 4] = [0x03, 0x11, 0x7c, 0x8c];
+
+pub(crate) fn is_asset_tracker_root_call(to: Option<Address>, input: &[u8]) -> bool {
+    to == Some(ASSET_TRACKER_ADDRESS) && input.starts_with(&ASSET_TRACKER_ROOT_SELECTOR)
```
0xVolosnikov: I think it is better to check "from" - it will always be the bootloader address. With the current version we can have confusing situations if the real call is to asset tracker with that selector

lib/rpc/src/sandbox.rs
```diff
@@ -224,6 +232,83 @@ impl CallTracer {
...
+        let last_root = roots.last().expect("multi-root trace is non-empty");
+        let first_error = roots
+            .iter()
+            .find_map(|call| call.error.as_ref().map(ToOwned::to_owned));
+        let first_revert_reason = roots
+            .iter()
+            .find_map(|call| call.revert_reason.as_ref().map(ToOwned::to_owned));
+
+        CallFrame {
+            from: tx.map_or(first_root.from, ZkTransaction::signer),
```
0xVolosnikov: Is "first_root" going to be a call to AssetTracker? Won't it be from the bootloader then?

0xVolosnikov: Ah, it is expected that "without_asset_tracker_root_frames" filters it out. Seems like this multi_root is going to be applied to 1 root, so it is a bit redundant here

0xVolosnikov:
Actually seems like we use L2_BASE_TOKEN_ADDRESS as a caller for asset tracker calls

0xVolosnikov (approved):

PR #1319, opened 2026-05-22
feat: S3 replay archive implementation

perekopskiy:
## Summary

- adds support for s3 storage for archiving
- parallel object processing in `ReplayArchiveComponent`

### DIFF
--- a/integration-tests/tests/node/replay_archive.rs
+++ b/integration-tests/tests/node/replay_archive.rs
@@ -63,10 +63,7 @@ async fn encrypted_replay_archive_recovers_node_storage_end_to_end(
 
     let archive_root = match &tester.config().replay_archive_config {
         ReplayArchiveConfig::FileSystem { root_path, .. } => root_path.clone(),
-        ReplayArchiveConfig::Noop => unreachable!("test enables replay archive"),
-        ReplayArchiveConfig::S3WithCredentialFile { .. } => {
-            unreachable!("test enables filesystem replay archive")
-        }
+        _ => unreachable!("test enables replay archive"),
     };
     let rocks_db_path = tester.config().general_config.rocks_db_path.clone();
     let stopped = tester.stop().await?;
--- a/integration-tests/tests/node/replay_archive.rs
+++ b/integration-tests/tests/node/replay_archive.rs
@@ -64,6 +64,9 @@ async fn encrypted_replay_archive_recovers_node_storage_end_to_end(
     let archive_root = match &tester.config().replay_archive_config {
         ReplayArchiveConfig::FileSystem { root_path, .. } => root_path.clone(),
         ReplayArchiveConfig::Noop => unreachable!("test enables replay archive"),
+        ReplayArchiveConfig::S3WithCredentialFile { .. } => {
+            unreachable!("test enables filesystem replay archive")
+        }
     };
     let rocks_db_path = tester.config().general_config.rocks_db_path.clone();
     let stopped = tester.stop().await?;
--- a/lib/replay_archive/src/bin/replay_archive_recovery.rs
+++ b/lib/replay_archive/src/bin/replay_archive_recovery.rs
@@ -96,11 +96,10 @@ async fn main() -> anyhow::Result<()> {
                 let auth_mode = if let Some(path) = s3_credential_file_path {
                     S3ReplayArchiveAuthMode::AuthenticatedWithCredentialFile(path)
                 } else {
-                    if !s3_anonymous {
-                        anyhow::bail!(
-                            "--s3-credential-file-path is required unless --s3-anonymous is set"
-                        );
-                    }
+                    anyhow::ensure!(
+                        s3_anonymous,
+                        "--s3-credential-file-path is required unless --s3-anonymous is set"
+                    );
                     S3ReplayArchiveAuthMode::Anonymous
                 };
                 let reader = S3ReplayArchiveReader::new(S3ReplayArchiveConfig {
--- a/lib/replay_archive/src/bin/replay_archive_recovery.rs
+++ b/lib/replay_archive/src/bin/replay_archive_recovery.rs
@@ -1,7 +1,8 @@
 use clap::{Parser, Subcommand};
 use std::path::PathBuf;
 use zksync_os_replay_archive::{
-    FileSystemReplayArchiveReader, download_all_replay_archive_objects, parse_age_x25519_identity,
+    FileSystemReplayArchiveReader, S3ReplayArchiveAuthMode, S3ReplayArchiveConfig,
+    S3ReplayArchiveReader, download_all_replay_archive_objects, parse_age_x25519_identity,
     read_age_x25519_identity, recover_replay_records_to_rocksdb_with_optional_decryption,
 };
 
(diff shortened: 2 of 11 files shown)


integration-tests/tests/node/replay_archive.rs
```diff
@@ -64,6 +64,9 @@ async fn encrypted_replay_archive_recovers_node_storage_end_to_end(
     let archive_root = match &tester.config().replay_archive_config {
         ReplayArchiveConfig::FileSystem { root_path, .. } => root_path.clone(),
         ReplayArchiveConfig::Noop => unreachable!("test enables replay archive"),
+        ReplayArchiveConfig::S3WithCredentialFile { .. } => {
+            unreachable!("test enables filesystem replay archive")
+        }
```
slowli: Nit: The match arms could be united via `|`.

lib/replay_archive/src/bin/replay_archive_recovery.rs
```diff
@@ -58,10 +82,37 @@ async fn main() -> anyhow::Result<()> {
...
+                download_all_replay_archive_objects(&reader, &output_root).await?
+            } else {
+                let auth_mode = if let Some(path) = s3_credential_file_path {
+                    S3ReplayArchiveAuthMode::AuthenticatedWithCredentialFile(path)
+                } else {
+                    if !s3_anonymous {
+                        anyhow::bail!(
+                            "--s3-credential-file-path is required unless --s3-anonymous is set"
+                        );
+                    }
```
slowli: Nit: It would be more idiomatic to use `anyhow::ensure!(s3_anonymous, ..)` here.

lib/replay_archive/src/component.rs
```diff
@@ -27,20 +30,81 @@ where
...
+                match records.try_recv() {
+                    Ok(record) => {
+                        in_flight.push(archive_replay_record(&archive, record));
+                    }
+                    Err(mpsc::error::TryRecvError::Empty) => break,
+                    Err(mpsc::error::TryRecvError::Disconnected) => records_closed = true,
+                }
+            }
+
+            if in_flight.is_empty() {
```
slowli: Dumb question: Is this special handling of `in_flight.is_empty()` case really required? I'd expect the following `tokio::select!` to cover it; if the are no in-flight futures, the corresponding arm would never be triggered. The only issue is that it would be necessary move the `if records_closed` check (presumably, to handling `in_flight.next() == None`.

Ditto for the special sync receive loop above; intuitively, it can be made a part of `tokio::select!` as well (maybe, by marking `select!` [as biased](https://docs.rs/tokio/latest/tokio/macro.select.html#fairness)).

lib/replay_archive/src/component.rs
```diff
@@ -27,20 +30,81 @@ where
...
-                })?;
-            REPLAY_ARCHIVE_METRICS
-                .last_archived_block_number
-                .set(block_number);
+    pub async fn run(self) -> anyhow::Result<()> {
+        let Self {
+            archive,
+            mut records,
+        } = self;
+        let mut in_flight = FuturesUnordered::new();
```
slowli: Since this already uses `futures` primitives, maybe try to use [`try_for_each_concurrent()`](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_for_each_concurrent) on the stream of `records`? (Or [`try_buffer_unordered()`](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_buffer_unordered) if it's necessary to preserve `highest_archived_block_number` handling.) These methods seem to provide necessary functionality (short-circuiting on error, limiting the number of concurrent futures). Or do these methods lack some necessary functionality?

lib/replay_archive/src/init.rs
```diff
@@ -70,6 +66,49 @@ pub async fn init_replay_archive(
...
+                .with_context(|| format!("failed to create replay archive S3 session {session}"))
+                .expect("failed to initialize S3 replay archive");
+            let archive = archive_for_storage(storage, &encryption);
+            let (sender, component) = ReplayArchiveComponent::new(archive.clone());
+            runtime.spawn_critical_task("replay archive", async move {
+                component
+                    .run()
+                    .await
+                    .expect("replay archive component failed");
+            });
```
slowli: Nit: It could make sense to return `archive` from the `match` expression and deduplicate this code. (`session` creation also can be deduplicated, although it would require short-circuiting on `ReplayArchiveConfig::Noop` early to avoid creating it uselessly.)

lib/replay_archive/src/s3.rs
```diff
@@ -0,0 +1,386 @@
...
+        };
+        continuation_token = Some(next_token.to_owned());
+    }
+
+    Ok(())
+}
+
+fn parse_s3_archive_key(object_key: &str) -> anyhow::Result<Option<ReplayArchiveKey>> {
+    let parts = object_key.split('/').collect::<Vec<_>>();
+    if parts.len() != 3 || parts[2] == SESSION_MARKER_FILE_NAME {
```
slowli, line 306: Dumb question: Doesn't `session_marker_key()` consist of 2 parts, rather than 3? If so, how can the second condition be triggered?

lib/replay_archive/src/s3.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    async fn contains_object(
+        &self,
+        block_number: BlockNumber,
+        block_hash: BlockHash,
+    ) -> anyhow::Result<bool> {
+        let key = self.object_key(block_number, block_hash);
+        let response = self
+            .client
+            .list_objects_v2()
```
slowli: Could `head_object()` be used here instead? I feel it would be more straightforward, but maybe it doesn't work for this use case?

lib/replay_archive/src/s3.rs
```diff
@@ -0,0 +1,386 @@
...
+
+    let session = parts[0]
+        .parse::<ReplayArchiveSession>()
+        .with_context(|| format!("failed to parse replay archive S3 session in {object_key}"))?;
+    let block_number = parts[1].parse::<BlockNumber>().with_context(|| {
+        format!("failed to parse replay archive S3 block number in {object_key}")
+    })?;
+    let block_hash = {
+        let block_hash = parts[2].strip_prefix("0x").unwrap_or(parts[2]);
+        BlockHash::from_str(block_hash).with_context(|| {
```
slowli: Nit: It's possible to use `block_hash.parse::<BlockHash>()`, similar to parsing `BlockNumber` above. This would be more consistent and allow to not import `FromStr` into the scope.

lib/replay_archive/src/s3.rs
perekopskiy: It does, it's rather a sanity check, in case at some point it will consist of 3 parts

lib/replay_archive/src/s3.rs
perekopskiy: agree, change this part

lib/replay_archive/src/component.rs
perekopskiy: try_for_each_concurrent serves the goal, applied the suggestion

lib/replay_archive/src/component.rs
perekopskiy: makes sense, but after try_for_each_concurrent is used it's not relevant anymore

lib/replay_archive/src/component.rs
```diff
@@ -31,65 +33,41 @@ where
...
-                            continue;
-                        }
-                    }
-                }
-                result = in_flight.next() => {
-                    result.expect("in-flight archive writes are not empty")?
-                }
-            };
+fn update_highest_archived_block_number(
+    highest_archived_block_number: &Mutex<Option<BlockNumber>>,
```
slowli, line 59: Bikeshedding: I'd use an atomic here (initially set to 0) to make things easier. Since `vise` gauges are already backed up by atomics, it would be even better to add a [`fetch_max()`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html#method.fetch_max) equivalent (requires cross-repository work, so this wouldn't be my first option). Alternatively, it could be possible to use an empty mutex to linearize accesses to `last_archived_block_number` (it already [has a getter](https://matter-labs.github.io/vise/vise/struct.Gauge.html#method.get)).

lib/replay_archive/src/component.rs
```diff
@@ -27,20 +32,57 @@ where
...
-                .last_archived_block_number
-                .set(block_number);
-        }
-        Ok(())
+    pub async fn run(self) -> anyhow::Result<()> {
+        let Self { archive, records } = self;
+        let highest_archived_block_number = Mutex::new(None);
+
+        ReceiverStream::new(records)
+            .map(Ok::<_, anyhow::Error>)
```
slowli, line 40: Micro-nit: `anyhow` has a special function `anyhow::Ok` to avoid using a turbofish in cases like this.

slowli (approved):

perekopskiy:
thanks for the review, I will fix latest nits in the follow-up PR

PR #1321, opened 2026-05-23
fix(consensus): forward follower transactions over RPC

romanbrodetski-ai:
## Summary
- use one transaction forwarder for both `main_node_rpc_url` and consensus leader RPC forwarding
- add consensus tx forwarding config as `<peer_id>@<host>:<rpc_port>` entries and check peer coverage at startup
- keep consensus forwarding single-shot: stale or missing leader info returns an error, no retry
- cover arbitrary-node tx submission and round-robin storm submission

## Testing
- `cargo check -p zksync_os_rpc -p zksync_os_server -p zksync_os_integration_tests`
- `cargo clippy -p zksync_os_rpc -p zksync_os_server -p zksync_os_integration_tests --all-targets -- -D warnings`
- `cargo nextest run -p zksync_os_integration_tests --profile no-pig consensus_cluster_accepts_transactions_from_any_node`
- `cargo nextest run -p zksync_os_integration_tests --profile no-pig consensus_restarted_node_catches_up_while_transaction_storm_continues`
- `cargo nextest run -p zksync_os_integration_tests --profile no-pig`
- `cargo nextest run -p zksync_os_integration_tests --profile no-pig --run-ignored only`

### DIFF
--- a/integration-tests/tests/consensus_node/mod.rs
+++ b/integration-tests/tests/consensus_node/mod.rs
@@ -1,6 +1,6 @@
-use alloy::eips::BlockId;
-use alloy::network::TransactionBuilder;
-use alloy::primitives::{Address, U256};
+use alloy::eips::{BlockId, Encodable2718};
+use alloy::network::{NetworkTransactionBuilder, ReceiptResponse, TransactionBuilder};
+use alloy::primitives::{Address, TxHash, U256};
 use alloy::providers::Provider;
 use alloy::rpc::types::TransactionRequest;
 use anyhow::Context as _;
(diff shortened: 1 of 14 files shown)


node/bin/src/init_tx_forwarder.rs
```diff
@@ -0,0 +1,64 @@
...
+            .await
+            .unwrap_or_else(|err| {
+                panic!("could not connect to consensus RPC {rpc_url} for peer {peer_id}: {err}")
+            })
+            .erased();
+        providers.insert(peer_id, TxForwardEndpoint::new(rpc_url, provider));
+    }
+    for peer_id in &config.consensus_config.peer_ids {
+        if !providers.contains_key(&peer_id.to_string()) {
+            panic!("missing consensus tx RPC forwarder for peer {peer_id}");
```
itegulov, line 42: Note: this means every node needs to specify its own tx forwarding RPC url, right? On one hand this is consistent with `peer_ids`, on the other the value is superficial because it will never be used in runtime. Consider skipping self-node peer id in this check but no strong opinion in any case

itegulov, line 42: I'd accumulate all missing peer ids and panic after the for loop with the full diff so operator can fix all missing entries at once

lib/rpc/src/tx_forwarder.rs
```diff
@@ -0,0 +1,205 @@
...
+    #[error("consensus leader is unknown")]
+    NoKnownLeader,
+    #[error("no RPC forwarder is configured for consensus leader {0}")]
+    NoProvider(String),
+    #[error(transparent)]
+    Rpc(#[from] RpcError<TransportErrorKind>),
+}
+
+impl TxForwardError {
+    pub(crate) fn as_rpc_error(&self) -> Option<&RpcError<TransportErrorKind>> {
```
itegulov, line 199: Nit: looks like the only usage is in `if let` matcher anyway so this feels redundant. Feel free to keep as is anyway

lib/rpc/src/tx_forwarder.rs
```diff
@@ -0,0 +1,205 @@
...
+                if status.as_ref().is_some_and(|status| status.is_leader) {
+                    Self::log_not_forwarding(call, tx_hash);
+                    return Ok(None);
+                }
+
+                let leader = status
+                    .and_then(|status| status.current_leader)
+                    .ok_or(TxForwardError::NoKnownLeader)?;
+                if leader == *node_id {
+                    return Err(TxForwardError::NoKnownLeader);
```
itegulov, line 104: This edge case looks weird to me. AFAIU this can happen while in transitional state during election (claiming leader but not confirmed yet). The transaction can be processed locally in such case no? Or is this defensive against cases where the node in question will not in fact win the election?

integration-tests/tests/consensus_node/mod.rs
```diff
@@ -218,6 +277,71 @@ async fn consensus_cluster_forms_with_three_nodes_and_replicates_blocks() -> any
...
+
+        Ok(())
+    }
+    .await;
+    let shutdown_result = cluster.shutdown_all().await;
+    result.and(shutdown_result)
+}
+
+#[test_log::test(tokio::test)]
+async fn consensus_cluster_send_raw_transaction_sync_accepts_leader_and_replica()
```
itegulov, line 305: Is "replica" a term we want to stick with? Might be mistaken but I don't think we've used it before this PR

itegulov (approved):
Is your intention to have this as a temporary workaround? If so feel free to ignore comments, they are mostly nits for the current implementation

PR #1325, opened 2026-05-25
fix(watcher): skip old transactions in upgrade watcher

kostia244:
## Summary

We changed what logs we use to track L1 Upgrade transactions, which means we are missing some old ones. In particular `30.0 -> 30.1` upgrade transactoin for stage-rollup. This breaks EN sync in two ways:

1. ENs get stuck when replaying, because they are waiting for watcher to find the transaction.
2. The watcher itself has some sanity checks 

Note that if we stop relying on upgrade watcher for replay validation this would only fix point 1.

Fix:
Introduce `MIN_VERSION_WITH_RELIABLE_UPGRADE_LOGS = v30.2`
1. Skip the legacy(`version <= 30.2`) upgrade transaction validation for replays.
2. Fix the watcher not to panic due to missed legacy upgrade transactions.

### DIFF
--- a/lib/types/src/protocol/mod.rs
+++ b/lib/types/src/protocol/mod.rs
@@ -38,10 +38,12 @@ impl Deref for ProtocolSemanticVersion {
 
 impl ProtocolSemanticVersion {
     /// Smallest version such that upgrading to that version uses the current log format
+    /// In other words: if replay record has protocol version this or greater,
+    /// we can expect the watcher to pick up the logs.
     ///
     /// Example:
     /// For 30.1 -> 30.2, 30.1 -> 31.0 we expect to find a log
-    /// Fo r 30.0 -> 30.1 or 30.1 -> 30.1 we don't
+    /// For 30.0 -> 30.1 or 30.1 -> 30.1 we don't
     pub const MIN_VERSION_WITH_RELIABLE_UPGRADE_LOGS: Self = Self::new(0, 30, 2);
 
     pub const fn new(major: u64, minor: u64, patch: u64) -> Self {
--- a/lib/types/src/protocol/mod.rs
+++ b/lib/types/src/protocol/mod.rs
@@ -37,6 +37,10 @@ impl Deref for ProtocolSemanticVersion {
 }
 
 impl ProtocolSemanticVersion {
+    /// Upgrade logs before this protocol version used a different format that
+    /// the current L1 upgrade watcher cannot find or process reliably.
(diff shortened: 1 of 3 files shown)


lib/types/src/protocol/mod.rs
```diff
@@ -37,6 +37,10 @@ impl Deref for ProtocolSemanticVersion {
 }
 
 impl ProtocolSemanticVersion {
+    /// Upgrade logs before this protocol version used a different format that
+    /// the current L1 upgrade watcher cannot find or process reliably.
+    pub const MIN_VERSION_WITH_RELIABLE_UPGRADE_LOGS: Self = Self::new(0, 30, 2);
```
perekopskiy, line 47: Shouldn't it be v30.1? I thought that we would upgrade 30.1->31.0 on L1

kostia244: (I changed the code a bit).

So this is meant to be the lowest version such that upgrading to that version leads to a reliable log. I think this is better because it reflects the reality.
If we're looking at 30.1 -> 30.1 "upgrade", we will not catch UpgradeInfo. But if we do `30.1 -> 31.0` we will.

Even though that UpgradeInfo is not utilized, I thought it makes sense to do the cutoff this way, as it reflects the reality.

Another argument is that when we do upgrade, the replay record of block with the upgrade transaction has the new version, so in a way the upgrade log is associated with the new protocol version.

LMK if this makes sense please.

perekopskiy: definition makes sense, but I see this code
```
// Since we don't have the old events, current_version might be wrong
        // Update it here to pass related sanity checks
        if self.current_protocol_version
            < ProtocolSemanticVersion::MIN_VERSION_WITH_RELIABLE_UPGRADE_LOGS
        {
            self.current_protocol_version = request.old_protocol_version.clone();
        }
        
        
        if request.old_protocol_version < self.current_protocol_version {
            tracing::info!(
                ?request.old_protocol_version,
                ?self.current_protocol_version,
                "ignoring upgrade timestamp for older protocol version"
            );
            return Ok(());
        }
```

IIUC self.current_protocol_version and request.old_protocol_version will be 0.30.1 so it will enter `if` that returns Ok

perekopskiy (approved):

lib/types/src/protocol/mod.rs
perekopskiy: nvm, makes sense

itegulov (approved):

PR #1326, opened 2026-05-25
feat(l1-sender):  tx resubmission after timeout

Artemka374:
## Summary

This PR adds resubmission of transaction after the configured timeout. Transactions is resubmitted with replacement fees configured once.

### DIFF
--- a/lib/l1_sender/src/lib.rs
+++ b/lib/l1_sender/src/lib.rs
@@ -64,7 +64,16 @@ impl StateLabel for L1SenderState {
 const METHOD_NOT_FOUND_CODE: i64 = -32601;
 /// Future that resolves into a (fallible) transaction receipt.
 type TransactionReceiptFuture = BoxFuture<'static, anyhow::Result<TransactionReceipt>>;
-type PendingTx<Input> = (TransactionReceiptFuture, Input, Instant);
+
+/// An L1 transaction that has been submitted and is awaiting inclusion. We track the original
+/// nonce so that, if the tx is not confirmed within `transaction_timeout`, it can be replaced
+/// in-place (RBF) with a higher-fee transaction at the same nonce.
+struct PendingTx<Input> {
+    receipt_fut: TransactionReceiptFuture,
+    command: Input,
+    submitted_at: Instant,
+    nonce: u64,
+}
 
 const REQUIRED_CONFIRMATIONS_L1: u64 = 3;
 /// In case there's only one chain connected to gateway, it is very likely that there will be not enough block production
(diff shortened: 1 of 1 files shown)


lib/l1_sender/src/lib.rs
```diff
@@ -202,110 +213,204 @@ where
...
+        } = pending_tx;
+        let timeout = self.config.transaction_timeout;
+        // A zero `transaction_timeout` disables both the inclusion-warning cadence and resubmission
+        // — fall through to plain `.await` so behavior matches the pre-resubmission code path.
+        if timeout.is_zero() {
+            let receipt = receipt_fut.await?;
+            return Ok((command, receipt, submitted_at));
+        }
+
+        match tokio::time::timeout(timeout, receipt_fut).await {
```
itegulov, line 367: There is a slight race condition here. The receipt can confirm while we are preparing a resubmission. There is no way to completely get rid of it and I think the best we could do is make the `Err(_) =>` logic tolerant to resubmission errors (e.g. if the error was "nonce too low" it means we should wait for the future some more).

lib/l1_sender/src/lib.rs
```diff
@@ -202,110 +213,204 @@ where
...
+                    timeout_secs = timeout.as_secs(),
+                    "L1 transaction not confirmed within timeout; resubmitting at replacement fees \
+                     using the original nonce",
+                );
+                let resubmitted = self
+                    .submit_transaction(command, fee_config, true, Some(nonce), &range)
+                    .await
+                    .context("resubmit L1 transaction at replacement fees")?;
+                let receipt = resubmitted.receipt_fut.await?;
+                Ok((resubmitted.command, receipt, resubmitted.submitted_at))
```
itegulov: This makes us track resubmission latency and ignore original tx's latency, is that what we want?

lib/l1_sender/src/lib.rs
```diff
@@ -202,110 +213,204 @@ where
...
+        pending_tx: PendingTx<Input>,
+        fee_config: L1SenderFeeConfig,
+    ) -> anyhow::Result<(Input, TransactionReceipt, Instant)> {
+        let PendingTx {
+            receipt_fut,
+            command,
+            submitted_at,
+            nonce,
+        } = pending_tx;
+        let timeout = self.config.transaction_timeout;
```
itegulov, line 359: You use the same timeout here as in `wait_for_confirmed_receipt`. So the warning log there might not get printed. I'd probably get rid of the timeout logic in `wait_for_confirmed_receipt` altogether and move the warning here as it doesn't bring a lot of value anymore

lib/l1_sender/src/lib.rs
Artemka374: I don't think so, will change

PR #1328, opened 2026-05-25
feat(sequencer): L1 batch revert

vasyl-ivanchuk:
## Summary

Adds three startup revert / rebuild modes and an L1 revert watcher for External Nodes.

#### Revert/rebuild modes:

**`block_rebuild`** — replay local blocks from `from_block_number`, no L1 interaction:
```yaml
sequencer:
  rebuild:
    mode: block_rebuild
    from_block_number: 5
    from_block_hash: '0x20536f06...'
    blocks_to_empty:
      - 5
    reset_timestamps: false
```

**`danger_block_rebuild_with_l1_revert`** — revert committed L1 batches (auto-derived from `from_block_number`), then rebuild local blocks from the same boundary:
```yaml
sequencer:
  rebuild:
    mode: danger_block_rebuild_with_l1_revert
    from_block_number: 7
    from_block_hash: '0x21f2ed4d...'
    blocks_to_empty:
      - 7
    reset_timestamps: false
    l1_reverter_sk: '0x...'
```

**`l1_revert`** — revert committed L1 batches only; local blocks are untouched:
```yaml
sequencer:
  rebuild:
    mode: l1_revert
    from_batch_number: 10
    from_batch_commit_tx_hash: '0x2a32849e...'
    l1_reverter_sk: '0x...'
```

#### Field reference:

| Field | Modes | Required | Description |
|---|---|---|---|
| `from_block_number` | `block_rebuild`, `danger_*` | yes | First L2 block to rebuild from. For `danger_*` must be inside a committed-only (not executed) batch. |
| `from_block_hash` | `block_rebuild`, `danger_*` | yes | Expected hash of `from_block_number` before the operation. Skipped on restart when hash no longer matches (idempotency guard). |
| `blocks_to_empty` | `block_rebuild`, `danger_*` | no | List of L2 block numbers to strip all transactions from during rebuild. |
| `reset_timestamps` | `block_rebuild`, `danger_*` | no | If `true`, rebuilt blocks receive fresh timestamps instead of replaying originals. Default: `false`. |
| `from_batch_number` | `l1_revert` | yes | First batch to revert (≥ 1). All committed batches from this number upward are reverted on L1. |
| `from_batch_commit_tx_hash` | `l1_revert` | yes | Hash of the L1 transaction that committed `from_batch_number`. The simplest way to get the hash is the sequencer's own logs (search for the next line): `commit batch <from_batch_number> ... succeeded on L1   (tx_hash=0x…)`. |
| `l1_reverter_sk` | `danger_*`, `l1_revert` | yes | Private key (or GCP KMS reference) holding the reverter role on ValidatorTimelock.

#### EN changes:
- **`L1RevertWatcher`**: External Nodes watch for `BlocksRevert` events on the settlement layer and restart to resync when the Main Node reverts batches.


node/bin/src/config/mod.rs
```diff
@@ -844,6 +851,17 @@ pub struct SequencerConfig {
...
     pub block_rebuild: Option<RebuildBlocksConfig>,
 
+    /// Optional startup L1 revert configuration.
+    ///
+    /// If set on the Main Node, the node will send `revertBatchesSharedBridge` before startup
+    /// replay / rebuild starts.
+    #[config(nest)]
+    #[config_validate(custom(
+        |root: &Config, value: &Option<L1RevertConfig>| !root.consensus_config.enabled || value.is_none(),
+        "requires `consensus.enabled=false`"
```
RomanBrodetski, line 985: shall we make it also `requires` `reverter_sk`? This way we'd catch an inconsistent configuration with config check in gitops (please ask Misha for more details if interested).

In fact, an alternative solution would be to put `reverter_sk` inside `L1RevetConfig` and make non-optional - this way we get this validation "for free" (because if `L1RevertConfig` present, the key also has to be set). This has a con that then reverter_sk is far from other secret keys that server wants. However I still prefer this because now `reverter_sk` is in `L1SenderConfig` where it arguably doesn't belong.

node/bin/src/config/mod.rs
```diff
@@ -774,6 +774,13 @@ pub struct RebuildBlocksConfig {
     pub reset_timestamps: bool,
 }
 
+#[derive(Clone, Debug, DescribeConfig, DeserializeConfig)]
+pub struct L1RevertConfig {
+    /// Highest L1 batch number to keep on the settlement layer.
+    /// All batches with number greater than this value will be reverted.
+    pub last_l1_batch_to_keep: u64,
```
RomanBrodetski: so, now to revert a block that's already committed one needs to know the batch number that it ended up in. Is this intentional?  

The alternative is to ditch `L1RevertConfig` altogether, add `danger_allow_l1_commit_revert: bool` to `RebuildBlocksConfig`, and make server figure out the batch number(s) that needs to be reverted - based on `from_block` value. This way we avoid adding one more config group - and reuse the existent revert-related one.

RomanBrodetski: > if set, node auto-derives `sequencer.block_rebuild.from_block`

just read this - ah interesting, so it works but other way round? You can put `last_l1_batch_to_keep` and then blocks will be rebuilt? 
You know what, there is actually a case for **reverting batch, but keeping blocks as is** - if we want already produced blocks to be reorganized. With current approach I think it's not possible.

So to support both cases I propose:
* remove `L1RevertConfig`
* add `danger_allow_l1_commit_revert: bool` to `L1RevertConfig` with the behavour I decribed above
* **also** add `revert_l1_commits_from_batch_number: Option<u64>`. If set, it will revert L1 commits - but it will not rebuild any blocks.

And then we should probably forbid to pass both `revert_l1_commits_from_batch_number` and `from_block` at the same time.

Please think about it - I may be missing smt here

node/bin/src/l1_revert.rs
```diff
@@ -0,0 +1,177 @@
...
+    let reverted_batch = l1_revert
+        .last_l1_batch_to_keep
+        .checked_add(1)
+        .expect("`sequencer.l1_revert.last_l1_batch_to_keep` overflowed");
+    // Block 0 is genesis and is never executed, so the earliest rebuild point is block 1.
+    // For batch 1 there is no predecessor in storage to look up its first block from.
+    let derived_from_block = if reverted_batch == 1 {
+        1
+    } else {
+        // Fast path: the batch watcher already persisted this batch locally.
```
RomanBrodetski: hm.. but if its already persisted, it means it's already executed on L1 right? If so, we cannot really revert such a  block/batch. So I would rather use `persistent_batch_storage` as a fail-fast path - if that batch is already there, we should refuse to rollback it

node/bin/src/lib.rs
```diff
@@ -1850,6 +1922,20 @@ fn determine_starting_block(
     desired_starting_block
 }
 
+async fn fetch_l1_state(
```
RomanBrodetski: a little bit of hairsplitting but this method can be added to `L1State` directly - no need to dispatch it externally. In general, we should make individual components interface accomodate the code in `lib.rs`, not other way round. The reason for this is that complexity in `lib.rs` is very expensive, as it's the integration layer which all features touch

node/bin/src/lib.rs
```diff
@@ -203,30 +210,97 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+        gateway_provider.as_ref().map(|p| p.clone().erased()),
+        bridgehub_address,
+        chain_id,
+    )
+    .await
+    .expect("failed to fetch L1 state");
+
+    if node_role.is_main() {
+        let l1_revert_config = config.sequencer_config.l1_revert.as_ref();
+        if let Some(l1_revert_config) = l1_revert_config {
```
RomanBrodetski: nit: no need to introduce the local variable above - just inline it

node/bin/src/lib.rs
```diff
@@ -203,30 +210,97 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+            tracing::warn!(
+                last_l1_batch_to_keep = l1_revert_config.last_l1_batch_to_keep,
+                "L1 revert config detected, starting startup revert sequence"
+            );
+
+            let last_l1_batch_to_keep = l1_revert_config.last_l1_batch_to_keep;
```
RomanBrodetski: same - no need for a local variable - please inline it. It has an added benefit that on the line it's accessed, the reader immediately sees that it comes from config - not a result of some computation

node/bin/src/lib.rs
```diff
@@ -203,30 +210,97 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+            assert!(
+                last_l1_batch_to_keep < l1_state.last_committed_batch,
+                "sequencer.l1_revert.last_l1_batch_to_keep ({last_l1_batch_to_keep}) must be \
+                 < current last committed batch ({})",
+                l1_state.last_committed_batch
+            );
```
RomanBrodetski: currently after the node restarts after batch reversial with the same config (default kubernetes behaviour), it:
* if it already committed a new batch with the same number , it will revert it again
* if it didn't commit it yet, it will panic on this assertion and enter crash loop

Note that we have the same problem on the block level (`config.sequencer_config.block_rebuild.from_block`) - if it's set in the config, it will be applied again and again on restarts (until removed from config).

To mitigate it, I suggest:
* as mentioned in another comment, let's infer the batch number to revert from the block number.
* let's add `from_block_hash` to the config - and only revert the `from_block` blocks if it matches the hash. This way on restart we won't be doing that anymore since hash will presumably be new
* for the usecase when we want to revert batches but not blocks (see my other comment on config/mod.rs) do the same - instead of passing the batch number, pass both the number and the hash.

node/bin/src/lib.rs
```diff
@@ -203,30 +210,97 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+                last_l1_batch_to_keep >= l1_state.last_executed_batch,
+                "sequencer.l1_revert.last_l1_batch_to_keep ({last_l1_batch_to_keep}) must be \
+                 >= last executed batch ({})",
+                l1_state.last_executed_batch
+            );
+
+            // Derive `block_rebuild.from_block` before the revert while the to-be-reverted
+            // batch is still committed on L1. The L1 fallback relies on `totalBatchesCommitted`
+            // being >= reverted_batch at the current L1 head, which is no longer true after
+            // the revert tx lands.
```
RomanBrodetski: FWIW we could derive it after the revert as well - it's the block after the last block of `last_l1_batch_to_keep`. But this logic will probably change if we apply other suggstions from this review

RomanBrodetski (changes requested):

node/bin/src/config/mod.rs
vasyl-ivanchuk: The config structure was completely reworked, please check updated PR description. Resolving comment as outdated.

node/bin/src/config/mod.rs
vasyl-ivanchuk: The config structure was completely reworked, please check updated PR description. Resolving comment as outdated.

node/bin/src/l1_revert.rs
vasyl-ivanchuk: Makes sense! Fixed.

node/bin/src/lib.rs
vasyl-ivanchuk: Agreed, done.

node/bin/src/lib.rs
vasyl-ivanchuk: Great suggestion! Implemented for both blocks and batch reverts!

node/bin/src/lib.rs
vasyl-ivanchuk: The config structure was completely reworked, please check updated PR description. Resolving comment as outdated.

integration-tests/tests/node/rebuild.rs
```diff
@@ -222,3 +321,558 @@ async fn rebuild_after_emptying_historical_block_preserves_unrelated_l2_txs(
...
+    let mut config = env.default_config().await?;
+    make_commit_only_config(&mut config);
+    let tester = env.launch(config).await?;
+
+    // Drive at least three committed (still unexecuted) batches onto L1 so batch 2 is a non-last
+    // committed batch (batch 1 below to keep, batches >= 3 above to skip during the scan).
+    mine_until_l1_state(
+        &tester,
+        "at least three committed but unexecuted batches",
+        |state| state.last_committed_batch >= 3 && state.last_executed_batch == 0,
```
RomanBrodetski: hmm so this will **wait** until three batches are committed. Makes sense. But it also says `state.last_executed_batch == 0,` - which should be guaranteed by the config above. 
It doesn't feel right to have it here - if there is a bug and batches **are** executed - this method will time out. A better approach would be to mine until three batches are committed, **and then** assert that no batches are executed. This way if there is a bug, we see it right away and with a clear error - not as a timeout.

RomanBrodetski: Saw the same pattern in some other tests as well

integration-tests/src/test_config.rs
```diff
@@ -7,6 +7,24 @@ use zksync_os_server::config::{Config, ProviderConfig};
...
+}
+
+/// Runs the full settlement pipeline so batches commit, prove, and execute on L1.
+pub fn make_full_pipeline_config(config: &mut Config) {
+    config.prover_api_config.fake_fri_provers.enabled = true;
+    config.prover_api_config.fake_fri_provers.compute_time = Duration::from_millis(200);
+    config.prover_api_config.fake_fri_provers.min_age = Duration::ZERO;
+    config.prover_api_config.fake_snark_provers.enabled = true;
+    config.prover_api_config.fake_snark_provers.max_batch_age = Duration::ZERO;
+}
```
RomanBrodetski, line 26: seems like some methods were extracted to be reused across rebuild + restart, whereas others (like `fetch_l1_state`) are duplicated across the two. 
Let's consolidte - the naive solution would be to have a `helpers.rs`, but maybe we can do better. At least something lie `l1_helpers.rs`

node/bin/src/lib.rs
```diff
@@ -216,36 +223,69 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+            l1_state = L1State::fetch_with_finality(
+                config.batcher_config.enabled,
+                l1_provider.clone(),
+                gateway_provider.clone(),
+                bridgehub_address,
+                chain_id,
+            )
+            .await
+            .expect("failed to fetch L1 state after startup revert");
+        }
```
RomanBrodetski: Hm.. the alternative would be to restart the server at this point. It's good that we re-fetch the l1 state, but I'm uneasy if there are some other invariants that we break by changing the committed l1 batch during the initialization process. 
@itegulov curious what you thinkg - and @vasyl-ivanchuk please think this through by reviewing the code in this file BEFORE this statement (that is, do we already init some components or vars that depend on L1?)

node/bin/src/lib.rs
```diff
@@ -137,6 +139,7 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
     runtime: &Runtime,
     config: Config,
 ) {
+    let mut config = config;
```
perekopskiy:
```suggestion
    mut config: Config,
) {
```

RomanBrodetski (commented):
We'll need to make sure to test it on a Sepolia L1 (assuming you tested on Anvil already). Easiest way to do that is after we merge it - do on stage. Please don't forget 🙌

node/bin/src/rebuild.rs
```diff
@@ -0,0 +1,342 @@
...
+        }
+
+        RebuildConfig::DangerBlockRebuildWithL1Revert {
+            bounds,
+            l1_reverter_sk,
+        } => {
+            if !from_block_hash_matches(repositories, bounds.from_block, bounds.from_block_hash) {
+                config.sequencer_config.rebuild = None;
+                return Ok(false);
+            }
```
perekopskiy: nit: you can deduplicate this code if you match for `BlockRebuild | DangerBlockRebuildWithL1Revert` before the main match

node/bin/src/rebuild.rs
```diff
@@ -0,0 +1,342 @@
...
+
+/// Handles the startup rebuild/revert config and performs L1 revert.
+///
+/// Returns `true` if an L1 revert was performed and the caller should refresh `L1State`.
+/// Returns `false` if no revert was needed (skipped or `BlockRebuild` which never touches L1).
+pub async fn handle_startup_rebuild(
+    config: &mut Config,
+    repositories: &dyn ReadRepository,
+    l1_state: &L1State,
+    persistent_batch_storage: &ExecutedBatchStorage,
```
perekopskiy: Why is batch storage needed? Isn't executed batch number in l1 state enough?

vasyl-ivanchuk: Indeed, it's enough. After the refactoring it's already gone.

node/bin/src/rebuild.rs
vasyl-ivanchuk: this part was refactored, resolving the comment as outdated.

node/bin/src/lib.rs
vasyl-ivanchuk: Applied!

integration-tests/src/test_config.rs
vasyl-ivanchuk: Moved the remaining duplicated fns to `l1_helpers`.

integration-tests/tests/node/rebuild.rs
vasyl-ivanchuk: Makes sense, fixed in rebuild and restart tests.

node/bin/src/lib.rs
vasyl-ivanchuk: I checked and L1 is barely used before it's mutated: it's used to read static identity fields like diamond proxy address, L1 chain id and sl chain id etc. OTOH, restarting the node might be a more resilient approach for future changes. Happy to hear thoughts from @itegulov .

perekopskiy: I think you can get `diamond_proxy_l1`, `l1_chain_id`, `sl_chain_id` pretty easily without creating L1State. Then you can extract `if node_role.is_main() && config.sequencer_config.rebuild.is_some() {...` to a separate method and make it return L1State. This way l1 state in `run` will always be correct one and we won't have to worry that someone uses pre-revert data

lib/internal_config/src/lib.rs
```diff
@@ -13,7 +13,13 @@ use std::sync::{Arc, Mutex};
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct InternalConfig {
     /// Number of the failing block that node wants to empty (causing a reorg).
+    #[serde(default)]
```
slowli: Nit: `#[serde(default)]` is implied for `Option`al fields; it's not necessary to specify it explicitly.

node/bin/src/rebuild.rs
```diff
@@ -0,0 +1,344 @@
...
+        top.first_block_number(),
+        top.last_block_number(),
+    );
+
+    // The first batch (scanning from the top) that starts at or before `from_block` is the
+    // batch containing it — i.e. the first batch to revert; last_to_keep is one below it.
+    if top.first_block_number() <= from_block {
+        return Ok(last_committed_batch - 1);
+    }
+    for batch in (last_executed_batch + 1..last_committed_batch).rev() {
```
slowli: Would binary search not make sense here? (E.g., because the searched batch is expected to be heavily skewed towards `last_committed_batch`.)

node/bin/src/rebuild.rs
```diff
@@ -0,0 +1,344 @@
...
+/// - block missing locally: likely a misconfigured `from_block` (typo / beyond local tip);
+/// - hash changed: the rebuild/revert already ran on a previous startup (the expected case).
+fn from_block_hash_matches(
+    repositories: &dyn ReadRepository,
+    from_block: u64,
+    from_block_hash: alloy::primitives::BlockHash,
+) -> bool {
+    let current_hash = repositories
+        .get_block_by_number(from_block)
+        .ok()
```
slowli: AFAIU, repository errors mostly (always?) correspond to I/O errors etc., which could be a result of node misconfiguration or otherwise something seriously wrong. If so, it could make sense to propagate errors here, with `Ok(None)` result handled as-is.

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,95 @@ pub struct StatusServerConfig {
...
+        /// Signer for `revertBatchesSharedBridge` transactions.
+        #[config(secret, with = SignerConfigDeserializer)]
+        l1_reverter_sk: SignerConfig,
+    },
+    /// Revert committed L1 batches only; local blocks are kept as-is.
+    ///
+    /// Use this when you want to undo L1 commits while keeping the locally-produced blocks.
+    L1Revert {
+        /// First batch to revert. All committed batches >= this number are reverted.
+        /// Must be >= 1 (batch 0 is genesis and cannot be reverted).
```
slowli: Nit: This can be ensured by using `NonZeroU64` (or adding param validation, but `NonZeroU64` is more idiomatic).

node/bin/src/lib.rs
itegulov: I like @perekopskiy's suggestion, if we can encapsulate "mutation" to just that one method/place where rebuild is happening then I think it's worth doing so. The problem with current approach is that `run` is a very large method already and making anything `mut` here makes it harder to reason about flow/logic. It also ensures "temporary" `L1State` cannot be used to initialize some component in the future by accident as it's very easy to add new logic above this `if` and not notice that `l1_state` can change down the line.

lib/internal_config/src/lib.rs
vasyl-ivanchuk: Removed.

node/bin/src/rebuild.rs
vasyl-ivanchuk: Makes sense, changed the implementation to the binary search.

node/bin/src/config/mod.rs
vasyl-ivanchuk: Done!

node/bin/src/rebuild.rs
vasyl-ivanchuk: Agreed, changed the implementation to propagate errors.

node/bin/src/lib.rs
vasyl-ivanchuk: Agreed 👍 . Encapsulated this part to the `fetch_l1_state_with_startup_revert` fn.

node/bin/src/rebuild.rs
```diff
@@ -64,22 +64,34 @@ async fn derive_last_l1_batch_to_keep(
         top.last_block_number(),
     );
 
-    // The first batch (scanning from the top) that starts at or before `from_block` is the
-    // batch containing it — i.e. the first batch to revert; last_to_keep is one below it.
-    if top.first_block_number() <= from_block {
-        return Ok(last_committed_batch - 1);
-    }
```
perekopskiy: nit: I'd keep this `if`, looks like it will be entered in common revert case

perekopskiy (dismissed):

node/bin/src/rebuild.rs
vasyl-ivanchuk: agreed! Brought it back.

perekopskiy (dismissed):

perekopskiy (approved):

lib/l1_watcher/src/committed_batch_provider.rs
```diff
@@ -236,12 +236,39 @@ async fn fetch_batch(
...
-    .await?;
+    .await
+    .with_context(|| format!("failed to find L1 commit block for batch {batch_number}"))?;
+
     util::fetch_stored_batch_data(diamond_proxy_sl, sl_block_with_commit, batch_number)
         .await?
         .with_context(|| format!("failed to find committed batch {batch_number} on L1"))
 }
 
+/// Resolves the hash of the L1 transaction that currently commits `batch_number`.
```
RomanBrodetski: nit: `currently commits `batch_number` - took me a while to understand what that means - first I thought that, like, at this very moment this bartch_number is being committed


```suggestion
/// Resolves the L1 transaction hash of  the Commit transaction of batch `batch_number` (not to be confused with batch header hash itself)
```

lib/l1_watcher/src/revert_watcher.rs
```diff
@@ -0,0 +1,102 @@
...
+    #[test]
+    fn does_not_restart_for_revert_below_startup_block() {
+        assert!(!should_restart_for_revert(100, Some(99)));
+    }
+
+    #[test]
+    fn does_not_restart_when_log_has_no_block_number() {
+        assert!(!should_restart_for_revert(100, None));
+    }
+}
```
RomanBrodetski, line 102: is this tested in int. tests?

local-chains/README.md
```diff
@@ -72,6 +72,25 @@ Commonly modified values include:
...
+- `from_block_hash` — required; both the L1 revert and rebuild are skipped on restart when the hash no longer matches
+- `l1_reverter_sk` — required; signer (private key or GCP KMS) for `revertBatchesSharedBridge`
+- `blocks_to_empty`, `reset_timestamps` — optional, same as `mode: block_rebuild`
+
+**`mode: l1_revert`** — revert committed L1 batches only; local blocks are kept as-is:
+- `from_batch` — required; first batch to revert (>= 1); all committed batches from this number upward are reverted
+- `from_batch_commit_tx_hash` — required; hash of the L1 transaction that committed `from_batch`; acts as an idempotency guard — the revert is skipped unless the transaction currently committing `from_batch` on L1 matches this hash. The simplest way to get the hash is the sequencer's own logs (search for the next line): `commit batch <from_batch> ... succeeded on L1   (tx_hash=0x…)`
+- `l1_reverter_sk` — required; signer (private key or GCP KMS) for `revertBatchesSharedBridge`
+
```
RomanBrodetski: hmmmm not so sure it belongs here... this reads like docs for config - which generally go in doc-comments of config/mod.rs. And I think theya re already there and are sufficient. So unless I'm misssing something - let's remove this file change

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,94 @@ pub struct StatusServerConfig {
...
+/// the starting boundary, the idempotency guard, and what to do during the local block replay.
 #[derive(Clone, Debug, DescribeConfig, DeserializeConfig)]
-pub struct RebuildBlocksConfig {
-    /// Number of the block to start rebuilding from.
-    /// All blocks starting from this number will be replayed - but unlike normal replay,
-    /// we'll not assert that the result will match the original ReplayRecord (block).
-    /// That is, a block may close earlier (with less transactions),
-    /// have different hash, have some transactions rejected etc
+pub struct RebuildBounds {
+    /// First block to rebuild from. All blocks from this number onward are replayed.
```
RomanBrodetski:
```suggestion
    /// First block to rebuild from. All blocks from this number onward are rebuilt.
    /// Depending on further settings and state, they may or may not change
```

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,94 @@ pub struct StatusServerConfig {
...
+    ///
+    /// For `danger_block_rebuild_with_l1_revert` this must fall within a committed-only batch
+    /// (not an executed one). It may be the first block of that batch or any block inside it.
     pub from_block: u64,
-    /// List of blocks to empty (i.e., remove all transactions from).
+    /// Expected hash of block `from_block` before the rebuild.
+    ///
+    /// The operation is skipped when the current hash does not match — prevents double-runs
+    /// on pod restarts after the operation already ran.
```
RomanBrodetski:
```suggestion
    /// The rebuild/revert operation is skipped when the current hash does not match — prevents double-runs
    /// on pod restarts after the operation already ran.
```

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,94 @@ pub struct StatusServerConfig {
     pub address: String,
 }
 
+/// Parameters shared by the `block_rebuild` and `danger_block_rebuild_with_l1_revert` modes:
+/// the starting boundary, the idempotency guard, and what to do during the local block replay.
```
RomanBrodetski: >  what to do during the local block replay.

Wait, what? 🤔

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,94 @@ pub struct StatusServerConfig {
...
+    /// The operation is skipped when the current hash does not match — prevents double-runs
+    /// on pod restarts after the operation already ran.
+    #[config(with = Serde![str])]
+    pub from_block_hash: BlockHash,
+    /// Blocks to empty (remove all transactions from) during the rebuild.
     #[config(default, with = Delimited::new(","))]
     pub blocks_to_empty: Vec<u64>,
-    /// Whether to reset timestamps of rebuilt blocks to the current time.
-    /// If false, original timestamps are kept. If true then current time is used.
+    /// If true, rebuilt blocks receive fresh timestamps instead of the originals.
```
RomanBrodetski, line 783:
```suggestion
    /// If true, rebuilt blocks receive fresh timestamps instead of the originals.
    /// If false, original timestamps are kept
    /// Note: when original timestamps are kept - and no block transactions are excluded, the rebuild process may not amend any blocks - having essentially no effect on chain.
```

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,94 @@ pub struct StatusServerConfig {
...
+/// The `mode` field selects one of three operations:
+///
+/// - `block_rebuild` — replay local blocks from `from_block` without touching L1.
+/// - `danger_block_rebuild_with_l1_revert` — revert committed L1 batches (batch auto-derived
+///   from `from_block`), then rebuild local blocks from the same boundary.
+/// - `l1_revert` — revert committed L1 batches only; local blocks are kept as-is.
+#[derive(Clone, Debug, DescribeConfig, DeserializeConfig)]
+#[config(tag = "mode", rename_all = "snake_case")]
+pub enum RebuildConfig {
+    /// Replay local blocks from `from_block` without any L1 interaction.
```
RomanBrodetski:
```suggestion
    /// Replay and rebuild local blocks from `from_block` without any L1 interaction.
```

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,94 @@ pub struct StatusServerConfig {
...
+    /// If true, rebuilt blocks receive fresh timestamps instead of the originals.
     #[config(default_t = false)]
     pub reset_timestamps: bool,
 }
 
+/// Startup block/batch revert configuration.
+///
+/// The `mode` field selects one of three operations:
+///
+/// - `block_rebuild` — replay local blocks from `from_block` without touching L1.
```
RomanBrodetski:
```suggestion
/// - `block_rebuild` — replay local blocks from `from_block` without touching L1. In this mode, L1-committed blocks cannot be rebuilt.
```

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,94 @@ pub struct StatusServerConfig {
...
-    /// All blocks starting from this number will be replayed - but unlike normal replay,
-    /// we'll not assert that the result will match the original ReplayRecord (block).
-    /// That is, a block may close earlier (with less transactions),
-    /// have different hash, have some transactions rejected etc
+pub struct RebuildBounds {
+    /// First block to rebuild from. All blocks from this number onward are replayed.
+    ///
+    /// For `danger_block_rebuild_with_l1_revert` this must fall within a committed-only batch
+    /// (not an executed one). It may be the first block of that batch or any block inside it.
     pub from_block: u64,
```
RomanBrodetski: let's maybe make it `from_block_number`? To avoid confusion with `from_block_hash`

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,94 @@ pub struct StatusServerConfig {
...
+/// - `l1_revert` — revert committed L1 batches only; local blocks are kept as-is.
+#[derive(Clone, Debug, DescribeConfig, DeserializeConfig)]
+#[config(tag = "mode", rename_all = "snake_case")]
+pub enum RebuildConfig {
+    /// Replay local blocks from `from_block` without any L1 interaction.
+    BlockRebuild {
+        #[config(flatten)]
+        bounds: RebuildBounds,
+    },
+    /// Revert committed L1 batches, then rebuild local blocks from the same boundary.
```
RomanBrodetski:
```suggestion
    /// Same as `BlockRebuild`, but also reverts L1 batch commits of `from_block` and subsequent blocks. Panics if `from_block` is not committed yet - use `BlockRebuild` instead.
```

> Panics if `from_block` is not committed yet - use `BlockRebuild` instead.

Is this the current logic? Or do we only revert if needed? I think in this case it's better to panic - batch reversal is a destructive operation and should only be used carefully - one shouldn't invoke it unless it's explicitely needed (just use normal `Blockrebuild` instead)

node/bin/src/config/mod.rs
```diff
@@ -760,23 +760,94 @@ pub struct StatusServerConfig {
...
+        #[config(secret, with = SignerConfigDeserializer)]
+        l1_reverter_sk: SignerConfig,
+    },
+    /// Revert committed L1 batches only; local blocks are kept as-is.
+    ///
+    /// Use this when you want to undo L1 commits while keeping the locally-produced blocks.
+    L1Revert {
+        /// First batch to revert. All committed batches >= this number are reverted.
+        from_batch: NonZeroU64,
+        /// Hash of the L1 transaction that committed batch `from_batch`.
```
RomanBrodetski:
```suggestion
        /// Hash of the L1 transaction that committed batch `from_batch`. Not to be confused with L1 batch header hash
```

local-chains/README.md
RomanBrodetski: On a second thought - this configuration is reach and complex enough that additional documentation doesn't hurt. I just don't think it belongs to `local-chains`. And also currently it reads as alterntive fields description - which may still be helpful... Actually rteading further I like these more than the ones in config/mod.rs - but I already commented some suggestions there.

Anyway, my final proposal is to remove these and add some **usage examples** to config/mod.rs - that is, bundles of compatible configs grouped by use-case (e.g. unprovable already committed batch; revm divergency). Feel free to push back - this is just a suggestion.

node/bin/src/rebuild.rs
```diff
@@ -0,0 +1,350 @@
...
+        return Ok(false);
+    };
+    let chain_id = config
+        .genesis_config
+        .chain_id
+        .context("`genesis.chain_id` is required for startup rebuild")?;
+    let max_blocks = config.l1_watcher_config.max_blocks_to_process;
+
+    match plan_startup_rebuild(&rebuild, repositories, l1_state, max_blocks).await? {
+        RebuildAction::SkipAndClearConfig => {
```
RomanBrodetski: I'm not 100% sure this is a good place to handle `            config.sequencer_config.rebuild = None;`:
* this file generally does l1-related things
* we mutate config mid-startup - may be confusing - imaginу we inspect this config elsewhere later on - it's be surprising to see it None even though user passed it as Some. Generally, passing `&mut` to a function is a mild code smell, and I think here it's the case where I'd like to fight it.

Ideally, we'd have two stages of revert handling: first is purely for l1 and second is purely for l2 blocks. In `lib.rs` (or dedicated function) we'll first apply the `from_block_hash_matches` logic by something like `let rebuild_config = if from_block_hash_matches() {None} else {Some(config.rebuild)}`, then do `let l1_was_reverted = handle_startup_rebuild(rebuild_config)` (note - no need for `repositories` anymore!) and then do the rest like we do now. This way we'd also be able to get rid of `RebuildAction::SkipAndClearConfig` in this file - and maybe of `RebuildAction` alttogether, allowing for further simplification.

The way it's done now is not really bad - but I think this suggestion will further improve it. You can try to give it to claude and just see how the result is. I will not insist on it further

RomanBrodetski (commented):
Overall looks much cleaner! Please consider my suggestions

local-chains/README.md
vasyl-ivanchuk: I agree that rebuild docs are not that relevant for local configs. Removed this section and added examples to the config module.

node/bin/src/config/mod.rs
vasyl-ivanchuk: improved the wording :)

node/bin/src/config/mod.rs
vasyl-ivanchuk: Yes, it will be better! I've also changed `from_batch` to `from_batch_number` for consistency.

node/bin/src/config/mod.rs
vasyl-ivanchuk: yes, the node panics if `from_block_number` is not committed yet.

lib/l1_watcher/src/revert_watcher.rs
vasyl-ivanchuk: It's not a real use case. The only reason `block_number` is an `Option<u64>` in `should_restart_for_revert` is that alloy's log `block_number` is passed to the function, and it is an `Option`. I saw that we did the same thing in a different watcher, so I followed that pattern.

As for the integration tests, I've added one for EN restart on a `BlockRevert` event.

node/bin/src/rebuild.rs
vasyl-ivanchuk: Like it! I've refactored it the way you suggested and it really looks cleaner with better separation of concerns.

node/bin/src/l1_revert.rs
```diff
@@ -0,0 +1,289 @@
...
+        Some(RevertPlan {
+            last_l1_batch_to_keep,
+            l1_reverter_sk,
+        }) => {
+            perform_l1_revert(
+                last_l1_batch_to_keep,
+                l1_state,
+                chain_id,
+                sl_provider,
+                &l1_reverter_sk,
```
RomanBrodetski: usage of `RevertPlan` seems arbitrary - we return it from `plan_l1_revert` as strcut, then we deconstruct it, and then pass individual fields to `perform_l1_revert` - whereas all methods are in the same file.
My suggestion is to get rid of `RevertPlan` altogether - I believe this is just a lefrover from a more elboarte struct in one of the previous review cycles

RomanBrodetski: nah, maybe keep it, but pass as a whole to `platform_l1_revert`

node/bin/src/lib.rs
```diff
@@ -204,37 +210,73 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+        config.general_config.rocks_db_path.join(REPOSITORY_DB_NAME),
+        &genesis,
+    )
+    .await;
+
+    // Apply the `from_block_hash` idempotency guard once, up front, to derive the *effective*
+    // rebuild config used for the rest of startup. If the configured rebuild already ran on a
+    // prior startup (hash no longer matches), it is dropped here so BOTH the L1 revert (stage 1,
+    // below) and the local block rebuild (stage 2, downstream) are skipped.
+    let rebuild = if node_role.is_main() {
```
RomanBrodetski: nit: rename to `rebuild_config`

RomanBrodetski (dismissed):
Only minor comments - apart from that, let's merge!
And let's test this on real environments please

node/bin/src/l1_revert.rs
vasyl-ivanchuk: Updated!

node/bin/src/lib.rs
vasyl-ivanchuk: agreed, updated!

vasyl-ivanchuk:
> Only minor comments - apart from that, let's merge! And let's test this on real environments please

Absolutely! I will test all modes on stage once deployed.

RomanBrodetski (approved):

PR #1337, opened 2026-05-27
feat: verify execute consistency in persist batch watcher

Artemka374:
## Summary

* `TreeBlockCacher` - new pipeline component that runs only on external node. Sits right between `TreeManager` and `BatchVerifier`. Caches the output of tree and as output sends the notification to `BatchVerifier`. Cache is being cleaned up after commited batches are verified against local replays.
* `BatchVerifier` - local `block_cache` native to this component was removed and made a separate pipeline component. Now it accesses the cache from it.
* `L1PersistBatchWatcher` - now also accesses the cache if running on external node. Also, if running on external node, now verifies the consistency of the batch coming from L1/GW against locally replayed blocks. On execute compares the commit data and execute data(as verification is done on commit events)
* Startup - if external node, now can start from latest unpersisted block to be able to cache the required blocks for `L1PersistBatchWatcher` for batch reconstrunction.

Later on(I want to make it a separate PR, as this one is not really small and it will involve touching block executor) - we will be able to disable most of the transaction watchers to rely on this verification instead.


lib/batch_verification/src/verifier/mod.rs
```diff
@@ -197,12 +184,12 @@ impl<Finality: ReadFinality, ReadState: ReadStateHistory> PipelineComponent
...
-                            let block_timestamp = tree_block.record.block_context.timestamp;
-                            self.block_cache.insert(block_number, tree_block)?;
-                            state_reporter.record_processed(block_number, Some(block_timestamp), None);
+                            if let Some((start, end)) = self.local_batch_data_cache.range() {
+                                BATCH_VERIFICATION_RESPONDER_METRICS.update_cache_range(start, end);
+                            }
+                            state_reporter.record_processed(notification.block_number, Some(notification.block_timestamp), None);
                         }
                         None => return Ok(()),
                     }
```
kostia244, line 195: do we need this? Are the cache-related`BatchVerificationResponderMetrics` even needed if we no longer have the cache here?
This looks like it can be simplified

kostia244: In fact. Does it have to be a pipeline component at all? It doesn't really use the input nor provide an output.

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -6,14 +6,17 @@ use alloy::sol_types::SolEvent;
...
+use zksync_os_contract_interface::models::DACommitmentScheme;
 use zksync_os_contract_interface::settlement_layer_intervals::{
     IntervalSettlementLayer, SettlementLayerIntervals,
 };
 use zksync_os_provider::NodeProvider;
 use zksync_os_storage_api::{PersistedBatch, WriteBatch};
+use zksync_os_tree_block_cache::{LocalBatchBlockData, LocalBatchDataCache};
+use zksync_os_types::PubdataMode;
 
 /// Watches finalized commit and execute events together and persists only irreversibly executed
```
kostia244, line 21: Please update the comments (in all places where relevant)

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -30,11 +33,18 @@ use zksync_os_storage_api::{PersistedBatch, WriteBatch};
...
     last_processed_commit_batch: u64,
     last_persisted_batch_on_start: u64,
 }
 
+#[derive(Debug)]
+struct PendingCommittedBatch {
+    discovered: DiscoveredCommittedBatch,
+    extended_info: ExtendedCommitBatchInfo,
+}
+
```
kostia244, line 48: This does a fair bit of data duplication.
`DiscoveredBatch` is `StoredBatchInfo + block range`.

`ExtendedCommitBatchInfo` has `into_stored` method to get a `StoredBatchInfo`.
As for block range, ExtendedBatchInfo has the following:
```
 // Note, that pre-zksync-os-v30 batches did not contain this field.
 pub first_block_number: Option<u64>,
 // Note, that pre-zksync-os-v30 batches did not contain this field.
 pub last_block_number: Option<u64>,
```
I think(double check if this is fine) that you can just set them to Some if they are None. And store only the extended_info here.

Whether you follow my exact suggestion or not, I advise to do some form of refactor here.

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -156,19 +168,152 @@ impl<BatchStorage: WriteBatch> L1PersistBatchWatcher<BatchStorage> {
...
         })
     }
 
+    async fn verify_committed_batch(&self, pending: &PendingCommittedBatch) -> anyhow::Result<()> {
+        let discovered = &pending.discovered;
+        let Some(cache) = &self.local_batch_data_cache else {
+            tracing::debug!(
+                batch_number = discovered.number(),
+                "local batch data cache is not configured; skipping local batch reconstruction"
+            );
```
kostia244, line 263: How about we have 1 info log during construction instead of debug log for every single batch?

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -156,19 +168,152 @@ impl<BatchStorage: WriteBatch> L1PersistBatchWatcher<BatchStorage> {
...
+    fn pubdata_mode_from_scheme(scheme: DACommitmentScheme) -> anyhow::Result<PubdataMode> {
+        match scheme {
+            DACommitmentScheme::EmptyNoDA => Ok(PubdataMode::Validium),
+            DACommitmentScheme::BlobsZKsyncOS => Ok(PubdataMode::Blobs),
+            DACommitmentScheme::BlobsAndPubdataKeccak256 => Ok(PubdataMode::Calldata),
+            DACommitmentScheme::None | DACommitmentScheme::PubdataKeccak256 => {
+                anyhow::bail!("unsupported DA commitment scheme for local verification: {scheme:?}")
+            }
+        }
+    }
```
kostia244, line 197: I thought we have Into<> and From<> traits for this?
Or at least put it outside of `impl L1PersistBatchWatcher`?

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,190 @@
...
+                return;
+            }
+
+            for _ in first_block..=(block_number - 1).min(last_block) {
+                self.data.pop_front();
+            }
+
+            if !self.data.is_empty() && first_block <= block_number {
+                self.first_block = Some(block_number);
+            }
```
kostia244: Shouldn't you set first_block to None otherwise?
I believe .range() can panic if first_block = Some but data.len() = 0.

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,190 @@
...
+    pub output: BlockOutput,
+    pub record: ReplayRecord,
+    pub tree_output: TreeBatchOutput,
+    pub multichain_root: B256,
+}
+
+#[derive(Debug, Default)]
+struct LocalBatchDataCacheInner {
+    cache: TreeBlockCache<LocalBatchBlockData>,
+}
```
kostia244: Why have this? There is no impl blocks for it or anything? By not having this struct you will reduce the PR size without any side effects.

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /// NOTE: this method is called by user and should be called after careful consideration to not clean up
+    /// the cache that can still be used
+    pub fn remove_lower_than(&self, block_number: u64) {
+        self.inner
+            .lock()
+            .expect("local batch data cache mutex poisoned")
+            .cache
+            .remove_lower_then(block_number);
```
kostia244: nit: lower_than vs lower_then

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    pub fn get_range(
```
kostia244, line 145: nit: why is every function in innner, but this one here? Does it operate on a different layer of abstraction?

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,190 @@
...
+        let start = *range.start();
+        let end = *range.end();
+        let guard = self
+            .inner
+            .lock()
+            .expect("local batch data cache mutex poisoned");
+
+        let Some((first_block, last_block)) = guard.cache.range() else {
+            return Ok(None);
+        };
```
kostia244: I don't understand why you want to return Ok in this case. This really needs to be documented

node/bin/src/node_state_on_startup.rs
```diff
@@ -11,6 +11,7 @@ pub struct NodeStateOnStartup {
     pub block_replay_storage_last_block: u64,
     pub tree_last_block: u64,
     pub repositories_persisted_block: u64,
+    pub first_unpersisted_batch_block: u64,
```
kostia244, line 14: maybe it's just me but I find this name hard to read. It also goes against the convention: everything here is "last" this is "first"

node/bin/src/lib.rs
```diff
@@ -1812,6 +1836,13 @@ fn determine_starting_block(
             // For compacted state, we need to replay all blocks that were not persisted yet.
             // For FullDiffs state (default) - this is always ahead of `last_l1_executed_block`.
             state.block_range_available().end() + 1,
+            // EN batch persistence verifies newly persisted batches against locally replayed
+            // batch data, so it must rebuild every not-yet-persisted batch from the first block.
+            if node_startup_state.node_role.is_external() {
+                node_startup_state.first_unpersisted_batch_block
+            } else {
+                u64::MAX
+            },
```
kostia244, line 1845: So what happens on the main node, if we haven't verified the last batch before restart/rotation?

lib/tree_block_cache/src/pipeline_step.rs
```diff
@@ -0,0 +1,74 @@
...
+pub struct TreeBlockCacher<ReadState> {
+    cache: LocalBatchDataCache,
+    read_state: ReadState,
+}
+
+impl<ReadState> TreeBlockCacher<ReadState> {
+    pub fn new(cache: LocalBatchDataCache, read_state: ReadState) -> Self {
+        Self { cache, read_state }
+    }
+}
```
kostia244, line 35: I understand this is necessary but I hope that we can detach batch verification from pipeline and get rid of it eventually.

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -156,19 +168,156 @@ impl<BatchStorage: WriteBatch> L1PersistBatchWatcher<BatchStorage> {
...
+
+        let blocks = cache.wait_for_range(discovered.block_range.clone()).await?;
+        let local_batch = Self::reconstruct_local_batch(
+            pending.discovered.number(),
+            Self::pubdata_mode_from_scheme(
+                pending.extended_info.commit_info.l2_da_commitment_scheme,
+            )?,
+            pending.extended_info.commit_info.sl_chain_id,
+            &blocks,
+        )?;
```
kostia244, line 276: I don't understand? Aren't you reconstructing it using l1-fetched data? You are using ExtendedCommitBatchInfo to build ExtendedCommitBatchInfo, no?

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -231,7 +380,21 @@ impl<BatchStorage: WriteBatch> L1PersistBatchWatcher<BatchStorage> {
...
+                .map_err(L1WatcherError::Other)?;
+
             self.committed_batches.insert(batch_number, committed_batch);
+
+            if let Some(cache) = &self.local_batch_data_cache {
+                // It should be safe to clean cache for already committed blocks,
+                // as the only other component that depend on cache is batch verifier
+                // which needs uncommited batches
+                cache.remove_lower_than(last_block_number + 1);
+            }
```
kostia244, line 396: This is the only use case for the cache here?

kostia244 (commented):
A lot of these are nits.
The code really needs more comments. I also think some structs are redundant.

lib/l1_watcher/src/persist_batch_watcher.rs
kostia244: Nevermind, I got confused here, because it takes a lot of arguments from "pending".
nit: Maybe just pass pending as an argument

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,221 @@
...
+        if self.data.is_empty() {
+            None
+        } else {
+            self.first_block
+                .map(|block| (block, block + self.data.len() as u64 - 1))
+        }
+    }
+
+    /// Removes all blocks lower than the given block number.
+    pub fn remove_lower_then(&mut self, block_number: u64) {
```
slowli: Typo: Should be `remove_lower_than` (you use the correct version in other names).

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,221 @@
...
+        if let Some((first_block, last_block)) = self.range() {
+            if block_number <= first_block {
+                return;
+            }
+
+            for _ in first_block..=(block_number - 1).min(last_block) {
+                self.data.pop_front();
+            }
```
slowli: Nit: I'd think [`VecDeque::drain()`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.drain) would be more efficient here ([`truncate_front`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.truncate_front) certainly is, but it's unstable).

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+impl<Data> Default for TreeBlockCache<Data> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
slowli: Bikeshedding: inline `mod tests` is usually placed at the end of a file; it's sort of weird to find more items after it.

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub output: BlockOutput,
+    pub record: ReplayRecord,
+    pub tree_output: TreeBatchOutput,
+    pub multichain_root: B256,
+}
+
+#[derive(Debug, Default)]
+struct LocalBatchDataCacheInner {
+    cache: TreeBlockCache<LocalBatchBlockData>,
+}
```
slowli: Nit: What's the point of another level of indirection here? Why not use `TreeBlockCache` directly in `LocalBatchDataCache`?

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,221 @@
...
+
+/// Shared cache used by EN batch verification and L1 batch persistence.
+#[derive(Clone, Debug, Default)]
+pub struct LocalBatchDataCache {
+    inner: Arc<Mutex<LocalBatchDataCacheInner>>,
+    notify: Arc<Notify>,
```
slowli: Bikeshedding: IMO, using `Arc<watch::Sender>` would be more straightforward. Internally, a `watch` channel *is* essentially a mutex + `Notify`.

lib/tree_block_cache/src/cache.rs
```diff
@@ -0,0 +1,221 @@
...
+            anyhow::bail!(
+                "requested local batch data range {start}..={end} was already evicted; cached range is {first_block}..={last_block}"
+            );
+        }
+        if end > last_block {
+            return Ok(None);
+        }
+
+        let mut result = Vec::with_capacity((end - start + 1) as usize);
+        for block_number in start..=end {
```
slowli: Nit: It'd be more efficient to iterate over a range of blocks using [`VecDeque::range()`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.range).

slowli (commented):
To understand the approach better: Is it correct that the cache is inserted into the pipeline solely because it would be difficult to recreate block data necessary to compute `ExtendedCommitBatchInfo::build()` data otherwise? I.e., a potential alternative would be to load `LocalBatchBlockData` for the given block number from storage on demand. IIUC, `record`, `tree_output` and `multichain_root` fields from `LocalBatchBlockData` can be straightforwardly loaded from the storage:

- `multichain_root` is *already* loaded from the storage
- `record` can be obtained from the replay records repo
- `tree_output` can be obtained from the Merkle tree.

So the only issue is `LocalBatchBlockData.output`, which is only used as an input to `ExtendedCommitBatchInfo::build()`, which doesn't seem to use much from it; it uses basic block header data (number, timestamp, hash), transaction receipts (more precisely, L2-to-L1 logs), and block pubdata. All of this other than block pubdata is already persisted. Would persisting pubdata be a dumb idea?

Artemka374:
On cache/pubdata persisting: I did consider it, but I think the biggest issue here is memory. So the default pubdata bytes limit we set in config is 110_000 bytes per block, considering block timeout is 250ms it will be ~1.6GB/hour overhead in the worst case. Also, it is relevant only for external node(well, main node can still become one during the rotation), and the cache existed before anyway, so I found making cache more general a bit more convenient in this case. But happy to reconsider if you think this tradeoff is okay for us, I'm not sure about our memory requirements

lib/tree_block_cache/src/pipeline_step.rs
Artemka374: I agree, it makes sense, but I'd better do it separately(as a follow-up), this PR is already a bit bloated

lib/tree_block_cache/src/cache.rs
Artemka374: this is pure AI slop when I asked it to fix a bug, will update

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -156,19 +169,157 @@ impl<BatchStorage: WriteBatch> L1PersistBatchWatcher<BatchStorage> {
...
+                        &block.output,
+                        block.record.transactions.as_slice(),
+                        &block.tree_output,
+                    )
+                })
+                .collect(),
+            last.record.block_context.chain_id,
+            batch_number,
+            pubdata_mode,
+            sl_chain_id,
```
kostia244, line 243: Dumb question: why sl_chain_id instead of first.record.block_context.chain_id?
From what I see you never really compare local_sl_chain_id to discovered_sl_chain_id? 
If you're building a local version, shouldn't you be using local sl_chain_id here?

slowli:
> it will be ~1.6GB/hour overhead in the worst case

This sounds pretty bad, so IMO not worth it if there are no use cases for persisted pubdata (like getting it via RPC, IDK). As a possible option, pubdata could be pruned once it's not needed any longer, although it complicates the design somewhat.

Also, I wonder how much pubdata is generated per block in practice. It seems that e.g., on the genlayer-testnet it's generally about 0.5 KB / block.

PR #1345, opened 2026-05-29
feat(l1-sender): retry override

Artemka374:
## Summary

This PR extends provider with `RetryOverride` feature, with which you can override existing retry config at call site.

RetryOverride allows overriding the following properties:
* `limit` - limit of retries
* `backoff` - custom backoff duration
* `retry_all_errors` - allow to retry on all types of errors, not only classified as retryable
* `on_retry` - callback, which allows you to specify an action you want to do on retry

`sendTransaction` calls in L1 sender are now using infinite retries, with `rpc_call_timeout_critical` time before emitting a metric `long_rpc_call` which should result in alert later on.


node/bin/src/config/mod.rs
```diff
@@ -1182,6 +1186,10 @@ pub struct GatewaySenderConfig {
     /// Maximum time to wait for a Gateway transaction to be included.
     #[config(default_t = 600 * TimeUnit::Seconds)]
     pub transaction_timeout: Duration,
+
+    /// Critical elapsed time for a still-retrying Gateway sender RPC transaction submission.
+    #[config(default_t = 10 * TimeUnit::Minutes)]
+    pub rpc_call_timeout_critical: Duration,
```
kostia244: I don't really think this belongs here. Why not set the exact value in alerts.yaml and simply report the time at which we retry or something of that sort?

Alerts can be modified without a restart, server configs cannot.
This can also be misleading because alerts are not being refreshed all the time, the alerts file has a refresh interval, which might conflict with your value.
Overall I can't think of a good reason to keep it here.

lib/alloy_ext/src/retry.rs
```diff
@@ -0,0 +1,131 @@
...
+
+/// Per-call retry limit override consumed by the node provider retry middleware.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RpcRetryLimit {
+    /// Retry at most this many times after the initial attempt.
+    Attempts(u32),
+    /// Keep retrying retryable errors until the call succeeds.
+    Infinite,
+}
+
```
kostia244: Nit: I really don't like this, we could use u32::MAX and get the same functionality, but with less abstractions. IMO this is overengineered.

Are there any cons to using u32::MAX here?

lib/l1_sender/src/lib.rs
```diff
@@ -243,7 +247,45 @@ where
...
+                                    L1_SENDER_METRICS.long_rpc_call
+                                        [&(command_name, event.call_name)]
+                                        .inc_guard(1)
+                                });
+                            });
+                        with_scoped_rpc_retry_override(
+                            retry_override,
+                            self.provider.send_transaction(tx_request),
+                        )
+                        .await?
```
kostia244, line 254: Is this logic meant for all write RPCs? If so, why can't we insert it into the retry layer instead of creating an override framework?

kostia244 (commented):
Overall I think we could simplify a couple things. 

As for the retry override logic, I guess the implementation is fine for what alloy lets us do. But I am not sold on this. I would be happy to have a simpler solution.(or be convinced why we can't have a simpler solutoin)
Perhaps:
1. Add special sender config for retries or
2. Add a config that would enable infinite retries specifically for the config.
etc

As for metrics I think they could also be integrated, I would strongly suggest to set the threshold in the alert.
You could for example track "submission timestamp" and make alerts based on time() - this_timestmamp > threshold. This way you don't even have to update it dynamically.

lib/l1_sender/src/lib.rs
Artemka374: I'd like to keep it scoped only to this call for now, but in the future it might change. If it comes to the fact that we keep this logic everywhere - happy to remove then, but for now I'd leave it customizable as here

Artemka374:
> Add special sender config for retries or

This feature is not only scoped to sender, as I believe it can be described as part of "L1 resilience" project

> Add a config that would enable infinite retries specifically for the config.
etc

I'd like to keep it scoped to a call site for now, I feel like then every usage of this feature can be more thought-through

P.S. I did simplify a few things though, as e.g. metrics part doesn't look good indeed, maybe it makes sense to leave a generic logging for reaching some amount of retries and rely on general alerts to tell us that something is wrong

itegulov:
Can you clarify what issue we are trying to address in this specific PR? Specifically why do we want to limit retry override to this one place? It feels like limiting it to one place will make the setup brittle as we can have writes in a few different places (e.g. we already have extra one in base token adjuster). Overall I share similar feelings to what @kostia244 expressed above and curious to hear out justification

PR #1346, opened 2026-05-29
feat(watcher): Add recent logs cache for L1 watchers

kostia244:
## Summary
Currently we have several watchers that are monitoring new blocks. They do so by calling eth_getLogs for all new blocks. Which is inefficient in terms of RPC usage. We do many RPC calls for the same block. It's especially a problem for our 2FA ENs, we have a lot of them and the cost scales.

This PR introduces a `LogCache` which caches the logs for recent blocks. It exposes the same api as alloy's provider.get_logs & falls back to alloy in case of errors or missing blocks. Watchers now use it instead of provider for logs. This reduces the watcher-related RPC calls to a constant of 2/block instead of watcher_count / block. (Can 1 RPC call as a follow up, but this involves refactoring BlockUpdates too).

`LogsCache` uses eth_getLogs to get all logs for a specific block, not filtered by address. The assumption here is that it's fine with all RPC providers(it's find with ethereum, but RPC providers make their own rules).

It is meant to be a stripped-down version of such caching to address the problem asap. Although if you have thoughts/comments about LogCache, BlockUpdates or L1Watcher in general that are out of scope this PR please share them(here or in slack) as I'll be making a couple follow ups here(just mark them as follow ups).


lib/l1_watcher/src/watcher.rs
```diff
@@ -158,7 +164,7 @@ impl L1Watcher {
         if let Some(topic1) = self.processor.topic1_filter() {
             filter = filter.topic1(topic1);
         }
-        let new_logs = self.provider.get_logs(&filter).await?;
```
itegulov, line 162: Can we make caching a part of the provider itself, similarly to what I did in https://github.com/matter-labs/zksync-os-server/pull/1350? That way it will be easier to disable/enable, also carrying one extra struct everywhere is a bit cumbersome

lib/l1_watcher/src/logs_cache.rs
```diff
@@ -0,0 +1,264 @@
...
+                .get_logs(&Filter::new().at_block_hash(block.header.hash))
+                .await?;
+            // TODO: A very rare corner case is introduced here.
+            // We get `block`. Reorg happens before we get `logs`. Some RPCs would return
+            // empty list(instead of proper logs) or an error.
+
+            // Reorg check: our cached parent hash doesn't match the block's parent_hash.
+            let parent_hash_mismatch = has_parent
+                && recent.cached_hash(block_number - 1) != Some(block.header.parent_hash);
+            if parent_hash_mismatch {
```
itegulov, line 293: Do I understand correctly that you rely on head moving exactly one block at a time? This may be true for Ethereum where block time is ~12s but definitely won't hold for Gateway

kostia244: No. We only check parent here because there is recursion. It should be able to handle arbitrary reorgs including new reorgs while we were still reacting to a reorg.

First this function checks that we have parent block if not it is fetched (L229).
Then for this reorg logic: it is also recursive so it will go as deep as the reorg.

lib/l1_watcher/src/watcher.rs
kostia244: It is possible. I thought doing it like this is "safer", because I'm not sure how this & code around it will change.

lib/l1_watcher/src/logs_cache.rs
itegulov: Ah I see, completely misread this logic. I like that you rely on block hash instead of trusting that block number always refers to the same block.

lib/l1_watcher/src/logs_cache.rs
```diff
@@ -0,0 +1,264 @@
...
+                .get_block_by_number(BlockNumberOrTag::Number(block_number))
+                .await?
+                .ok_or_else(|| TransportErrorKind::custom_str("block not found"))?;
+            let logs = self
+                .provider
+                .get_logs(&Filter::new().at_block_hash(block.header.hash))
+                .await?;
+            // TODO: A very rare corner case is introduced here.
+            // We get `block`. Reorg happens before we get `logs`. Some RPCs would return
+            // empty list(instead of proper logs) or an error.
```
itegulov, line 283: This is okay-ish because the propagated error does not trigger panic and instead just clears cache, correct?

itegulov: If so - is this really a todo? I don't see anything actionable here

lib/l1_watcher/src/logs_cache.rs
```diff
@@ -0,0 +1,264 @@
...
+            let has_parent = block_number > floor;
+            if has_parent && recent.cached_hash(block_number - 1).is_none() {
+                self.update_block(recent, block_number - 1, floor).await?;
+            }
+
+            let block = self
+                .provider
+                .get_block_by_number(BlockNumberOrTag::Number(block_number))
+                .await?
+                .ok_or_else(|| TransportErrorKind::custom_str("block not found"))?;
```
itegulov, line 276: Purely food for thought: does it make sense to store latest block hash in `BlockUpdates`? Then this place wouldn't need this extra call

lib/l1_watcher/src/logs_cache.rs
```diff
@@ -0,0 +1,264 @@
...
+            let logs = self
+                .provider
+                .get_logs(&Filter::new().at_block_hash(block.header.hash))
+                .await?;
```
itegulov, line 280: Another idea: ensure two methods above are called in the same JSON-RPC query (you can call multiple methods in one request thus ensuring you request the same exact node even if it's behind LB). Then you can refer to block by its number here instead of hash. Not sure if alloy even gives you a way to do it though, also there is still a race condition between processing two queries as AFAIU multi-query processing does not happen atomically

lib/l1_watcher/src/watcher.rs
itegulov: IMHO it's worth to refactor this

lib/l1_watcher/src/logs_cache.rs
kostia244: Latest is not enough as we have already discussed we might need to insert more than one block. This is one of the changes I am planning to add. Also it's not clear how this will play out with BlockUpdates. I guess we might want to have this owned by the provider to make it cleaner?
I'm not eager to add more complexity around L1 code, so I'd want to think this through.

lib/l1_watcher/src/logs_cache.rs
kostia244: > Then you can refer to block by its number here instead of hash

If the provider or blockupdates supply us with fresh hashes, this would work. Thanks, I like this idea!

lib/l1_watcher/src/logs_cache.rs
kostia244: I'm glad I left this here because I forgot to add the check: I'm worried about the case where instead of an error some wild RPC returns an empty block.
The check to add is that logs match the header.

lib/l1_watcher/src/watcher.rs
kostia244: We put this into provider, but we use BlockUpdates here. BlockUpdates uses provider. So there is a little bit of a cycle.
I think BlockUpdates could also go into provider. (Actually alloy has built in polling for latest, but not finalized).

I guess I'll try to put both of them into provider?
Also, are you fine with this being implicit?

itegulov: Agreed, `BlockUpdates` most likely should be in provider. I think it's detached enough that we can tackle it in a separate PR just to keep the diff contained here

kostia244: I agree about separate PR. So do you want me to try to put just the LogsCache into provider with this PR or move both of them in the follow up?

itegulov: > Also, are you fine with this being implicit?

Yes, IMHO that's part of why we'd want to do it

itegulov: > So do you want me to try to put just the LogsCache into provider with this PR or move both of them in the follow up?

Up to you what's more convenient, I am happy with either if it's temporary

lib/l1_watcher/src/logs_cache.rs
```diff
@@ -0,0 +1,264 @@
...
+            // TODO: A very rare corner case is introduced here.
+            // We get `block`. Reorg happens before we get `logs`. Some RPCs would return
+            // empty list(instead of proper logs) or an error.
+
+            // Reorg check: our cached parent hash doesn't match the block's parent_hash.
+            let parent_hash_mismatch = has_parent
+                && recent.cached_hash(block_number - 1) != Some(block.header.parent_hash);
+            if parent_hash_mismatch {
+                tracing::warn!(block_number, "recent logs cache detected reorg");
```
Artemka374: Nit: do I remember correctly that we agreed not to rely on structured logging? or maybe something changed from the last time I checked

lib/l1_watcher/src/logs_cache.rs
```diff
@@ -0,0 +1,269 @@
...
+            }
+
+            // Reorg check: our cached parent hash doesn't match the block's parent_hash.
+            let parent_hash_mismatch = has_parent
+                && recent.cached_hash(block_number - 1) != Some(block.header.parent_hash);
+            if parent_hash_mismatch {
+                tracing::warn!(block_number, "recent logs cache detected reorg");
+                // Update blocks to match current chain
+                self.update_block(recent, block_number - 1, floor).await?;
+                // Re-fetch this block from the start for the rare case where `blcok` & `logs` got
```
Artemka374:
```suggestion
                // Re-fetch this block from the start for the rare case where `block` & `logs` got
```

lib/l1_watcher/src/logs_cache.rs
```diff
@@ -0,0 +1,269 @@
...
+    ///
+    /// This function does not verify that previous block's hash matches the parent_hash of the new
+    /// one. This should be done by the user.
+    fn push_head(&mut self, number: u64, hash: B256, logs: Vec<Log>) -> TransportResult<()> {
+        if self.capacity == 0 {
+            return Ok(());
+        }
+        while self
+            .latest_block()
+            .is_some_and(|latest_block| latest_block >= number)
```
Artemka374, line 115: dumb question - this is for handling reorgs, do i understand that correctly?

lib/l1_watcher/src/watcher.rs
kostia244: IMO it's much easier to address this in the follow up.

lib/l1_watcher/src/logs_cache.rs
kostia244: I removed TODO/added a check for the corner case: block with this hash gets dropped after reorg, but provider returns empty list instead of error. Please take a look at the change. (Github doesn't render this as outdated tho, which makes me a bit mad.)

Artemka374 (commented):
Overall it looks good to me, I really like the design. Do you know the tradeoffs here though, like how much memory can the cache possibly consume(for 128 blocks e.g. as you set it to default)? I don't think it's a blocker, but a genuine question

lib/l1_watcher/src/logs_cache.rs
kostia244: Yes. I wanted to minimize the API surface for the storage struct, so I added this logic to push_head(that's why it's not push_back) to have a single write method.

kostia244:
I addressed Artem's comments:
use NodeProvider instead of DynProvider (overlooked this at first)
Improve metrics: now they are indexed by chain_id AND we have metrics for memory usage by this structure.

RAM usage: In my tests it's sitting around 40-50MB for both sepolia and mainnet. I believe this should not be a short-term problem realistically. Theoretical limit is higher, probably close to a GB. I would address this as follow-up, to avoid this PR blowing up.

Artemka374:
yeah sure, I don't think the memory part should be a blocker right now

itegulov (approved):

PR #1347, opened 2026-06-01
feat: Support EIP-7702 set-code transactions

dimazhornyk:
## Summary

Accept and execute type-0x04 (set-code) transactions end-to-end: keep Prague active in the L2 mempool validator so authorization lists are applied, require a concrete `to` address for 7702 txs in call / fill_transaction handlers, and validate delegated-sender bytecode. Adds an end-to-end integration test covering delegate-and-call plus a delegated EOA submitting its own transaction.

### DIFF
--- a/lib/mempool/src/subpools/l2.rs
+++ b/lib/mempool/src/subpools/l2.rs
@@ -149,7 +149,10 @@ pub fn in_memory(
         let chain_spec = zk_provider_factory.chain_spec();
         RethPool::new(
             EthTransactionValidatorBuilder::new(zk_provider_factory, EthEvmConfig::new(chain_spec))
-                .no_prague()
+                // EIP-7702 (set-code) transactions require the Prague fork to be active in the
+                // validator. The builder derives this from the chain spec (Prague-activated in
+                // `ZkProviderFactory`), so we must NOT disable it here. `eip7702` already
+                // defaults to enabled.
                 .with_max_tx_input_bytes(validator_config.max_input_bytes)
                 // set tx_fee_cap to 0, effectively disabling the tx fee checks in the reth mempool
                 // this is necessary to process transactions with more than 1e18 tx fee
--- a/lib/mempool/src/subpools/l2.rs
+++ b/lib/mempool/src/subpools/l2.rs
@@ -149,10 +149,10 @@ pub fn in_memory(
         let chain_spec = zk_provider_factory.chain_spec();
         RethPool::new(
             EthTransactionValidatorBuilder::new(zk_provider_factory, EthEvmConfig::new(chain_spec))
-                // EIP-7702 (set-code) transactions require the Prague fork to be active in the
-                // validator. The builder derives this from the chain spec (Prague-activated in
-                // `ZkProviderFactory`), so we must NOT disable it here. `eip7702` already
-                // defaults to enabled.
+                // Keep Prague active (do not call `.no_prague()`): it gates EIP-7702 acceptance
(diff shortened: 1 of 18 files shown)


lib/mempool/src/subpools/l2.rs
```diff
@@ -149,7 +149,10 @@ pub fn in_memory(
         let chain_spec = zk_provider_factory.chain_spec();
         RethPool::new(
             EthTransactionValidatorBuilder::new(zk_provider_factory, EthEvmConfig::new(chain_spec))
-                .no_prague()
+                // EIP-7702 (set-code) transactions require the Prague fork to be active in the
+                // validator. The builder derives this from the chain spec (Prague-activated in
+                // `ZkProviderFactory`), so we must NOT disable it here. `eip7702` already
+                // defaults to enabled.
```
itegulov: What's the repercussion of enabling Prague? AFAIR it contains a few other changes, anything there that can potentially affect `EthTransactionValidator` here?

itegulov: One thing that comes to mind: this might impact intrinsic gas for some transactions

lib/revm_consistency_checker/src/node.rs
```diff
@@ -240,16 +241,39 @@ where
...
+                                // V6 (AtlasV3 -> Prague). Genuine execution errors are still
+                                // propagated.
+                                if is_unsupported_tx_type(&err) {
+                                    PUSH_METRICS.revm_blocks_skipped.inc();
+                                    tracing::warn!(
+                                        block_number = replay_record.block_context.block_number,
+                                        "Skipping REVM consistency check for block, transaction type \
+                                         unsupported by the checker's revm spec: {err}"
+                                    );
+                                    skip_block = true;
```
itegulov: Didn't you make some adjustments to revm checker so that this wouldn't be needed or did I misunderstand?

lib/revm_consistency_checker/src/revm_state_provider.rs
```diff
@@ -14,6 +15,31 @@ fn fixed_bytes_to_bytes32(x: B256) -> Bytes32 {
...
+        && full_bytecode.starts_with(&EIP7702_DELEGATION_PREFIX)
+    {
+        Bytecode::new_raw_checked(Bytes::copy_from_slice(
+            &full_bytecode[..EIP7702_DELEGATION_DESIGNATOR_LEN],
+        ))
+        .expect("valid EIP-7702 delegation designator")
+    } else {
+        Bytecode::new_raw(Bytes::copy_from_slice(full_bytecode))
+    }
+}
```
itegulov, line 26: Nit: looks like same code as in `ZkProvider`. Maybe put it into some common crate like `multivm`?

dimazhornyk: done, unified

lib/revm_consistency_checker/src/node.rs
dimazhornyk: good catch, removed the skip

lib/mempool/src/subpools/l2.rs
dimazhornyk: it is impossible to make it work without enabling Prague

the only thing here is that mempool could reject a low gas calldata heavy tx through (that might theoretically have passed before). i have tested it with all real use-cases, and everything passes correctly, but a specific set of transactions (DA-style, big blobs of data sent to a EOA) it does revert in the mempool when the execution could've passed

this is a valid concern, but all of it can be mitigated on the client side by providing a bigger gas limit

PR #1350, opened 2026-06-02
feat(l1): use deployment block as lower bound for binsearch

itegulov:
## Summary

At startup the `l1_watcher` runs a series of binary searches over L1 history to locate the L1 blocks for the last committed/executed batch, priority transactions, protocol-version upgrades, migrations, and
  interop roots. Each search spanned `[0, latest]` and, on every iteration, issued an extra `eth_getCode` call to guard against probing a block where the contract wasn't deployed yet - roughly **2 RPC calls × 
  ~log₂(latest)** iterations per search, multiplied across ~6 searches.
  
  Most of that range is empty: nothing relevant exists before the chain's contracts were deployed. By anchoring each search at the contract's deployment block, we both shrink the search range and drop the
  per-iteration code-presence guard.

### What changed

  - `l1_watcher`: every binary-search site now starts from the contract's deployment block instead of 0, and `find_l1_block_by_predicate` no longer issues a per-iteration `eth_getCode` guard (the caller guarantees
  `start_block >= deployment_block`). The `start_block` parameter is retained because the commit search's revert-rescan must restart from the revert block, not the deployment block, to find the post-revert commit.
  - `contract_interface`: removed the previously eager deployment-block discovery from `L1State::fetch` / `SettlementLayerIntervals::discover` - discovery is now lazy, on first search.
  - Workspace-wide rename `EthDynProvider` -> `NodeProvider` and removal of `.erased()` calls feeding the contract interface.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15262,10 +15262,7 @@ name = "zksync_os_alloy_ext"
 version = "0.20.5"
 dependencies = [
  "alloy",
- "anyhow",
- "async-trait",
  "serde",
- "serde_json",
  "tokio",
  "tracing",
  "zksync_os_contract_interface",
@@ -15322,6 +15319,7 @@ dependencies = [
  "zksync_os_contract_interface",
  "zksync_os_external_price_api",
  "zksync_os_operator_signer",
+ "zksync_os_provider",
  "zksync_os_types",
 ]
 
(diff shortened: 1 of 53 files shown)


Artemka374 (dismissed):
LGTM

Artemka374 (approved):

PR #1356, opened 2026-06-03
fix: patch computational_native_used calculation

perekopskiy:
## Summary

Patches `computational_native_used` so it properly reflects effective cycles in some specific cases.

### DIFF
--- a/node/bin/src/batcher/seal_criteria.rs
+++ b/node/bin/src/batcher/seal_criteria.rs
@@ -1,3 +1,4 @@
+use alloy::primitives::U256;
 use std::collections::HashSet;
 use zk_ee::{common_structs::MAX_NUMBER_OF_LOGS, system::MAX_NATIVE_COMPUTATIONAL};
 use zksync_os_batcher_metrics::BATCHER_METRICS;
@@ -40,7 +41,7 @@ impl BatchInfoAccumulator {
     }
 
     pub fn add(&mut self, block_output: &BlockOutput, replay_record: &ReplayRecord) -> &Self {
-        self.native_cycles += block_output.computational_native_used;
+        self.native_cycles += computational_native_used(block_output, replay_record);
         self.pubdata_bytes += block_output.pubdata.len() as u64;
         self.l2_to_l1_logs_count += block_output
             .tx_results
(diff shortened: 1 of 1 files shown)


node/bin/src/batcher/seal_criteria.rs
```diff
@@ -181,3 +182,27 @@ impl BatchInfoAccumulator {
...
+// so we re-calculate it as `native_used - pubdata_used * native_per_pubdata`
+// and take maximum of what we computed locally and what zksync os returns.
+//
+// TODO: get rid of method and just use `block_output.computational_native_used` after v30 proving support is removed.
+fn computational_native_used(block_output: &BlockOutput, replay_record: &ReplayRecord) -> u64 {
+    let native_used = block_output
+        .tx_results
+        .iter()
+        .flatten()
+        .map(|tx| tx.computational_native_used)
```
antoniolocascio: I think this should be native_used, no?

node/bin/src/batcher/seal_criteria.rs
```diff
@@ -181,3 +182,27 @@ impl BatchInfoAccumulator {
...
+        .sum::<u64>();
+
+    let native_per_pubdata = replay_record.block_context.pubdata_price
```
antoniolocascio, line 199: There might be a caveat, L1 transactions use a different native_per_pubdata, which is more complicated. @AntonD3 do you think it's important?

AntonD3: Good catch. Yes, for blocks with l1 -> l2 txs this method can underestimate pubdata_native_used a lot

AntonD3: With 0.25 gwei gas price it will be `0.25 * 10^9 / 10 * 800` > 2^34 native per pubdata byte

perekopskiy: Indeed, I think it can overestimate as well (if e.g. block_context.pubdata_price=0)

antoniolocascio: I think we can simplify a bit: for l1 txs always use computational_native_used, for L2 use this. So, in the sum here we do an if and we collect both native used and pubdata used. For L1 txs we don't count pubdata used, so we dont subtract it at the end

antoniolocascio: This is safe, L1 transactions don't have this issue even on the old version (main)

AntonD3: Yes, this is what I mean, native_used for l1 -> l2 txs can be really huge, even close to u64::max, but the current formula will underestimate `pubdata_native_used`(especially with pubdata_price=0), so native_computational_used will be overestimated, it can be really huge(>2^36 for a single block), so batches with l1 txs most likely will often just contain a single block.

perekopskiy: >For L1 txs we don't count pubdata used, so we dont subtract it at the end

the problem is that we don't know how much pubdata->native we used for L1 txs specifically

PR #1357, opened 2026-06-03
refactor(sequencer): do not require starting block in BCP

itegulov:
## Summary

`BlockContextProvider` (BCP) previously had to be primed at startup with a snapshot of the previous block's state - block number, timestamp, block hashes, protocol version, SL chain id, and the
  `sl_chain_id_set` flag. This made initialization fragile and duplicated state that already lives in `ReplayRecord`s.

This refactor makes BCP derive all per-block state from the records it replays. The genesis block (#0) is now replayed to bootstrap that state instead of being skipped, so the node starts from block `0`
  rather than `1`.

### Key changes
  - **BCP holds `last_block: Option<LastBlock>`** instead of a dozen mutable fields. Each replayed/produced block updates it; `produce`/`rebuild` derive the next block's number, timestamp, and `block_hashes` 
  from the previous record.
  - **`prepare_command` now returns `Option<PreparedBlockCommand>`.** Replaying genesis (#0) seeds `last_block` and returns `None` (no block to execute); `BlockExecutor` skips and continues.
  - **Static params grouped into a `Config` struct**, replacing the `#[allow(clippy::too_many_arguments)]` constructor.
  - **SL chain id is derived from `SettlementLayerIntervals`** at construction, dropping the `sl_chain_id_set` / `sl_chain_id_at_startup` / `protocol_version` bookkeeping fields. The one-shot `SetSLChainId` tx
  is now gated on the previous record's protocol version (`< 31` or block `0`).
  - **`genesis_header()` extracted** as a public fn in `zksync-os-genesis` (reused by genesis build and by BCP to compute the genesis block hash).
  - **`BlockHashes::push(hash)`** helper added; inlined hash-ring advancement logic removed.
  - **Startup (`node/bin`) simplified**: starts from block `0`, removes `block_hashes_for_first_block`, and no longer special-cases `starting_block - 1`.

### DIFF
--- a/lib/sequencer/src/execution/block_context_provider.rs
+++ b/lib/sequencer/src/execution/block_context_provider.rs
@@ -1,19 +1,24 @@
 use crate::execution::fee_provider::{FeeParams, FeeProvider};
 use crate::execution::metrics::EXECUTION_METRICS;
-use crate::model::blocks::{BlockCommand, InvalidTxPolicy, PreparedBlockCommand, SealPolicy};
-use alloy::consensus::Sealed;
-use alloy::primitives::{Address, TxHash, U256};
+use crate::model::blocks::{
+    BlockCommand, InvalidTxPolicy, PreparedBlockCommand, RebuildCommand, SealPolicy,
+};
+use alloy::primitives::{Address, BlockHash, TxHash, U256};
 use anyhow::Context as _;
 use futures::StreamExt;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use tokio::{sync::watch, time::Instant};
+use zksync_os_contract_interface::settlement_layer_intervals::{
+    IntervalSettlementLayer, SettlementLayerIntervals,
+};
+use zksync_os_genesis::genesis_header;
 use zksync_os_mempool::subpools::l2::L2Subpool;
 use zksync_os_mempool::{MarkingTxStream, Pool};
+use zksync_os_storage_api::BlockContext;
 use zksync_os_storage_api::ReplayRecord;
-use zksync_os_storage_api::{BlockContext, BlockHashes};
 use zksync_os_types::{
-    BlockOutput, BlockStartCursors, ExecutionVersion, ProtocolSemanticVersion, SystemTxEnvelope,
-    SystemTxType, ZkEnvelope, ZkTransaction,
+    BlockOutput, BlockStartCursors, ExecutionVersion, SystemTxEnvelope, SystemTxType, ZkEnvelope,
+    ZkTransaction,
 };
 
 /// Component that turns `BlockCommand`s into `PreparedBlockCommand`s.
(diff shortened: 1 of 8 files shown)


node/bin/src/lib.rs
```diff
@@ -878,25 +868,22 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+            l1_chain_id: node_startup_state.l1_state.l1_chain_id,
+            gas_limit: config.sequencer_config.block_gas_limit,
+            pubdata_limit: config.sequencer_config.block_pubdata_limit_bytes,
+            fee_collector_address: config.sequencer_config.fee_collector_address,
+            block_time: config.sequencer_config.block_time,
+            service_block_delay: config.sequencer_config.service_block_delay,
+            max_transactions_in_block: config.sequencer_config.max_transactions_in_block,
+            // We set the value to the same as for the batch, since it should be enforced by batcher, but don't want to exceed it for the block
+            interop_roots_per_block: config.batcher_config.interop_roots_per_batch_limit,
+        },
```
itegulov, line 866: I am hoping to refactor this further and fold BCP intialization into sequencer itself. Then create a single sequencer-level config for all of its subcomponents

node/bin/src/lib.rs
perekopskiy, line 1805: will it work for 0?

node/bin/src/lib.rs
```diff
@@ -662,8 +661,8 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
     let interop_roots_subpool =
         InteropRootsSubpool::new(config.sequencer_config.interop_roots_per_tx);
 
-    // If we start from the very first block, we should start by sending upgrade tx for genesis.
-    if starting_block == 1 {
+    // If we start from genesis, we should start by sending upgrade tx for genesis.
+    if starting_block == 0 {
```
perekopskiy, line 665: shouldn't it be <= 1?

lib/sequencer/src/execution/block_context_provider.rs
```diff
@@ -27,376 +32,402 @@ use zksync_os_types::{
...
-                    starting_cursors: self.next_cursors.clone(),
-                    interop_roots_per_block: self.interop_roots_per_block,
-                    strict_subpool_cleanup: false,
-                }
+    async fn rebuild(
+        &mut self,
+        rebuild: Box<RebuildCommand>,
+    ) -> anyhow::Result<Option<PreparedBlockCommand<'_>>> {
+        let (previous_block_timestamp, next_cursors, block_hashes) =
+            if let Some(last_block) = self.last_block.as_ref() {
```
perekopskiy, line 315: Do we even need this if? Can't we just always use else branch?

node/bin/src/lib.rs
```diff
@@ -829,26 +828,17 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
         let prev_record = block_replay_storage
-            .get_replay_record(starting_block - 1)
+            .get_replay_record(starting_block)
             .unwrap_or_else(|| {
                 panic!(
-                    "Missing replay record for block `starting_block - 1` = {}",
-                    starting_block - 1
+                    "Missing replay record for block `starting_block` = {}",
+                    starting_block
                 )
```
perekopskiy: tbh I don't really get this change. Also, looks like we can get rid of `previous_block_fee_params` completely, at least one block must be replayed so we have `previous_block_fee_params` set for first produce command anyway.
And one more different issue, when choosing fee params for block 1 we don't want to take into account genesis hardcoded dummy values, previously it wasn't the case because `if starting_block=1 then previous_block_fee_params=none` but now it should be handled separately I think

lib/sequencer/src/execution/block_context_provider.rs
itegulov: Not if you are rebuilding multiple blocks. Then `starting_cursors` and `block_hashes` might become different from what they were before rebuilding.

node/bin/src/lib.rs
itegulov: We never actually execute genesis (BCP returns `None` for it which skips execution phase in sequencer), added a comment clarifying this

node/bin/src/lib.rs
itegulov: I think it should be `0` now. Our old invariant was a bit strange as _most of the time_ `starting_block` pointed to some existing block and only in extreme circumstances to the next block to be produced (you need to set `config.general_config.min_blocks_to_replay` to `0` essentially and then make sure a bunch of other conditions are met during restart). The only exception was `starting_block=1` on the very first run.

After this refactoring `starting_block` **always** points at a block that we have a replay record for (as at the very least we have genesis).

Returning to your original question `starting_block=1` would mean that block 1 was already executed in the past and we already have a replay record for it. Actually now that I think about it old code technically had a bug where you could accidentally have `starting_block=1` when you already have replay record for it (essentially land on it in `determine_starting_block` by meeting all the conditions there). Then you'd end up with two upgrade transactions. Very rare situation though.

node/bin/src/lib.rs
itegulov: Indeed, all very good points. Addressed in the last commit PTAL. Note that `on_canonical_state_change` is not called for genesis so it shouldn't populate `FeeProvider` anymore.

lib/sequencer/src/execution/block_context_provider.rs
RomanBrodetski: maybe worth adding as code comment?

lib/sequencer/src/execution/block_executor.rs
```diff
@@ -79,11 +79,10 @@ where
             tracing::info!("Command {cmd} received by BlockExecutor");
             let cmd_type = cmd.command_type();
             state_reporter.enter_state(SequencerState::WaitingForApplier);
-            wait_for_block_applier(
-                &mut self.applied_block_number_receiver,
-                self.block_context_provider.next_block_number() - 1,
-            )
-            .await?;
+            if let Some(last_block_number) = self.block_context_provider.last_block_number() {
```
RomanBrodetski, line 82: maybe add a code comment that this is a `None` for the very first block after sequencer starts - which can be `0` in case of genesis.

RomanBrodetski: on a second thought, `last_block_number()` being `None` is pretty self-explonatory..

RomanBrodetski (approved):

lib/sequencer/src/execution/block_context_provider.rs
itegulov: Will add in a follow-up!

PR #1359, opened 2026-06-05
fix(test-contracts): remove passkey factory action hash helper

sidhujag:
## Summary
- Remove the factory-level passkey action hash helper that only matched initial no-sponsor state.
- Keep deployment tests covered by computing the initial action hash locally in the test suite.

## Test plan
- `forge test --match-path test/PasskeySmartAccount.t.sol`


sidhujag:
Opened against the wrong base repository from local gh defaults; recreating against syscoin/zksync-os-server.

PR #1360, opened 2026-06-05
feat(external-node): L1 consistency checker

Artemka374:
## Summary

This PR adds a feature to actually verify L1 data against local replays on external nodes.

* `BatchVerificationResponder` - moved out of pipeline
* `LocalBatchDataCacher` - now is the sink of EN pipeline, which caches block data
* `L1ConsistencyChecker` - user of cache, receives requests for verification(commitment,state_commitment) and verifies that they match locally replayed data
* `PersistBatchWatcher` - now on execute it sends a verification request to `L1ConsistencyChecker` to verify the execute data. If the request is successful - it persists the data.


lib/l1_consistency_checker/src/cache.rs
```diff
@@ -0,0 +1,201 @@
...
+
+impl TreeBlockCache {
+    pub fn new() -> Self {
+        Self {
+            data: VecDeque::new(),
+            first_block: None,
+        }
+    }
+
+    /// Insert a block into the cache. Blocks must arrive in strictly ascending order.
```
kostia244: Nit: strictly ascending means 1 2 3 5 is ok.

lib/l1_consistency_checker/src/cache.rs
```diff
@@ -0,0 +1,139 @@
...
+    first_block: Option<u64>,
+}
+
+impl TreeBlockCache {
+    pub fn new() -> Self {
+        Self {
+            data: VecDeque::new(),
+            first_block: None,
+        }
+    }
```
kostia244: = default

lib/l1_consistency_checker/src/checker.rs
```diff
@@ -0,0 +1,223 @@
...
+            chain_id,
+            sl_chain_id,
+            last_persisted_block_on_start,
+            read_state,
+            cache,
+            latest_verified_batch_tx,
+            l1_events_rx,
+        }
+    }
+}
```
kostia244: I'm leaning towards Not having a constructor here. As we simply pass the variables & it is only invoked once

lib/l1_consistency_checker/src/checker.rs
```diff
@@ -0,0 +1,223 @@
...
+
+impl L1CommittedBatch {
+    pub fn batch_number(&self) -> u64 {
+        self.batch_info.batch_number
+    }
+
+    pub fn last_block_number(&self) -> u64 {
+        *self.range.end()
+    }
+}
```
kostia244: If members are pub, maybe remove the helpers. Second helper's name is longer than the function itself.

lib/l1_consistency_checker/src/checker.rs
```diff
@@ -0,0 +1,223 @@
...
+                        &block.tree_output,
+                    )
+                })
+                .collect(),
+            self.chain_id,
+            commit.batch_number(),
+            PubdataMode::from_da_commitment_scheme(commit.batch_info.l2_da_commitment_scheme),
+            self.sl_chain_id,
+            last_block.multichain_root,
+            &first_block.record.protocol_version,
```
kostia244: Use last block protocol version and you don't need first block?

lib/l1_watcher/src/persist_batch_watcher.rs
```diff
@@ -30,11 +31,17 @@ use zksync_os_storage_api::{PersistedBatch, WriteBatch};
...
-    committed_batches: HashMap<u64, DiscoveredCommittedBatch>,
+    consistency_checker_tx: Option<mpsc::Sender<L1ConsistencyCheckRequest>>,
+    latest_verified_batch_rx: Option<watch::Receiver<u64>>,
+    committed_batches: HashMap<u64, PendingCommittedBatch>,
     last_processed_commit_batch: u64,
     last_persisted_batch_on_start: u64,
 }
 
+struct PendingCommittedBatch {
+    committed_batch: DiscoveredCommittedBatch,
```
kostia244: pass-through type do we really need it?

lib/l1_consistency_checker/src/checker.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    pub fn last_block_number(&self) -> u64 {
+        *self.range.end()
+    }
+}
+
+/// Request to verify that L1 committed batch data matches locally executed blocks.
+pub struct L1ConsistencyCheckRequest {
+    pub commit: L1CommittedBatch,
```
kostia244: pass-through type do we really need it?

lib/l1_consistency_checker/src/cache.rs
```diff
@@ -0,0 +1,139 @@
...
+    ) -> anyhow::Result<Vec<LocalBatchBlockData>>;
+}
+
+#[async_trait]
+impl TreeBlockCacheReceiverExt for watch::Receiver<TreeBlockCache> {
+    async fn wait_for_range(
+        &self,
+        range: RangeInclusive<u64>,
+    ) -> anyhow::Result<Vec<LocalBatchBlockData>> {
+        let mut cache_rx = self.clone();
```
kostia244, line 129: do we need to clone here?

node/bin/src/lib.rs
```diff
@@ -1611,6 +1667,15 @@ fn block_hashes_for_first_block(repositories: &dyn ReadRepository) -> BlockHashe
...
 
+fn read_last_persisted_block_on_start(batch_storage: &impl ReadBatch) -> u64 {
+    let latest_batch = batch_storage.latest_batch();
+    batch_storage
+        .get_batch_by_number(latest_batch)
+        .expect("failed to read latest persisted batch")
+        .map(|batch| batch.last_block_number())
+        .unwrap_or(0)
+}
+
```
kostia244: Add this function to the trait instead?

lib/l1_consistency_checker/src/checker.rs
```diff
@@ -0,0 +1,223 @@
...
+                        "received L1 committed batch {} for consistency checking in range {:?}",
+                        request.commit.batch_number(),
+                        request.commit.range,
+                    );
+                    pending = Some(request.commit);
+                }
+            }
+        }
+    }
+}
```
kostia244: I think there are a couple of problems with l1_events channel:
1. EN sync, we'll be verifying 100s of thousands of blocks:
- you can trigger backpressure or very high memory usage, we do not want this change to increase EN hardware requirements significantly. And definitely don't want to crashloop.
2. Redundancy: I don't know what is the best way to address this, but:
- We already have CommitBatchProvider for recent batches & executed storage for older ones.
- There is Finality, to learn how many executed/committed batches we have discovered.

In short let's make sure we are not destroying performance or introducing too much redundancy. This is kind of open ended.

lib/l1_consistency_checker/src/checker.rs
```diff
@@ -0,0 +1,223 @@
...
+                    .send_if_modified(|latest| {
+                        if batch_number > *latest {
+                            *latest = batch_number;
+                            true
+                        } else {
+                            false
+                        }
+                    });
+                pending = None;
+            }
```
kostia244: I don't understand what are the functional benefits to having this code here instead of in the if at the bottom. I see none, but it makes the code harder to read.

kostia244 (commented):
This is the best version of this you've had so far. I left a couple nits, the most important thing to look at is the fn run() stuff.

lib/l1_consistency_checker/src/checker.rs
Artemka374: > Redundancy: I don't know what is the best way to address this, but:
We already have CommitBatchProvider for recent batches & executed storage for older ones.
There is Finality, to learn how many executed/committed batches we have discovered.

TBH I don't think I understand this point, as I don't think it's particularly related to this change and not much redundancy was introduced here.

> you can trigger backpressure or very high memory usage, we do not want this change to increase EN hardware requirements significantly. And definitely don't want to crashloop.

backpressure can be triggered only on persist batch watcher side, it shouldn't affect the main pipeline(and amount of memory consumed there is relatively small). Although I think the issue might happen in cache itself, as during EN sync cache can grow indefinitely. I'll think more what to do with it

kostia244: > Although I think the issue might happen in cache itself, as during EN sync cache can grow indefinitely.
Yes, let's make sure that there is no unbounded RAM.


> TBH I don't think I understand this point, as I don't think it's particularly related to this change and not much redundancy was introduced here.

You can get the L1 batches from Executed & CommittedBatch storages instead of the channel. It would be slower, but we will not need special channel, we can re-use existing data structures. I do not know if this is a trade-off we should take.

lib/l1_consistency_checker/src/cache.rs
```diff
@@ -0,0 +1,139 @@
...
+    pub fn get_range(
+        &self,
+        range: RangeInclusive<u64>,
+    ) -> anyhow::Result<Option<Vec<LocalBatchBlockData>>> {
+        let Some((first_block, last_block)) = self.range() else {
+            return Ok(None);
+        };
+        if *range.start() < first_block {
+            anyhow::bail!(
+                "requested local batch data range {}..={} was already evicted; cached range is {}..={}",
```
slowli: Can this happen in practice? If this is a sanity check, it could be an `assert!`, which could simplify code (including downstream; e.g., `wait_for_range` can become non-fallible as well and use [`wait_for`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.wait_for) internally.

lib/l1_consistency_checker/src/cache.rs
```diff
@@ -0,0 +1,139 @@
...
+    ) -> anyhow::Result<Vec<LocalBatchBlockData>> {
+        let mut cache_rx = self.clone();
+        loop {
+            {
+                // if block range is available already - return it. If not - wait until the cache is updated and check again
+                if let Some(blocks) = cache_rx.borrow_and_update().get_range(range.clone())? {
+                    return Ok(blocks);
+                }
+            }
+            cache_rx.changed().await?;
```
slowli, line 134: The error here corresponds to the corresponding `watch::Sender` getting dropped. Does it logically make sense to error in this case? A potential alternative would be to hang up indefinitely (via `pending()`), or return a special value, e.g., `Ok(ControlFlow::Break)`.

lib/l1_consistency_checker/src/cache.rs
```diff
@@ -0,0 +1,139 @@
...
+            data: VecDeque::new(),
+            first_block: None,
+        }
+    }
+
+    /// Insert a block into the cache. Blocks must arrive in strictly ascending order.
+    pub fn insert(&mut self, block_number: u64, block: LocalBatchBlockData) -> anyhow::Result<()> {
+        if let Some((_, last_block)) = self.range() {
+            if block_number != last_block + 1 {
+                anyhow::bail!("Out of order block received. This should never happen");
```
slowli: Again, if this is a sanity check, it should be an `assert!`. This would allow simplifying code downstream (in this case, `insert_tree_block()`.

lib/l1_consistency_checker/src/cache.rs
```diff
@@ -0,0 +1,139 @@
...
+        if *range.end() > last_block {
+            return Ok(None);
+        }
+        let start = (*range.start() - first_block) as usize;
+
+        Ok(Some(
+            self.data
+                .iter()
+                .skip(start)
+                .take(range.count())
```
slowli: Nit: It's possible to use [`VecDeque::range()`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.range) here.

lib/l1_consistency_checker/src/checker.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    async fn run(
+        mut self,
+        mut input: PeekableReceiver<Self::Input>,
+        _output: mpsc::Sender<Self::Output>,
+        state_reporter: ComponentStateReporter,
+    ) -> anyhow::Result<()> {
+        tracing::info!("starting L1 consistency checker");
+        // At most one request is held outside the channel at a time; new requests stay queued
+        // in `l1_events_rx` (providing natural backpressure) until this slot is empty again.
```
slowli: Dumb question somewhat mirroring @kostia244's comments: Do we *want* to have sequential processing of requests? An alternative would be spawning verification tasks concurrently once they are received from `l1_events_rx` (details may vary; e.g., it's possible to use Tokio tasks, maybe with a semaphore to limit concurrency, or [`try_for_each_concurrent`](https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_for_each_concurrent) from the `futures` library). Concurrent processing would make cache cleanup more complicated though; one potential solution would be to regress from `VecDeque` for block data to `BTreeMap` and clear ranges in it since each block is required for verifying exactly one batch IIUC.

lib/l1_consistency_checker/src/checker.rs
```diff
@@ -0,0 +1,238 @@
...
+    /// is a pending batch larger than that bound: we must keep caching its blocks, otherwise
+    /// it could never be verified and evicted and the pipeline would deadlock. Once its full
+    /// range is cached it is verified and evicted at the top of the loop, restoring backpressure.
+    fn can_accept_tree_block(&self, pending: Option<&L1CommittedBatch>) -> bool {
+        let cache = self.cache.borrow();
+        if cache.has_capacity() {
+            return true;
+        }
+        match (pending, cache.range()) {
+            (Some(commit), Some((_, last_cached))) => last_cached < commit.last_block_number(),
```
slowli: Can `last_cached >= commit.last_block_number()` in practice given where `can_accept_tree_block()` is called? Subjectively, it'd make sense to have `cache.has_capacity() || pending.is_some()` as the guard in `tokio::select!` so that it's clear that `tokio::select!` cannot deadlock (has at least one hand active at all times).

lib/l1_consistency_checker/src/checker.rs
Artemka374: this is a very good suggestion and it optimizes things a lot!

lib/l1_consistency_checker/src/cache.rs
```diff
@@ -0,0 +1,175 @@
...
+        if *range.end() >= next_expected_block {
+            return Ok(None);
+        }
+
+        let mut blocks = Vec::with_capacity(range.clone().count());
+        for block_number in range.clone() {
+            let block = self
+                .data
+                .get(&block_number)
+                .expect("Cache should not be evicted while can still be used");
```
itegulov, line 107: This will make the node crash if the block is not available in cache for verification anymore, right? Is this scenario possible?

lib/l1_consistency_checker/src/cacher.rs
```diff
@@ -0,0 +1,96 @@
...
+        mut input: PeekableReceiver<Self::Input>,
+        _output: mpsc::Sender<Self::Output>,
+        state_reporter: ComponentStateReporter,
+    ) -> anyhow::Result<()> {
+        tracing::info!("starting local batch data cacher");
+        let mut cache_rx = self.cache.subscribe();
+        loop {
+            state_reporter.enter_state(GenericComponentState::Idle);
+            // Bound memory by waiting for verified batches to be evicted.
+            while !cache_rx.borrow_and_update().has_capacity() {
```
itegulov, line 82: Is there any chance this can lead to deadlock? AFAIU we rely on invariant that cache has enough capacity for entire batch's worth of blocks. I guess if `max_cached_bytes` is configured too low then there is a chance a batch is too big to fit into cache and hence will it never be evicted by checker?

itegulov (commented):
Looking good already!

lib/l1_consistency_checker/src/cache.rs
Artemka374: This is very unlikely to happen since cache is evicted only when verification is done. I can't imagine scenario when this can happen, but if it does - I believe the restart should help(since we'll cache these blocks once again)

lib/l1_consistency_checker/src/cacher.rs
Artemka374: so the batch limit of pubdata(which is the biggest part) is `110_000` bytes. With current default set to 512 MB it should be enough to store up to 4000 such batches, so this is very unlikely to happen

Artemka374: but I guess we can just set some kind of warning or assert of the configured value being too low to avoid such situations from happening at all

node/bin/src/lib.rs
```diff
@@ -410,10 +423,21 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
         // No batches committed - starting from genesis.
         0
     };
+    if !node_role.is_main() {
+        starting_block = starting_block.min(last_persisted_block_on_start.saturating_add(1));
```
itegulov, line 427: Why is this needed? I am trying to get rid of `starting_block` here (most usages are gone already)

node/bin/src/lib.rs
```diff
@@ -410,10 +423,21 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
+        if starting_block > 0 && starting_block < state.block_range_available().start() + 1 {
+            panic!(
+                "Cannot start: last_persisted_block_on_start + 1 < state.block_range_available().start() + 1: {} < {}",
+                last_persisted_block_on_start.saturating_add(1),
```
itegulov, line 431: You compare `starting_block` above which is not always `last_persisted_block_on_start.saturating_add(1)`

lib/l1_consistency_checker/src/cacher.rs
itegulov: Yeah I thought this was very unlikely. But please add an assert as this would be pretty difficult to debug if someone does run into this

node/bin/src/lib.rs
Artemka374: basically this would be first unverified batch - so we would want to start from it and actually verify it

itegulov: I see, shouldn't this logic be in `determine_starting_block` then? (basically another entry for `want_to_start_from` to choose from)

node/bin/src/lib.rs
Artemka374: I don't think I understand the concern here tbh

node/bin/src/lib.rs
Artemka374: Yeah makes sense

node/bin/src/lib.rs
itegulov: You compare `starting_block` above but print `last_persisted_block_on_start.saturating_add(1)` here

PR #1363, opened 2026-06-08
fix(backpressure): block ingestion under backpressure

antonbaliasnikov:
## Summary

Introduce `PipelineAdmissionGate` — a `watch::Sender<bool>` wrapper in `lib/backpressure` — driven by `BackpressureMonitor` and wired through both `ConsensusNodeCommandSource` and `ExternalNodeCommandSource`.

When the monitor detects pipeline lag (and flips the acceptance state to `NotAccepting`), the gate closes and command sources pause forwarding work into the pipeline. Once lag clears and the state returns to `Accepting`, the gate re-opens and ingestion resumes automatically.

Concretely:
- `BackpressureMonitor` sets the gate in `update_acceptance_state`, mirroring the RPC acceptance signal.
- `ConsensusNodeCommandSource` respects the gate during WAL replay, block rebuilds, and in the main produce loop (both the replay recv arm and the produce arm are disabled while the gate is closed; a `wait_until_open` wakeup arm re-enters the loop as soon as the gate opens).
- `ExternalNodeCommandSource` calls `wait_until_open()` before each `recv()` so EN replay also pauses under backpressure.

### DIFF
--- a/lib/backpressure/src/gate.rs
+++ b/lib/backpressure/src/gate.rs
@@ -1,0 +1,57 @@
+use tokio::sync::watch;
+
+/// Admission gate for internal pipeline sources.
+///
+/// This is deliberately separate from RPC transaction acceptance. `true` means
+/// the block pipeline can accept more work; `false` means command sources should
+/// stop forwarding new work until downstream lag clears.
+#[derive(Debug, Clone)]
+pub struct PipelineAdmissionGate {
+    tx: watch::Sender<bool>,
+}
+
+#[derive(Debug, Clone)]
+pub struct PipelineAdmissionReceiver {
+    rx: watch::Receiver<bool>,
+}
+
+impl PipelineAdmissionGate {
+    pub fn open() -> (Self, PipelineAdmissionReceiver) {
+        let (tx, rx) = watch::channel(true);
+        (Self { tx }, PipelineAdmissionReceiver { rx })
+    }
+
+    pub fn subscribe(&self) -> PipelineAdmissionReceiver {
+        PipelineAdmissionReceiver {
+            rx: self.tx.subscribe(),
+        }
+    }
+
+    pub fn set_open(&self) {
+        self.set(true);
+    }
+
+    pub fn set_closed(&self) {
+        self.set(false);
+    }
+
+    pub fn set(&self, open: bool) {
+        let _ = self.tx.send_if_modified(|current| {
+            if *current == open {
+                return false;
+            }
+            *current = open;
+            true
+        });
+    }
+}
+
+impl PipelineAdmissionReceiver {
+    pub fn is_open(&self) -> bool {
+        *self.rx.borrow()
+    }
+
+    pub async fn wait_until_open(&mut self) {
+        let _ = self.rx.wait_for(|open| *open).await;
+    }
+}
--- a/lib/backpressure/src/gate.rs
+++ b/lib/backpressure/src/gate.rs
@@ -15,24 +15,22 @@ pub struct PipelineAdmissionReceiver {
     rx: watch::Receiver<bool>,
 }
 
+impl Default for PipelineAdmissionGate {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl PipelineAdmissionGate {
-    pub fn open() -> (Self, PipelineAdmissionReceiver) {
-        let (tx, rx) = watch::channel(true);
-        (Self { tx }, PipelineAdmissionReceiver { rx })
+    pub fn new() -> Self {
+        let (tx, _) = watch::channel(true);
+        Self { tx }
     }
 
     pub fn subscribe(&self) -> PipelineAdmissionReceiver {
         PipelineAdmissionReceiver {
             rx: self.tx.subscribe(),
         }
-    }
-
-    pub fn set_open(&self) {
-        self.set(true);
-    }
-
-    pub fn set_closed(&self) {
-        self.set(false);
     }
 
     pub fn set(&self, open: bool) {
--- a/lib/backpressure/src/gate.rs
+++ b/lib/backpressure/src/gate.rs
@@ -2,10 +2,13 @@ use tokio::sync::watch;
 
 /// Admission gate for internal pipeline sources.
 ///
-/// This is deliberately separate from RPC transaction acceptance. `true` means
-/// the block pipeline can accept more work; `false` means command sources should
-/// stop forwarding new work until downstream lag clears.
-#[derive(Debug, Clone)]
+/// Driven by the same backpressure condition as RPC transaction acceptance
+/// (the monitor closes the gate exactly when it suspends acceptance), but
+/// delivered over a separate channel so command sources don't depend on the
+/// RPC-facing state type. `true` means the block pipeline can accept more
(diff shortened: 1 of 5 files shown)


node/bin/src/command_source.rs
```diff
@@ -172,8 +180,47 @@ impl<Replay: ReadReplay> ConsensusNodeCommandSource<Replay> {
...
+        for block_num in start..=end {
+            self.pipeline_gate.wait_until_open().await;
+            let record = self
+                .block_replay_storage
+                .get_replay_record(block_num)
+                .ok_or_else(|| anyhow::anyhow!("missing replay record for block {block_num}"))?;
+            output
+                .send(BlockCommand::Replay(Box::new(record)))
+                .await
+                .map_err(|_| anyhow::anyhow!("command output channel closed"))?;
```
slowli: - Nit: This can be rewritten via `context("...")`.
- Should an error be returned if the output channel is closed? AFAIU, previously used `forward_range_with()` doesn't error in this case. If it's desirable to single out the case when the output channel is closed, it could be possible by returning a value (e.g., a [`ControlFlow::Break`](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html)). Ditto in `forward_replay()` below; again, the previous implementation doesn't seem to error on closed output channel.

node/bin/src/command_source.rs
```diff
@@ -122,7 +122,29 @@ impl<Replay: ReadReplay> ConsensusNodeCommandSource<Replay> {
         tracing::info!(?role, "Consensus role initialized");
 
         loop {
+            let gate_open = self.pipeline_gate.is_open();
+            // Drain any already-queued canonized replays while the gate is open.
+            for _ in 0..Self::MAX_REPLAYS_TO_DRAIN_PER_LOOP {
+                if !gate_open {
```
slowli: Since `gate_open` isn't mutable, why is it checked on each loop iteration? Should it be updated on each `for` loop iteration?

lib/backpressure/src/gate.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    pub fn set_open(&self) {
+        self.set(true);
+    }
+
+    pub fn set_closed(&self) {
+        self.set(false);
+    }
```
slowli: These methods don't look to be used.

node/bin/src/command_source.rs
```diff
@@ -133,29 +155,15 @@ impl<Replay: ReadReplay> ConsensusNodeCommandSource<Replay> {
...
-                        .is_err()
-                    {
-                        tracing::info!("Command output channel closed, stopping source");
-                        break;
-                    }
-                    state_reporter.record_processed(block_number, Some(timestamp), None);
+                    Self::forward_replay(record, &output, &state_reporter).await?;
                 }
-                send_res = output.send(BlockCommand::Produce(ProduceCommand)), if role == ConsensusRole::Leader => {
+                _ = self.pipeline_gate.wait_until_open(), if !gate_open => {}
```
slowli, line 174: Dumb question: Is it OK not having the opposite clause (waiting for the `pipeline_gate` to close if it's currently open)? Assuming this event occurs before an existing clause in `tokio::select!` is triggered (say, `self.replays_to_execute.recv()`), is it OK that this clause may operate on an outdated value of `gate_open`?

lib/backpressure/src/monitor.rs
```diff
@@ -58,19 +59,26 @@ fn compute_adjacent_snapshots(
...
     config: BackpressureConfig,
     acceptance_tx: watch::Sender<TransactionAcceptanceState>,
+    pipeline_gate: PipelineAdmissionGate,
     stop_receiver: watch::Receiver<bool>,
 }
 
 impl BackpressureMonitor {
     pub fn new(config: BackpressureConfig, stop_receiver: watch::Receiver<bool>) -> Self {
         let (acceptance_tx, _) = watch::channel(TransactionAcceptanceState::Accepting);
+        let (pipeline_gate, _) = PipelineAdmissionGate::open();
```
slowli: Nit: Looks like this is the only use of the `open()` constructor, so it may make sense to return just a `PipelineAdmissionGate`.

node/bin/src/command_source.rs
antonbaliasnikov: Yeah, it's ok here. Since this is soft backpressure (reduces pipeline load, not a hard stop), a bounded 1-block overshoot after closure is acceptable. We will probably anyway, refactor this later, maybe even removing input sources completely. So, this is acceptable.

node/bin/src/command_source.rs
antonbaliasnikov: Fixed

node/bin/src/command_source.rs
antonbaliasnikov: Fixed

lib/backpressure/src/gate.rs
antonbaliasnikov: Fixed

lib/backpressure/src/monitor.rs
antonbaliasnikov: Fixed

lib/backpressure/src/gate.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+impl PipelineAdmissionReceiver {
+    pub fn is_open(&self) -> bool {
+        *self.rx.borrow()
+    }
+
+    pub async fn wait_until_open(&mut self) {
+        let _ = self.rx.wait_for(|open| *open).await;
```
slowli: This will also return if the corresponding `watch` channel is closed, which I believe could be unexpected by the callers (I'm not sure it's processed correctly by the *current* call sites either). If the current behavior is preserved, I'd suggest to add a doc comment and maybe a more verbose name (`wait_until_open_or_dropped`?). Alternatively, add `pending().await` (i.e., infinite waiting) if `wait_for` returns an error.

node/bin/src/command_source.rs
```diff
@@ -122,7 +123,31 @@ impl<Replay: ReadReplay> ConsensusNodeCommandSource<Replay> {
...
+                    }
+                    Err(mpsc::error::TryRecvError::Empty) => break,
+                    Err(mpsc::error::TryRecvError::Disconnected) => {
+                        tracing::info!("inbound channel closed");
+                        return Ok(());
+                    }
+                }
+            }
+
+            let can_produce = role == ConsensusRole::Leader && gate_open;
```
slowli, line 150: Just to be clear: Is it intentional that `gate_open` will be outdated here if `!self.pipeline_gate.is_open()` check breaks from the loop above? If it is, IMO it's worth explaining the reasoning in a comment.

antonbaliasnikov: Not intentional, fixed by moving the is_open() read to after the drain loop.

lib/backpressure/src/gate.rs
antonbaliasnikov: Good catch.

Went with your `pending().await `suggestion: on `Err` the future never resolves, so the gate freezes in its last state, consistent with `is_open()` (which also keeps reporting the last value after sender drop).

slowli (approved):

PR #1366, opened 2026-06-08
refactor(rpc): introduce Policy trait and Limiter

LogvinovLeon:
## Summary

Pure refactor. Sets up the structure for the tiered RPS limits — second `Policy` impl will land in a follow-up.

- New `rpc_limits` crate: `Policy` trait, config-driven `PerMethod` impl, and `Limiter` (lazily-populated DashMap of governor token buckets).
- `rate_limit_middleware` shrinks to the jsonrpsee glue — `Limiter::check(method)` then dispatch.
- The `*`-magic-string split moved out of `spawn` into `config::rpc_limit_policy`, next to `RpcRateLimit`.

Behavior unchanged.


lib/rpc_limits/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+
+mod limiter;
+
+pub use limiter::Limiter;
+
+use std::collections::HashMap;
+use std::num::NonZeroU32;
+
+/// Returns the global and per-method RPS limits. Different impls build them from different sources — a config list, tier numbers, etc.
+pub trait Policy: Send + Sync {
```
slowli: Bikeshedding: IMO, this name is too general; something like `RateLimitingPolicy` could be better.

lib/rpc_limits/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+    fn global(&self) -> Option<NonZeroU32>;
+    fn method(&self, method: &str) -> Option<NonZeroU32>;
```
slowli: Bikeshedding: IMO, it'd make more sense to *list* method limits:

```rust
fn methods(&self) -> HashMap<String, NonZeroU32>;
```

In this way, `Limiter` could be simplified to not construct per-method limiters lazily (and it could use an ordinary `HashMap`).

lib/rpc_limits/Cargo.toml
slowli, line 1: Bikeshedding: Not a fan of small crates with unclear reusability 🙃 Is it supposed to be used outside the `rpc` crate in a follow-up PR? Otherwise (e.g., if the follow-up PR would just extend the RPC configuration), I'd place it in the `rpc` or maybe `rpc_api`. This is not a strong preference, so feel free to disagree.

itegulov (commented):
Conceptually LGTM, but agree with @slowli's comments. Please address them and we can merge this

lib/rpc_limits/src/lib.rs
LogvinovLeon: My idea for naming is not to repeat the entities that already exist in the chain.
Example:
If I'm in the module: `rpc_limits` - I don't name the struct `RpcLimitsPolicy` Otherwise all structs within this module will have this name. Then later if someone uses it within the context where Policy is clear enough - they can do: `Policy`. But if additional clarity is required - they can do `rpc_limits::Policy`.

Or more general - if you have `crate/module/submodule/Struct` - if you name `Struct` - `ModuleStruct` or `SubmoduleModuleStruct` - you don't have the flexibility over the context you're dragging. But if you name if `Struct` - you can use `crate::module::submodule::Struct` is sibling crate or just `submodule::Struct` in a sibling submodule.

Although I understand that this might not be the convention we currently use and other arguments may prevail here. Please LMK if this speaks to you or if I should still change it?

lib/rpc/src/limits/limiter.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+/// Stateful enforcer for a [`Policy`]. Owns the token buckets; middleware calls `check` per request to gate it.
+pub struct Limiter {
+    global: Option<DefaultDirectRateLimiter>,
+    per_method: HashMap<String, DefaultDirectRateLimiter>,
+}
+
+impl Limiter {
+    pub fn new(policy: Arc<dyn Policy>) -> Arc<Self> {
```
slowli, line 28: Nit: This constructor doesn't require an `Arc` now; `&dyn Policy` would suffice.

lib/rpc_limits/src/lib.rs
slowli: I understand your reasoning, but IMO it would work the best if the item is fully intended to be prefixed by the module (like channel primitives in Tokio or std). Otherwise, the easier solution would be to *just* import it to the scope, which could make the use confusing. This is pure bikeshedding territory though; I'm fine with the current naming too, especially now that it's confined in the `rpc` crate.

slowli (approved):

lib/rpc/src/limits/limiter.rs
LogvinovLeon: Will be resolved in the next PR

LogvinovLeon:
Closing — splitting into two focused PRs directly off main: one for method_filter, one for rate_limits including the data-struct refactor. Avoids the intermediate Policy-trait state ever landing on main.

PR #1367, opened 2026-06-09
refactor: move pollers & log cache to NodeProvider

kostia244:
## Summary

- `LogsCache` is moved inside the provider and is used implicitly if enabled at provider creation.
- `BlockUpdates` removed in favor of two new separate pollers for block headers(finalized/latest) with lazy start.
- `LogsCache` re-uses latest block header from the poller, which mean it should almost never be be doing redundant calls.


integration-tests/src/lib.rs
```diff
@@ -746,6 +746,7 @@ impl Tester {
             l1.provider.clone()
         };
         let gateway_eth_provider = gateway_rpc_url.as_ref().map(|_| sl_provider.clone());
+        let l2_provider = NodeProvider::new(l2_provider).await;
```
itegulov: Why? I think it's already being wrapped below anyway

lib/provider/src/lib.rs
```diff
@@ -6,11 +6,16 @@
 //! [`NodeProvider::deployment_block`]), so the many startup binary searches over L1 history can use
 //! a tight lower bound without each rediscovering it.
 
+mod logs_cache;
+mod metrics;
+
+pub use logs_cache::LogsCache;
```
itegulov: Nit: please move re-exports under its respective `mod` for visibility

lib/provider/src/lib.rs
```diff
@@ -117,20 +130,167 @@ pub struct NodeProvider {
...
         let capabilities = ProviderCapabilities::detect(&provider).await;
         Self {
             inner: Box::new(provider),
             capabilities,
             deployment_blocks: Arc::new(Mutex::new(HashMap::new())),
+            latest_header_poller: Arc::new(OnceCell::new()),
+            finalized_header_poller: Arc::new(OnceCell::new()),
+            latest_poll_interval: Duration::from_secs(1),
+            finalized_poll_interval: Duration::from_secs(1),
+            log_cache: None,
```
itegulov: Nit: re-implement through a call to `new_with_features`

itegulov: Actually, is there a strong reason to have this at all? Integration tests?

lib/provider/src/lib.rs
```diff
@@ -117,20 +130,167 @@ pub struct NodeProvider {
...
+                log_cache_capacity,
+                chain_id,
+            ));
+        }
+
+        Ok(this)
+    }
+
+    /// Returns a shared poller for the latest block header via `eth_getBlockByNumber(latest, false)`.
+    pub async fn latest_header_poller(
```
itegulov: Mega-nit: `_poller` suffix here seems to refer to an implementation detail. This is maybe more like `latest_header_watcher`?

lib/provider/src/lib.rs
```diff
@@ -117,20 +130,167 @@ pub struct NodeProvider {
...
+                    .await
+            })
+            .await
+            .subscribe()
+    }
+
+    /// Returns a shared poller for the finalized block header via
+    /// `eth_getBlockByNumber(finalized, false)`.
+    /// Falls back to latetst if the chain does not support finalized tag.
+    pub async fn finalized_header_poller(
```
itegulov: Same here

lib/provider/src/lib.rs
```diff
@@ -117,20 +130,167 @@ pub struct NodeProvider {
...
+        tokio::spawn(async move {
+            let mut timer = tokio::time::interval(poll_interval);
+            loop {
+                timer.tick().await;
+                let Some(client) = weak_client.upgrade() else {
+                    return;
+                };
+
+                let block: Option<<Ethereum as Network>::BlockResponse> = client
+                    .request("eth_getBlockByNumber", (block, false))
```
itegulov, line 249: Hmm, why use `reqwest` here directly instead of the underlying provider?

lib/provider/src/lib.rs
```diff
@@ -103,11 +109,18 @@ impl ProviderCapabilities {
     }
 }
 
+impl Clone for Box<dyn EthWalletProvider + 'static> {
+    fn clone(&self) -> Self {
+        self.dyn_clone()
+    }
+}
```
itegulov: Nit: not against it but I don't think it is used(?)

node/bin/src/lib.rs
```diff
@@ -225,53 +241,10 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
-            provider.clone(),
-            gateway_block_updates
-                .clone()
-                .expect("gateway block updates must be initialized when gateway provider exists"),
-            config.l1_watcher_config.logs_cache_capacity,
-            l1_state.sl_chain_id,
-        )
-    });
-    let sl_logs_cache = if l1_state.l1_chain_id == l1_state.sl_chain_id {
-        l1_logs_cache.clone()
```
itegulov, line 270: Nice to see all this code gone from `bin/src/lib.rs`

lib/provider/src/logs_cache.rs
```diff
@@ -171,26 +162,25 @@ impl RecentLogs {
 /// And remembers them for last `watcher_config.capacity` blocks.
 ///
 /// TODO: As of now there is no filtering for these logs. Although with current settings memory usage shouldn't be a problem.
-/// TODO: In reorg checks we do additional eth_getBlockByNumber - this can be avoided by extending BlockUpdates.
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct LogsCache {
```
itegulov: I believe this, along with other stuff from this module can become `pub(crate)` now

lib/provider/src/lib.rs
itegulov: I think I see now. It's used in `LogsCache`, but I'll leave a separate comment on that

lib/provider/src/logs_cache.rs
```diff
@@ -171,26 +162,25 @@ impl RecentLogs {
...
-    provider: NodeProvider,
-    block_updates: watch::Receiver<BlockUpdates>,
+    provider: Box<dyn EthWalletProvider + 'static>,
```
itegulov: I think holding `provider` here is a bit of a smell. As an idea, can we pass `inner` from `NodeProvider` as `&impl Provider` into where it is currently used (i.e. `get_logs` and its call tree)?

itegulov: Maybe even `&RootProvider` so you can't accidentally pass self-referential `NodeProvider` or something like that

itegulov (commented):
LGTM overall, just a bunch of nits

lib/provider/src/lib.rs
kostia244: Yes integration tests, they are annoying here because the use some of NodeProvider features & our utilities, but dont' need server specific things.

lib/provider/src/lib.rs
kostia244: Originally I wanted to use alloy's poller but it didn't work out because it returns a broadcast. So I just stole their implementation flow. They use weak client and it is also used to figure out when the provider dies.
There isn't really a reason other than steal alloy's logic here, I could change it if you don't like it.

kostia244: I'd keep it like this because we don't have to deal with provider ownership or cloning

lib/provider/src/logs_cache.rs
kostia244: Did as you suggested, looks clean enough IMO

itegulov (approved):

PR #1371, opened 2026-06-10
refactor(l1-watcher): decouple initialization from constructor

itegulov:
## Summary

Refactor L1 watchers so the starting point is supplied at `run()` time instead of at construction. `create_watcher` now only takes static dependencies; the cursor / protocol version / finality-derived start is passed to `run(start)`.

This lets watcher components be created in one place and started in another - e.g. later, once the first replayed block (and thus the starting cursors) is known.

### DIFF
--- a/lib/l1_watcher/src/commit_watcher.rs
+++ b/lib/l1_watcher/src/commit_watcher.rs
@@ -1,6 +1,6 @@
 use crate::committed_batch_provider::CommittedBatchProvider;
-use crate::watcher::{L1Watcher, L1WatcherError};
-use crate::{L1WatcherConfig, ProcessL1Event, util};
+use crate::watcher::{L1Watcher, L1WatcherError, resolver};
+use crate::{L1WatcherConfig, ProcessL1Event, ProcessRawEvents, util};
 use alloy::rpc::types::Log;
 use tokio::sync::watch;
 use zksync_os_batch_types::DiscoveredCommittedBatch;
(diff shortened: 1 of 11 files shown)


kostia244:
Question: What is the benefit of adding the Resolver, instead of, for example, wrapping create_watcher part in something like
```
fn create_watcher_task<F, W>(create_watcher: F) -> impl Future<Output = ()> + Send
where
    F: Future<Output = Result<W>> + Send + 'static,
    W: WatcherRunner,
{
    async move {
        create_watcher
            .await
            .expect("failed to create L1 watcher")
            .run()
            .await;
    }
}
```

Then you can just use create reth task out of this helper, it seems like it would work for sl aware watcher too.

kostia244 (commented):
I thought about this, the benefit of being able to pass the arguments later is clear. But still I find the flow a bit strange.
Processor::run -> (L1Watcher + Resolver) -> (RunningL1Watcher + Processor)

Perhaps we could have something like 
Processor::run -> resolver -> (L1Watcher + Processor)

Or in other words return a resolver and make the resolver produce a watcher?

itegulov:
@kostia244 this refactoring is not just because I want to make initialization non-blocking but also because I am trying to move `first_replay_record` logic from `bin/src/lib.rs` to BlockContextProvider (I already did a part of this refactoring). So essentially watchers are still created by the main logic but then provided a starting point later down the flow once first replay record is known.

I am still experimenting whether it makes sense for BCP to own L1TxWatcher, InteropWatcher and GatewayMigrationWatcher or, counterintuitively, if they should actually be owned be mempool (feels like they should be owned by mempool but I am not 100% convinced yet).

kostia244 (approved):

Issue #1378, opened 2026-06-11
[Feature request] EN-to-EN block relay over the `zks` subprotocol


fadeevab:
## Problem

* `v0.13` had a block replay via HTTP. It was useful for custom topologies, e.g., MN could be hidden behind the set of external nodes (EN).
* `v0.17` introduced the `devp2p` while removing the block replay between external nodes. This forces every EN in the fleet to maintain a direct connection to an MN, blocking any fan-out tree.

## Feature request

Make the topology `MN → EN → EN` (and deeper) possible over the `zks` devp2p subprotocol, so an EN can serve block replays to another EN without every node needing a direct connection to the MN.

## Use cases

* Hide the MN behind a set of ENs for security or network isolation.
* Reduce direct load on the MN in large deployments by fanning out through intermediate ENs.
* Make the 3rd-party external nodes topology more agile (external EN-EN networks).

fadeevab:
Contribution PR: https://github.com/matter-labs/zksync-os-server/pull/1379

PR #1379, opened 2026-06-11
feat: MN→EN↔EN block replay over the `zks` subprotocol

fadeevab:
## Summary

Enables the topology `MN → EN → EN` (and deeper) for `zks` block replay.
Solves the issue #1378.

Wire format is untouched; no versioned files under `wire/replays/v*.rs` changed.

## Use cases
* Hide the MN behind a set of ENs for security or network isolation.
* Reduce direct load on the MN in large deployments by fanning out through intermediate ENs.
* Make the 3rd-party external nodes topology more agile (external EN-EN networks).

## Safety 
* Only boot nodes may be upstream.
* MN cannot be poisoned over zks (it's a server-only).
* An EN consumes from exactly one upstream.
* Serving is read-only, thus it's harmless.
* No peer-claimed role is trusted.

## Changes

The changes combine **a few features and edge-case handling**; each is crucial for **_safety_** and **_liveness_**.

**Core change (`en.rs`):** replace the one-sided consume loop with a per-connection sync-state machine. Each EN link probes the peer (`GetBlockReplays`), pins the first peer that delivers a *cursor-aligned* block (first record == local cursor) as the single upstream via a capacity-1 `UpstreamGuard` semaphore, consumes that stream, and serves every other link from its own WAL via `stream_from_forever`. Direction is ignored - an MN that dials an EN works the same as an EN that dials an MN.

**Re-parenting:** when the pinned link drops, the guard is released and the next cursor-aligned stream wins it.

**Trusted peers:** `trusted_peers` populated from `network_boot_nodes`. Only those peer IDs can become the pinned upstream; untrusted peers can connect and receive blocks but cannot pin. Empty set = accept any (tests).

**Readiness middleware:** The RPC server now binds immediately on startup, so a downstream EN can call `load_remote_config` before this node has fully booted. State-dependent methods return `-32000 Node is not ready` until
the DB readiness signal fires; a small set of static config methods (`zks_getGenesis`, `eth_chainId`, etc.) are always available. 

* _The issue it fixes:_ moving RPC start before DB readiness exposed DB-dependent methods to early calls; the middleware gates them while letting static config methods through.

**Tests added:**
- 5 network-layer e2e tests: transitive   `MN→EN1→EN2`, MN-dials-EN, single-upstream (no panic), verifier-on-EN-server,  untrusted-peer-rejected.
- 2 integration tests: 
  * `transaction_replay_en_chain` (strict EN chain topology end-to-end),
  * `en_chain_survives_en1_restart` (EN1 stops, block produced, EN1 restarts with same peer ID, EN2 reconnects and catches up)

### DIFF
--- a/lib/mempool/src/pool.rs
+++ b/lib/mempool/src/pool.rs
@@ -283,7 +283,6 @@ impl<T: L2Subpool> Pool<T> {
         let mut interop_txs = Vec::new();
         let mut interop_fee_txs = Vec::new();
         let mut l1_transactions = Vec::new();
-        let mut l2_transactions = Vec::new();
         for tx in &replay_record.transactions {
             match tx.envelope() {
                 ZkEnvelope::System(system_tx) => match system_tx.system_subtype() {
@@ -301,12 +300,10 @@ impl<T: L2Subpool> Pool<T> {
                 ZkEnvelope::L1(l1_tx) => {
                     l1_transactions.push(l1_tx);
                 }
-                ZkEnvelope::L2(l2_tx) => {
-                    l2_transactions.push(*l2_tx.hash());
-                }
                 ZkEnvelope::Upgrade(upgrade) => {
                     upgrade_txs.push(upgrade);
                 }
+                ZkEnvelope::L2(_) => {}
             }
         }
         self.upgrade_subpool
@@ -325,6 +322,31 @@ impl<T: L2Subpool> Pool<T> {
             .on_canonical_state_change(l1_transactions)
             .await;
 
+        self.notify_l2_canonical_state_change(header, account_diffs, replay_record);
+
+        StateChangeOutcome {
+            last_interop_log_id,
+            last_l1_priority_id,
+            last_interop_fee_number,
+        }
+    }
+
+    /// Notifies the L2 subpool of a newly committed block: evicts mined transactions
+    /// and updates account nonces/balances. Does not touch L1-fed subpools.
+    pub fn notify_l2_canonical_state_change(
+        &self,
+        header: Sealed<Header>,
+        account_diffs: &[AccountDiff],
+        replay_record: &ReplayRecord,
+    ) {
+        let l2_transactions: Vec<TxHash> = replay_record
+            .transactions
+            .iter()
+            .filter_map(|tx| match tx.envelope() {
+                ZkEnvelope::L2(l2_tx) => Some(*l2_tx.hash()),
+                _ => None,
+            })
+            .collect();
         let (header, hash) = header.into_parts();
         let body = BlockBody::default();
         let block = Block::new(header, body);
@@ -347,12 +369,6 @@ impl<T: L2Subpool> Pool<T> {
                 mined_transactions: l2_transactions,
                 update_kind: PoolUpdateKind::Commit,
             });
-
-        StateChangeOutcome {
-            last_interop_log_id,
-            last_l1_priority_id,
-            last_interop_fee_number,
-        }
     }
 }
 
--- a/lib/mempool/src/pool.rs
+++ b/lib/mempool/src/pool.rs
@@ -283,6 +283,7 @@ impl<T: L2Subpool> Pool<T> {
         let mut interop_txs = Vec::new();
         let mut interop_fee_txs = Vec::new();
         let mut l1_transactions = Vec::new();
+        let mut l2_transactions = Vec::new();
         for tx in &replay_record.transactions {
             match tx.envelope() {
                 ZkEnvelope::System(system_tx) => match system_tx.system_subtype() {
@@ -300,10 +301,12 @@ impl<T: L2Subpool> Pool<T> {
                 ZkEnvelope::L1(l1_tx) => {
                     l1_transactions.push(l1_tx);
                 }
+                ZkEnvelope::L2(l2_tx) => {
+                    l2_transactions.push(*l2_tx.hash());
+                }
                 ZkEnvelope::Upgrade(upgrade) => {
                     upgrade_txs.push(upgrade);
                 }
-                ZkEnvelope::L2(_) => {}
             }
         }
         self.upgrade_subpool
@@ -322,31 +325,6 @@ impl<T: L2Subpool> Pool<T> {
             .on_canonical_state_change(l1_transactions)
             .await;
 
-        self.notify_l2_canonical_state_change(header, account_diffs, replay_record);
-
-        StateChangeOutcome {
-            last_interop_log_id,
-            last_l1_priority_id,
-            last_interop_fee_number,
-        }
-    }
-
-    /// Notifies the L2 subpool of a newly committed block: evicts mined transactions
-    /// and updates account nonces/balances. Does not touch L1-fed subpools.
-    pub fn notify_l2_canonical_state_change(
-        &self,
-        header: Sealed<Header>,
-        account_diffs: &[AccountDiff],
-        replay_record: &ReplayRecord,
-    ) {
-        let l2_transactions: Vec<TxHash> = replay_record
-            .transactions
-            .iter()
-            .filter_map(|tx| match tx.envelope() {
-                ZkEnvelope::L2(l2_tx) => Some(*l2_tx.hash()),
-                _ => None,
-            })
-            .collect();
         let (header, hash) = header.into_parts();
         let body = BlockBody::default();
         let block = Block::new(header, body);
@@ -369,6 +347,12 @@ impl<T: L2Subpool> Pool<T> {
                 mined_transactions: l2_transactions,
                 update_kind: PoolUpdateKind::Commit,
             });
+
+        StateChangeOutcome {
+            last_interop_log_id,
+            last_l1_priority_id,
+            last_interop_fee_number,
+        }
     }
 }
 
--- a/lib/rpc/src/lib.rs
+++ b/lib/rpc/src/lib.rs
@@ -155,6 +155,7 @@ pub async fn spawn<RpcStorage: ReadRpcStorage, Mempool: L2Subpool>(
     let max_response_size_bytes = config.max_response_size_bytes();
     let limiter = LoggingLimiter::new(Limiter::new(config.rate_limits.clone().into_limits()));
     let rate_limit_logging = LoggingLimiter::run(limiter.clone());
+    let (ready_tx, ready_rx) = watch::channel(false);
     let method_filter = Arc::new(config.method_filter.clone());
     // Snapshot the registered method names so monitoring can bound metric label cardinality:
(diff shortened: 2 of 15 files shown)


fadeevab:
@itegulov Hi! If you have a chance, could you please take a look at this PR? I added **Use Cases** and **Safety notes**.

These changes are meant to be reviewed together because none of them can be shipped independently without compromising safety or liveness. For example, boot_nodes (trusted nodes) are needed to ensure safety, and the RPC change (I moved RPC _before_ DB is ready) allows static methods to be accessible to fellow ENs before it's fully ready (thus, ReadinessMiddleware appears to gate DB-dependent methods).

itegulov:
@fadeevab apologies for the slow answer, I am currently preoccupied with other things so haven't had a chance to do a full review yet. From a brief look the PR makes sense to me (seems to contain _seemingly_ unrelated changes as well but I'll comment on this during the review). Only question I have so far: what happens if two ENs set each other as boot nodes, then a loop is possible right? Maybe moving the responsibility onto infra config is an acceptable tradeoff for now but I just wanted to understand the intended usage a bit better.

fadeevab:
@itegulov Thank you for your reply. Appreciate your time!

I've double checked, and I would sum it up this way:

1. "Upstream pinning" starts from a successful "receiving".
2. Successful receiving means that `new block height` = `local cursor` > `local block height`.
3. So generally, nodes `A != B` in terms of block heights.

So, on the code side, there are few main guards:
  * `BlockReplays(response)` has a condition, e.g., serving from `B` to -> `A` iff `B height >= A cursor > A height`:
 
     ```rust
      // Skip, if response is not aligned with the local cursor.
      if first_block != *starting_block.read().unwrap() {
          continue;
      }
     ```
  * `GetBlockReplays` has a guard to not serve its own upstream:
  
      ```
      if upstream.peer_id() != Some(peer_id){...}
      ```
   * Only one upstream can exist:
   
      ```
      match upstream.try_acquire(peer_id) {
      ```
      
(I guess I should make those three code points clearer in the comments.)
   
### Edge cases
1. If A pins B first, B can't get blocks from A.
2. If B pins A first, A can't get blocks from B.
3. To pin each other simultaneously, there should exist a race where they both **receive** a successful `BlockReplays(response)` from each other, where A>B and B>A, and it is a contradiction.
4. It might be possible if they both had their own upstream sources. But they both have a guard that only one upstream can exist: if they both have their upstreams (C && D), they can't pin a new upstream.
5. If only A or B has its own upstream source (C || D), then one can serve, and another can pin, so a loop is impossible as well.

lib/mempool/src/pool.rs
```diff
@@ -248,6 +245,33 @@ impl<T: L2Subpool> Pool<T> {
...
+            last_l1_priority_id,
+            last_migration_number: sl_chain_id_outcome.map(|o| o.last_migration_number),
+            last_sl_chain_id_target: sl_chain_id_outcome.map(|o| o.last_sl_chain_id_target),
+            last_interop_fee_number,
+        }
+    }
+
+    /// Notifies the L2 subpool of a newly committed block: evicts mined transactions
+    /// and updates account nonces/balances. Does not touch L1-fed subpools.
+    pub fn notify_l2_canonical_state_change(
```
itegulov: This looks weird, you shouldn't need to touch this logic

lib/sequencer/src/execution/block_context_provider.rs
```diff
@@ -441,45 +444,56 @@ impl<Subpool: L2Subpool> BlockContextProvider<Subpool> {
         strict_subpool_cleanup: bool,
     ) {
         let mut next_cursors = replay_record.starting_cursors.clone();
-        let outcome = self
-            .pool
-            .on_canonical_state_change(
+        // Skip L1-fed subpools on EN: they are empty (no watcher populates them).
+        // The cursors come from the upstream (MN or EN) in `replay_record.starting_cursors`.
+        if !self.is_external_node {
```
itegulov: Ok, I see where you use that `notify_l2_canonical_state_change` method now. The AI comment here is incorrect, the L1-fed subpools are not empty on ENs and are expected to be populated from L1. I _think_ what's happening here is that `starting_block` in `node/bin/src/lib.rs` is being decided against main node RPC but then replay starts from some EN which results in mempools being pre-populated from a wrong block.

I am currently working on a refactoring that removes that whole logic so that starting block will be decided during p2p handshake which should eventually unblock this.

node/bin/src/lib.rs
```diff
@@ -186,31 +187,26 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
...
                 .expect("Missing `main_node_rpc_url` in external node config");
-            load_remote_config(&main_node_rpc_url, &config.genesis_config)
+            // Exponential backoff: this is a cold-start probe that runs once.
+            // If MN isn't ready yet, backing off to 30s without flooding during its own startup.
+            (|| load_remote_config(&main_node_rpc_url, &config.genesis_config))
+                .retry(
+                    ExponentialBuilder::default()
+                        .with_min_delay(Duration::from_secs(1))
+                        .with_max_delay(Duration::from_secs(30))
+                        .with_max_times(0),
```
itegulov: Doesn't `with_max_times(0)` mean this will never get retried? Also, not sure why this is a part of this PR at all. Is it because `main_node_rpc_url` can point to an EN now that is not ready yet?

itegulov: Also, I think this logic clashes with what @LogvinovLeon is doing in https://github.com/matter-labs/zksync-os-server/pull/1405. I'd remove it from this PR

itegulov (commented):
I think this might be difficult to merge until we get rid of `starting_block` thing I mentioned in one of the comments. There is a chance I am wrong though, please try this PR without the mempool/block-context-provider changes and lmk if that works.

node/bin/src/lib.rs
fadeevab: Yes, my patch is basically a simple few-lines variation of what @LogvinovLeon is doing in the PR #1405 here: [`const RETRY: ExponentialBuilder = ExponentialBuilder::new()`](https://github.com/matter-labs/zksync-os-server/pull/1405/changes#diff-f6d64263010d75c69764c795410dc9566d6f41d6a9b1c13cc728406505df3cb2R14).

After PR #1405 is merged, this part won't be needed.

lib/mempool/src/pool.rs
fadeevab: I have a deadlock if EN notifies the L1 subpools, so I had to make sure it notifies only the L2 subpools

node/bin/src/lib.rs
fadeevab: True, the correct way would be `.without_max_times()`. 
Either way, after PR #1405 is merged, this part won't be needed.

I had end-to-end testing issues when EN couldn't load the remote config, but it seems like, I was eventually lucky, so that EN loaded the remote config without any retries.

lib/sequencer/src/execution/block_context_provider.rs
fadeevab: I see. What happened on my side: while e2e testing on my local setup, I kept hitting a deadlock in `on_canonical_state_change`, so skipping the L1-fed subpools on the EN was a way to get past it (while still handling the L2 subpool).

I mean, it's gonna be hard to test this PR without either this current change, or without your `starting_block` refactoring, because I was hitting the deadlock relentlessly.

fadeevab (commented):
Thank you for the review!

I guess there are at least 2 options for moving forward:

1. I can wait for both your PRs merged (refactoring of the `starting_block` and PR #1405) and then refresh my PR accordingly.
2. Or, I can remove the retries and fix the comment about the "L1-fed subpools" keeping the current implementation as-is, for faster merge.

WDYT?

P.S.: As I noted below, it will be hard to test without the subpools workaround because of the deadlock I'm constantly hitting during the end-to-end testing.

lib/sequencer/src/execution/block_context_provider.rs
itegulov: That's what I expected yeah... 99% chance this is the issue I was talking about in the comment above

itegulov:
Ok, I propose we wait until the refactoring is complete (FWIW #1405 is already merged). It will take some time, I'll need https://github.com/matter-labs/zksync-os-server/pull/1409 first which is only going to be reviewed next week. But AFAIU you guys are not in a rush?

fadeevab:
> until the refactoring is complete

Great 👍

> AFAIU you guys are not in a rush?

Yes, no rush, take your time

lib/mempool/src/pool.rs
fadeevab: After rebasing, this change is gone

lib/sequencer/src/execution/block_context_provider.rs
fadeevab: After rebasing, this change is gone as well

node/bin/src/lib.rs
fadeevab: Removed

fadeevab:
@itegulov I rebased it to the fresh `main` branch. Exponential backoff and the L2 subpool "hack" are both gone now, and the quick end-to-end test works fine.

lib/storage/src/lazy/repository.rs
```diff
@@ -43,13 +43,18 @@ impl RepositoryManager {
...
+        let db_has_existing_blocks = db.get_latest_block() > 0;
+
         RepositoryManager {
             // Initializes in-memory repository with genesis block. It is never pruned from cache.
             in_memory: RepositoryInMemory::new(genesis_block),
             db,
             max_blocks_in_memory: blocks_to_retain as u64,
             block_sender,
-            db_ready_to_process_blocks: Arc::new(AtomicBool::new(false)),
+            db_ready_to_process_blocks: Arc::new(AtomicBool::new(db_has_existing_blocks)),
```
itegulov, line 57: AFAIU the changes in this file can be reverted, no?

lib/rpc/src/lib.rs
```diff
@@ -162,6 +165,8 @@ pub async fn spawn<RpcStorage: ReadRpcStorage, Mempool: L2Subpool>(
         .layer_fn(move |service| {
             Monitoring::new(service, max_response_size_bytes, known_methods.clone())
         })
+        // Make static config methods available while gating DB-dependent calls until they are ready.
+        .layer_fn(move |service| Readiness::new(service, ready_rx.clone()))
```
itegulov, line 169: My gut feeling is that this isn't necessary as well, but I'll experiment and double check locally first

itegulov (commented):
Looks much better now! I tried it locally and I believe we need to rework trusted peer whitelisting that was done in #1438
The integration tests in the PR did not work for me without patching this logic locally

PR #1389, opened 2026-06-15
fix: ignore tx limit for v31 upgrade block

perekopskiy:
## Summary

v31 upgrade block should contain exactly 2 transactions, ignore tx_limit=1 if set

### DIFF
--- a/lib/sequencer/src/execution/execute_block_in_vm.rs
+++ b/lib/sequencer/src/execution/execute_block_in_vm.rs
@@ -71,6 +71,17 @@ pub async fn execute_block_in_vm<V: ViewState>(
     let mut interop_roots_count = 0;
     let expect_sl_chain_id_tx_after_upgrade = command.expect_sl_chain_id_tx_after_upgrade;
 
+    if expect_sl_chain_id_tx_after_upgrade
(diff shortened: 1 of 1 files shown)


lib/sequencer/src/execution/execute_block_in_vm.rs
```diff
@@ -71,6 +71,17 @@ pub async fn execute_block_in_vm<V: ViewState>(
     let mut interop_roots_count = 0;
     let expect_sl_chain_id_tx_after_upgrade = command.expect_sl_chain_id_tx_after_upgrade;
 
+    if expect_sl_chain_id_tx_after_upgrade
```
EmilLuta, line 74: Does it make sense for protocol upgrade to make it more flexible? (rather than just for v31)

Maybe it would be nice to relax this to bigger amount of txs in the future (but also, maybe there'll be no use case, I don't have full context of what other upgrades could look like).

perekopskiy: > I don't have full context of what other upgrades could look like

Well, I guess no one really knows, but normal protocol upgrade requires only 1 tx in block

antonbaliasnikov (approved):

PR #1391, opened 2026-06-16
refactor(mempool): directly own watchers needed for subpools

itegulov:
## Summary

This PR moves all mempool-related watchers directly into `Pool` thus simplifying `bin/src/lib.rs`. Watchers are started lazily once mempool is initialized with the very first replay record we are replaying. This is one of the last remaining things we need to get rid of so that we can drop `starting_block` logic in `bin/src/lib.rs`.


lib/mempool/src/subpools/interop_fee.rs
```diff
@@ -16,33 +16,48 @@ pub struct InteropFeeSubpool {
...
+            "tried to re-initialize InteropFeeSubpool"
+        );
+        inner.next_interop_fee_number = Some(next_interop_fee_number);
+    }
 
     pub async fn best_transactions_stream(&self) -> InteropFeeTransactionsStream {
         let (sender, receiver) = mpsc::channel(1);
         let mut inner = self.inner.write().await;
-        let next_interop_fee_number = inner.next_interop_fee_number;
+        let Some(next_interop_fee_number) = inner.next_interop_fee_number else {
```
perekopskiy: Isn't it unexpected that InteropFeeSubpool was not yet initialized? Perhaps, it's better to panic instead? Same for `consume_expected_tx`, `on_canonical_state_change`

lib/mempool/src/pool.rs
```diff
@@ -6,46 +6,190 @@ use crate::subpools::l2::{L2Subpool, L2TransactionsStreamMarker};
...
+    /// Initializes mempool with the starting block, expects to be called exactly once during the
+    /// node's lifetime.
+    pub async fn init(&mut self, replay: &ReplayRecord) {
+        let current_protocol_version = &replay.protocol_version;
+        self.upgrade_subpool
+            .init(current_protocol_version.clone())
+            .await;
+
+        // If we start from genesis, we should start by sending upgrade tx for genesis.
+        if replay.block_context.block_number == 0 {
```
perekopskiy: Coming back to this discussion https://github.com/matter-labs/zksync-os-server/pull/1357#discussion_r3359659039. You said "Returning to your original question starting_block=1 would mean that block 1 was already executed in the past and we already have a replay record for it. ". I agree, but still we won't have genesis upgrade tx in mempool, won't we?

lib/mempool/src/subpools/interop_fee.rs
itegulov: Indeed, I simplified the logic a bit. Should be better now

lib/mempool/src/pool.rs
itegulov: Ah true, somehow I forgot that we need this upgrade tx during replay... Good point, fixed

perekopskiy (approved):

PR #1392, opened 2026-06-16
feat: wait for execution delay

PatrickAlphaC:
The `l1_sender_execute` pipeline treats a not-yet-executable batch as a fatal error and crashes the whole server. This makes it wait for the `ValidatorTimelock` execution-delay window to open and then send, instead of panicking.

## Problem

When a batch is committed and proven but still inside the `ValidatorTimelock`'s `executionDelay` window, `executeBatchesSharedBridge` reverts with `TimeNotReached(uint256 executableFrom, uint256 currentTimestamp)` (selector `0x08753982`).

In the current code that revert is only seen indirectly, during gas estimation:

1. `estimate_gas_limits` runs `eth_simulateV1`; the execute call reverts, so it falls back to the hardcoded `L1_GAS_LIMIT_FALLBACK` (15,000,000).
2. The send path then builds the transaction with that 15M limit. The L1 node's pre-flight balance check (`balance >= gas_limit * max_fee_per_gas + value`) fails — `15,000,000 * max_fee_per_gas` is a large phantom requirement that has nothing to do with the real execute cost — and `send_transaction` returns `-32003 insufficient funds`.
3. That error propagates out of the `l1_sender_execute` critical task, which the supervisor treats as unrecoverable and shuts the entire process down (sequencer, commit, prove, RPC). systemd restarts it, it immediately re-attempts the same not-yet-executable batch, and crash-loops until the window happens to have elapsed.

The misleading part is the insufficient funds message — it's a symptom of "batch not executable yet," not an actual balance shortfall. No amount of funding or fee tuning fixes it; lowering `max_fee_per_gas` only lets the doomed transaction get broadcast and revert on-chain (burning gas), which then fails in the receipt-validation path instead.

## When it surfaces

Any chain with a non-zero `executionDelay` hits this once the prove/execute pipeline runs close to real time. A backlog hides it: backlogged batches are always well past the delay by the time they reach the execute stage, so the timelock is never the binding constraint. As soon as the pipeline catches up, each fresh batch reaches execute within minutes of commit — inside the delay window — and trips the crash. There is currently no awareness of the timelock delay anywhere in the sender.

## Fix

Detect the `TimeNotReached` revert from the `eth_simulateV1` call that already runs for gas estimation, and wait instead of sending a doomed transaction:

- `estimate_gas_limits` now returns `GasEstimation::Ready(gas_limits)` or `GasEstimation::NotExecutableUntil(executableFrom)`. The `executableFrom` timestamp is decoded directly from the revert payload (`decode_time_not_reached`).
- The send path loops: on `NotExecutableUntil`, it sleeps until the window opens — measured against the L1 pending-block timestamp (the same clock the timelock compares against), plus a small buffer, capped per-iteration so it periodically re-simulates and remains responsive to shutdown / clock skew — then re-simulates and only sends once `Ready`.
- A new `WaitingExecutionDelay` component state makes the wait observable.

## Scope / safety

- Only the execute command can hit this; commit and prove always resolve to `Ready` on the first pass, so their behavior is unchanged.
- Every other simulate revert keeps the existing 15M-fallback behavior — only `TimeNotReached` is special-cased.
- No transaction is broadcast while waiting, so no nonce is consumed and no gas is spent.
- Blocking the execute component is correct: batch N+1 can't be executed before batch N anyway, and commit/prove/sequencer run as independent tasks that keep making progress. This mirrors how zksync-era tolerates its execution delay.

### DIFF
--- a/lib/l1_sender/src/lib.rs
+++ b/lib/l1_sender/src/lib.rs
@@ -41,6 +41,9 @@ pub enum L1SenderState {
     SendingToL1,
     /// Transaction submitted; waiting for L1 block inclusion.
     WaitingL1Inclusion,
+    /// The batch cannot be executed yet because the ValidatorTimelock execution delay
+    /// has not elapsed; waiting for the window to open before sending.
+    WaitingExecutionDelay,
 }
 
 impl StateLabel for L1SenderState {
@@ -49,6 +52,7 @@ impl StateLabel for L1SenderState {
             Self::Idle => GenericComponentState::Idle,
             Self::SendingToL1 => GenericComponentState::Active,
             Self::WaitingL1Inclusion => GenericComponentState::Active,
+            Self::WaitingExecutionDelay => GenericComponentState::Active,
         }
     }
     fn specific(&self) -> &'static str {
@@ -56,6 +60,7 @@ impl StateLabel for L1SenderState {
             Self::Idle => "idle",
             Self::SendingToL1 => "sending_to_l1",
             Self::WaitingL1Inclusion => "waiting_l1_inclusion",
+            Self::WaitingExecutionDelay => "waiting_execution_delay",
         }
     }
 }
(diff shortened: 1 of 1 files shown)


PatrickAlphaC:
execution delay should be zero

PR #1393, opened 2026-06-17
refactor(batcher): split pending/committed batch info

itegulov:
## Summary

  Splits the dual-purpose `ExtendedCommitBatchInfo` into two purpose-built types and lets the L1 watcher read a batch's `commitment` directly from the `BlockCommit` event instead of deriving it from unreliable L1 state calls.                                                                          
                                                                                                                                                                                                                                                                                                           
  - **`PendingBatchInfo`** (renamed from `ExtendedCommitBatchInfo`) — the batcher's working type, driven through the pipeline. Keeps `commit_info` + `upgrade_tx_hash` + `protocol_version`; computes the commitment lazily.                                                                               
  - **`CommittedBatchInfo`** (new, slim) — `commit_info` + `commitment`, used for batches discovered already-committed on L1.

Main reasoning is removing old assumptions that can break in extreme circumstances. Which can lead to very hard-to-debug issues down the line

### DIFF
--- a/lib/l1_watcher/src/util.rs
+++ b/lib/l1_watcher/src/util.rs
@@ -384,7 +384,7 @@ pub async fn fetch_stored_batch_data(
     }) else {
         return Ok(None);
     };
-    let batch_info = fetch_committed_batch_data(zk_chain, tx_hash, l1_block_number)
+    let batch_info = fetch_committed_batch_data(zk_chain, tx_hash, l1_block_number, batch_number)
         .await?
         .into_stored();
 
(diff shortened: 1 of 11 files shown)


lib/l1_watcher/src/util.rs
```diff
@@ -403,79 +400,67 @@ pub async fn fetch_committed_batch_data(
...
-        // there is none. If none, `upgrade_batch_number` would be `0` and thus never equal
-        // to the currently inspected batch as genesis does not get committed via this flow.
-        None
-    };
-    Ok(ExtendedCommitBatchInfo {
+    // The batch commitment is emitted in the `BlockCommit` event (indexed by batch number) of
+    // the commit transaction. Reading it from L1 directly is safe and accurate, unlike deriving
+    // it from the current protocol version / upgrade transaction hash, which reflect the latest
+    // chain state rather than the state at the moment this batch was committed.
+    let logs = zk_chain
```
antonbaliasnikov: Why do we call `get_transaction_by_hash` with retries, but `get_logs` without?

Does it make sense to also use `tokio::try_join!` for these two calls to win a bit of time?

lib/l1_watcher/src/util.rs
```diff
@@ -403,79 +400,67 @@ pub async fn fetch_committed_batch_data(
...
+                .to_block(l1_block_number),
+        )
+        .await
+        .map_err(|e| L1WatcherError::Other(e.into()))?;
+    // A single commit transaction may commit a range of batches, so there can be multiple
+    // `BlockCommit` logs in this L1 block - pick the one matching our batch number.
+    let commitment = logs
+        .into_iter()
+        .find_map(|log| {
+            let event = IExecutor::BlockCommit::decode_log(&log.inner).ok()?;
```
antonbaliasnikov: nit: maybe, we need to catch decode error instead of swallowing it here?
Just wondering that seeing "event for batch not found" for decode error might be confusing in logs if this happens.

lib/l1_watcher/src/util.rs
```diff
@@ -403,79 +400,67 @@ pub async fn fetch_committed_batch_data(
...
-    };
-    Ok(ExtendedCommitBatchInfo {
+    // The batch commitment is emitted in the `BlockCommit` event (indexed by batch number) of
+    // the commit transaction. Reading it from L1 directly is safe and accurate, unlike deriving
+    // it from the current protocol version / upgrade transaction hash, which reflect the latest
+    // chain state rather than the state at the moment this batch was committed.
+    let logs = zk_chain
+        .provider()
+        .get_logs(
+            &Filter::new()
```
antonbaliasnikov: Can we use topic filter here`.topic1(B256::from(U256::from(commit_batch_info.batch_number)))`?
It would remove the need to filter all logs later on.

antonbaliasnikov (approved):

PR #1394, opened 2026-06-17
refactor(bta): encapsulate sender/receiver channel

itegulov:
## Summary

Encapsulate the base-token price `watch` channel inside `BaseTokenPriceUpdater`.

Previously the channel was created in `node/bin` and the `Sender` was threaded into the
  updater while raw `Receiver`s were handed to `FeeProvider` and `InteropFeeUpdater`. Now the
  updater owns the `Sender` and returns a cloneable `BaseTokenPriceHandle` that exposes only what
  consumers need.

Also simplifies `BaseTokenPriceUpdater` constructor by relying on `&L1State` instead of `zk_chain_l1` / `zk_chain_gateway` / `l1_provider`.


antonbaliasnikov (approved):
LGTM

PR #1396, opened 2026-06-17
feat(rpc): add method_filter to disable stateful methods

LogvinovLeon:
## Summary

Add `method_filter: HashSet<String>` on `RpcConfig`. Listed methods get -32601 \"Method disabled\" + `warn` log before any rate-limit check. New `MethodFiltering` middleware sits between `Monitoring` and `RateLimiting`.

- Default empty → deploy is a no-op.
- YAML lists or comma-separated env vars (`RPC_METHOD_FILTER=eth_newFilter,…`).
- Typical use case: filter the stateful filter family (`eth_newFilter`, `eth_getFilterChanges`, …) when running behind a load balancer without sticky sessions.


node/bin/src/config/mod.rs
```diff
@@ -1019,6 +1019,10 @@ pub struct RpcConfig {
         "each per-method limit must not exceed the global `*` limit"
     ))]
     pub rate_limits: HashMap<String, NonZeroU32>,
+
+    /// List of RPC methods to reject with -32601. Default is empty.
```
RomanBrodetski: please explain what `-32601` is and why we came up with this feature in the first place

RomanBrodetski (commented):
Looks good! Please add a short comment and I'll approve

RomanBrodetski (dismissed):

RomanBrodetski (approved):

PR #1397, opened 2026-06-17
feat(rpc): configurable rate limits via RateLimits enum

LogvinovLeon:
## Summary

Implements the rate-limit knobs from the [iRaaS Downside Protection doc](https://app.notion.com/p/matterlabs/iRaas-Downside-Protection-366a48363f23806a9c9bcdee3b212db2). The doc's tiered shape (global cap, M-bucket shared across listed methods, per-method overrides) maps directly onto a typed enum on `RpcConfig`:

```rust
pub enum RateLimits {
    None,
    Tiered { global_rps, m_rps, m_methods, custom_methods },
}
```

Operators set every value via YAML or env vars; default `None` is a no-op deploy.

- Bin uses smart-config's tagged enum (`#[config(tag = "type")]`), so env vars flatten:
  ```
  RPC_RATE_LIMITS_TYPE=Tiered
  RPC_RATE_LIMITS_GLOBAL_RPS=1000
  RPC_RATE_LIMITS_M_RPS=200
  RPC_RATE_LIMITS_M_METHODS=eth_call,eth_estimateGas,...
  RPC_RATE_LIMITS_CUSTOM_METHODS=eth_getLogs=200,eth_simulateV1=1,...
  ```
- Internal: middleware switches from `Arc<HashMap<String, RateLimiter>>` to a typed `Limiter` built from a `Limits` spec. `RateLimits::into_limits` resolves the enum.
- Drops `RpcRateLimit`, `build_limiters`, and the `rate_limits_within_global` validator.

### DIFF
--- a/lib/rpc/src/config.rs
+++ b/lib/rpc/src/config.rs
@@ -10,8 +10,37 @@ pub enum RateLimits {
     /// No rate limiting.
     #[default]
     None,
-    /// One global cap, an `m_rps` bucket shared by `m_methods`, and per-method overrides
-    /// in `custom_methods`.
+    /// One global cap, plus per-method buckets: `m_rps` applied to each entry in
+    /// `m_methods`, and the explicit RPS in `custom_methods` for each entry there.
+    ///
+    /// Production example:
+    ///
+    /// ```yaml
+    /// rpc:
+    ///   rate_limits:
+    ///     type: Tiered
+    ///     global_rps: 1000
+    ///     m_rps: 200
+    ///     m_methods:
+    ///       - eth_call
+    ///       - eth_estimateGas
+    ///       - eth_getBlockReceipts
+    ///       - eth_fillTransaction
+    ///       - zks_getProof
+    ///       - ots_getBlockTransactions
+    ///       - txpool_inspect
+    ///     custom_methods:
+    ///       eth_getLogs: 200
+    ///       eth_simulateV1: 1
+    ///       debug_traceTransaction: 10
+    ///       debug_traceCall: 10
+    ///       debug_traceBlockByHash: 10
+    ///       debug_traceBlockByNumber: 10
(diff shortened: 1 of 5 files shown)


lib/rpc/src/config.rs
```diff
@@ -1,22 +1,42 @@
...
+    #[default]
+    None,
+    /// One global cap, an `m_rps` bucket shared by `m_methods`, and per-method overrides
+    /// in `custom_methods`.
+    Tiered {
+        global_rps: NonZeroU32,
+        m_rps: NonZeroU32,
+        m_methods: HashSet<String>,
+        custom_methods: HashMap<String, NonZeroU32>,
+    },
```
RomanBrodetski, line 49: nice! Very flexible way to configure - and yet very concise.
I'd only ask you to add an example - I think it's sensible to put our production plan as an example here (that is, what we'd put in gitiops anyway - but I agree it should probably not apply by default)

RomanBrodetski (commented):
Looks good!
Please add an example (see my other comment).
Also, please ensure we have some grafana and logs coverage of this feature (maybe we already do). For logs, I think it'd be good to have e.g. one log per second that'd say how many throttled requests were there and for what reason(s). Well we can see it in grafana, but I think it shouldn't be hard to add logs. So, unless it adds some unwanted complexity in places which are complex as is, I'd still go for it

LogvinovLeon:
As for grafana - we already have panels that show errors by codes. This is just another error code that will be visible if we start hitting it. We could potentially have separate panel only for this error code. I'll add it as soon as this lands on staging and I can test it.

As for logging. I think the main reason for it is for someone looking through logs to see if throttling is happening or no.
For what reason - can be seen from metrics and I don't see a way to log it concisely. Logging the list of all methods that triggered per method limits will produce potentially very long log lines

RomanBrodetski (approved):

PR #1400, opened 2026-06-18
fix: harden ephemeral node port handling

antonbaliasnikov:
## Summary

Hardens ephemeral port handling for integration-test nodes and removes the remaining dropped-port races. The default node configuration is unchanged: `network.port` still defaults to `3060`, and tests explicitly opt into `network.port = 0` when they need ephemeral p2p ports.

## What Changed

- Replaced test-side `LockedPort` / file-lock probing with server-owned listener binding and actual bound-port reporting.
- HTTP services (`rpc`, status, prover API) are prebound once and handed into the server, so port-0 binding has no bind/drop/rebind window.

There is no production config impact:
- No default config values changed.
- `network.port` remains `3060` by default.
- Fixed nonzero p2p ports continue to fail if the configured port is already occupied.
- The new p2p retry behavior only applies when config explicitly sets `network.port = 0`.


node/bin/src/ports.rs
```diff
@@ -0,0 +1,74 @@
...
+        )
+        .await
+    }
+
+    async fn bind(
+        rpc_address: &str,
+        status_address: Option<&str>,
+        prover_api_address: Option<&str>,
+    ) -> anyhow::Result<Self> {
+        let rpc = bind_tcp_listener(rpc_address, "RPC").await?;
```
EmilLuta: Bikeshedding: Would it be worthwhile to have an enum for these services?

node/bin/src/ports.rs
```diff
@@ -0,0 +1,74 @@
+use crate::config::Config;
+use anyhow::Context;
+use std::net::SocketAddr;
+use tokio::net::TcpListener;
+use zksync_os_network::NetworkPorts;
+
+/// Actual ports bound by each service after `run()` starts.
+/// Fields are `None` when the corresponding service is disabled in config.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct BoundPorts {
```
EmilLuta: Bikeshedding: This seems a bit confusing. We have no bound ports IIUC, these are just declared. FWIW, binding might fail, right? Maybe worthwhile to call this something else? Thinking `ServerPorts` or something similar, with `BoundListeners` for `PreboundPorts`. Might be personal preference.

lib/network/src/protocol/config.rs
```diff
@@ -38,6 +38,7 @@ pub struct ExternalNodeVerifierConfig {
 }
 
 /// Role-specific protocol configuration for the `zks` subprotocol.
+#[derive(Debug, Clone)]
```
EmilLuta, line 41: Thank you, my miss!

lib/network/src/service.rs
```diff
@@ -169,14 +233,88 @@ pub struct PeerVerifyBatchResult {
...
+                protocol_config,
+                replay,
+                client,
+                raft_handler,
+            )
+            .await;
+        }
+
+        let mut last_error = None;
+        for attempt in 1..=EPHEMERAL_NETWORK_PORT_RESERVATION_ATTEMPTS {
```
EmilLuta, line 266: Is this on purpose? You basically loop `EPHEMERAL_NETWORK_PORT_RESERVATION_ATTEMPTS`, but then inside `reserve_ephemeral_tcp_udp_port` you loop over it again. I reckon we should do one or the other, WDYT?

lib/network/src/service.rs
```diff
@@ -169,14 +233,88 @@ pub struct PeerVerifyBatchResult {
 }
 
 impl NetworkService {
-    pub async fn new(
+    /// Builds the network service and returns the TCP/UDP ports it bound.
+    ///
+    /// When `config.port` is 0, startup first finds a port available for both TCP and UDP, then
+    /// retries if another process claims it before reth binds its sockets.
+    pub async fn new<Replay, Client>(
```
EmilLuta, line 240: Ah, the entire try/release/bind setup is a bit nasty, but TBH, I don't know of any better way to do (other than changing reth to accept listeners). Not blocking, but might be worthwhile to contribute upstream to reth. (consider this outside the scope of this PR)

integration-tests/tests/node/replay_archive.rs
```diff
@@ -67,6 +67,9 @@ async fn encrypted_replay_archive_recovers_node_storage_end_to_end(
     };
     let rocks_db_path = tester.config().general_config.rocks_db_path.clone();
     let stopped = tester.stop().await?;
+    tokio::task::spawn_blocking(zksync_os_rocksdb::RocksDB::<()>::await_rocksdb_termination)
```
EmilLuta, line 70: I believe this is outside the scope of this PR, but solves another tangential problem, right?

EmilLuta (dismissed):
Left a bunch of nits. LGTM, can be merged as is, feel free to consider if the nits are worthwhile to address.

Great improvement over previous setup.

P.S. I ran the tests locally, but no idea how to reproduce the issue you fixed. I reckon we'll know in CI if anything breaks.

integration-tests/tests/node/replay_archive.rs
antonbaliasnikov: Yeah, exactly. Another flakiness fixed along the way.

lib/network/src/service.rs
antonbaliasnikov: Yeah, I've had this PR: https://github.com/paradigmxyz/reth/pull/25424
but they don't want to accept it, as they don't understand the problem.

For now, I've decided to go with the solution that does not touch reth.

lib/network/src/service.rs
antonbaliasnikov: Added comments about it

node/bin/src/ports.rs
antonbaliasnikov: Renamed, a good idea.

node/bin/src/ports.rs
antonbaliasnikov: added a small one, yeah.

antonbaliasnikov:
@EmilLuta , I've adderessed some nitpicks and also added a small fix for cargo audit here. I believe we can merge as is to stabilize main.

EmilLuta (approved):

iluwaa (approved):

PR #1405, opened 2026-06-23
feat(external-node): retry main-node RPC instead of crash-looping on startup

LogvinovLeon:
An external node started before its main node panics in `load_remote_config`,
which k8s turns into `CrashLoopBackOff` — so it picks the main node up minutes
late instead of seconds.

Now all main-node RPC goes through one `MainNodeClient` that retries while the
main node is unreachable. Errors the main node **answers** with (e.g. an old node
missing a method) aren't retried, so the bytecode-supplier fallback still works.
The EN waits for its main node instead of crashing.

Also: collapse 3 ad-hoc RPC client builds into one; drop a priority-tree check
already enforced by config validation.

No tests — it's trivial retry/classification logic.


node/bin/src/en_remote_config.rs
```diff
@@ -90,19 +82,19 @@ pub async fn load_remote_config(
...
 #[derive(Debug)]
 pub struct MainNodeGenesisInputSource {
-    rpc_client: HttpClient,
+    client: MainNodeClient,
 }
 
 impl MainNodeGenesisInputSource {
-    pub fn new(rpc_client: HttpClient) -> Self {
-        Self { rpc_client }
+    pub fn new(client: MainNodeClient) -> Self {
```
RomanBrodetski: nit: why rename from rpc_client to client? I think `rpc_` prefix here is load-bearing - there may be various ways to access main node, and via rpc is just one of them

node/bin/src/lib.rs
```diff
@@ -136,10 +136,6 @@ pub async fn run<State: ReadStateHistory + WriteState + StateInitializer + Clone
     let node_role = config.general_config.node_role;
     let role: &'static str = node_role.as_str();
 
-    // Priority tree is required for main node
-    if node_role.is_main() && !config.general_config.run_priority_tree {
-        panic!("`general_run_priority_tree` must be true for Main Node");
-    }
```
RomanBrodetski, line 142: wait, did this just disappear?..

LogvinovLeon: It's validated in other place:
```rs
    #[config(default_t = true)]
    #[config_validate(custom(
        |root: &Config, value: &bool| !root.general_config.node_role.is_main() || *value,
        "must be true when `general.node_role=main`"
    ))]
```
it was redundant here

node/bin/src/main_node_client.rs
```diff
@@ -0,0 +1,76 @@
...
+    )
+}
+
+async fn with_retry<T, Fut>(call: impl FnMut() -> Fut) -> Result<T, ClientError>
+where
+    Fut: Future<Output = Result<T, ClientError>>,
+{
+    call.retry(RETRY)
+        .when(is_transient)
+        .notify(|err, after| tracing::warn!(%err, ?after, "main node unreachable; retrying"))
```
RomanBrodetski: please add `err` and `after` to the log body itself - the way grafana renders our logs makes it inconvinent to see labels (one needs to click every log line)

RomanBrodetski (commented):
overall looks good please see comments

node/bin/src/en_remote_config.rs
LogvinovLeon: Reverted

node/bin/src/lib.rs
RomanBrodetski: oh cool

RomanBrodetski (approved):
overall looks good please see comments

PR #1413, opened 2026-06-24
feat(rpc): L1 aggregation-hop proof for L1-settled chains (enable interop on L1)

kelemeno:
## What

Enable atomic interop between chains that settle directly on L1.

  The node now watches L1 `MessageRoot` updates, imports shared interop roots through system
  transactions once protocol version v32.0 is active, and exposes the RPC proofs required to construct
  and verify atomic bundles.

  For `LogProofTarget::MessageRoot`, an L2-to-L1 log proof now extends:

  `source log → source batch root → chain batch tree → shared L1 MessageRoot`

  The resulting proof is non-final because consuming chains authenticate it against the L1 interop root
  they imported.

### L1 interop-root ingestion

  - Add an L1 watcher for `MessageRoot.NewInteropRoot` events.
  - Resume from the persisted `starting_interop_root_id` cursor.
  - Resolve the corresponding historical L1 block through the MessageRoot counter.
  - Deduplicate repeated updates for the same log ID before forwarding roots to the interop subpool.
  - Keep the watcher running before v32 so a live v31 → v32 upgrade does not require restarting the
  node.

  ### MessageRoot proofs

  - Resolve the deployed L1 MessageRoot through `Bridgehub.messageRoot()`.
  - Reconstruct the source chain’s batch-tree and chain-tree proof segments at the batch’s L1 execution
  block.
  - Bind the batch leaf to the `l1Timestamp` emitted by `AppendedChainBatchRoot`.
  - Return the L1 execution block through the legacy `gatewayBlockNumber` response field.
  - Preserve the existing final-node behavior for the default `L1BatchRoot` target used by flows such
  as withdrawal finalization.

  ### Atomic-interop IMT RPCs

  Add:

  - `zks_getImtInclusionProof(commitValue, blockNumber)` to return a historical commitment-tree
  membership proof.
  - `zks_getImtLowNullifierIndex(value, blockNumber)` to return the predecessor leaf needed when
  inserting a value.

  The node reads historical leaf preimages through local `eth_call`, reconstructs the contract’s
  dynamic-height Indexed Merkle Tree, and self-verifies generated inclusion proofs before returning
  them.


lib/l1_watcher/src/interop_watcher.rs
```diff
@@ -0,0 +1,130 @@
...
+        log: Log,
+    ) -> Result<(), L1WatcherError> {
+        let event = NewInteropRoot::decode_log(&log.inner)?.data;
+
+        let log_id: u64 = event
+            .logId
+            .try_into()
+            .map_err(|e: FromUintError<u64>| L1WatcherError::Other(e.into()))?;
+
+        if log_id < self.starting_interop_root_id {
```
RomanBrodetski, line 113: maybe add a comment that this may only happen for the very first l1 block that we are processing

lib/l1_watcher/src/util.rs
```diff
@@ -156,6 +157,35 @@ pub async fn find_l1_execute_block_by_batch_number(
     .await
 }
 
+/// Finds the first L1 block where MessageRoot's counter reached `next_interop_root_id`.
+///
+/// The input is the next root the chain has not imported. For example, cursor 42 resolves to the
+/// block that advanced the counter to 42. The watcher rescans that block and drops any log with an
+/// id below 42, so roots are not skipped across a restart. A zero cursor has no on-chain anchor yet
+/// and starts scanning at block 0.
```
RomanBrodetski: hairsplitting: perhaps move or duplicate this comment to lib/l1_watcher/src/interop_watcher.rs? This comment really helped me to understand the filtering logic/reasoning in that file. And this file is a generic lib/l1_watcher/src/util.rs.

lib/rpc/src/interop_commitment_tree.rs
```diff
@@ -0,0 +1,115 @@
...
+        commit_value: U256,
+        leaf_index: u64,
+        recomputed_root: B256,
+        tree_root: B256,
+    },
+}
+
+/// Loads index-ordered commitment-tree leaves through this node's local `eth_call` implementation.
+pub(crate) struct InteropCommitmentTreeReader<RpcStorage> {
+    eth_call_handler: EthCallHandler<RpcStorage>,
```
RomanBrodetski, line 50: it took me some thinking to realize that we are (re)using eth call logic here to get the tree - relying on the system contract implementation.

This makes sense, and in fact there is a comment already - so can leave as is, or maybe we can enhance the documentaion somehow

RomanBrodetski: Maybe just briefly ecxplain where/how the tree is stored

lib/rpc_api/src/types.rs
```diff
@@ -37,41 +37,73 @@ impl RpcBlockConvert for Sealed<alloy::consensus::Block<TxHash>> {
...
-    /// Proof anchored to the SL L1 batch aggregated root.
+    /// Stops at the source chain's batch root committed to L1.
     ///
-    /// The proof covers the full gateway batch range and includes the local-root extension,
-    /// making it suitable for L1 verification.
+    /// This is the default used by L1 verification flows such as withdrawal finalization.
     #[default]
     L1BatchRoot,
-    /// Proof anchored to the SL block-level message root.
+    /// Extends the proof to L1's shared MessageRoot at the source batch's execution block.
```
RomanBrodetski: it says "committed to L1" for L1BarchRoot variant above, and "source batch's execution block" here. Is this on purpose? Is this load-bearing?
Overall the comments in this enum are very helpful, thanks.

lib/rpc_api/src/types.rs
```diff
@@ -37,41 +37,73 @@ impl RpcBlockConvert for Sealed<alloy::consensus::Block<TxHash>> {
...
+    /// Source batch root reached by the first proof segment.
     pub root: B256,
-    /// The gateway block number where this L2 batch was executed.
+    /// Settlement-layer block used by a [`LogProofTarget::MessageRoot`] proof.
     ///
-    /// Always `None`: Gateway settlement is not supported, so the local batch root is the final
-    /// proof node. Kept in the response shape for API compatibility.
+    /// Despite the legacy field name, this is the L1 block where the source batch was executed.
+    /// It is `None` when [`LogProofTarget::L1BatchRoot`] stops at the source batch root.
     pub gateway_block_number: Option<u64>,
```
RomanBrodetski: can we rename the field and have some attribute support/use the old name for serde? 

Separately,
> It is `None` when [`LogProofTarget::L1BatchRoot`] stops at the source batch root.` 

is this well-formed? `LogProofTarget::L1BatchRoot` assumes  "stops at the source batch root", so I got a little confused

RomanBrodetski (dismissed):
looks very good - just some nits about docs

lib/rpc_api/src/types.rs
itegulov: Yes, I rephrased it so hopefully it reads better now

lib/rpc_api/src/types.rs
itegulov: Good point, I actually decided to rename this field wholesale to `settlement_layer_block_number` (agnostic of whether it's L1 or GW). Gateway-based interop was never used in production so IMHO it makes sense to break backwards compatibility now. We'll have to update interop SDK (and other tooling) regardless.

RomanBrodetski (approved):

PR #1421, opened 2026-06-29
feat: zksync-os 0.4.0

dimazhornyk:
## Summary

Integrates the zksync-os 0.4.0 STF (draft-0.4.0 branch) and switches batch proving to native
  batch PIG for the new protocol version.

  - New protocol lane: ExecutionVersion::V7 (protocol 0.32.x/0.33.x) runs blocks on the zksync-os
  0.4.0 forward system; ProvingVersion::V8 (0.32.1+) with the real VK hash generated from
  zksync-os v0.4.0, airbender 73d69b5, and zkos-wrapper a9eec62.
  - New crate lib/native_pig: isolates version-specific native batching APIs from multivm;
  generate_batch_run replays a batch's blocks and produces prover input, canonical pubdata,
  state/DA commitments, and L1↔L2 counters in one shot.
  - Batcher: for V8+ batches, batch_builder builds PendingBatchInfo from the native batch run's
  canonical commit data instead of accumulating per-block outputs; legacy path is kept for older
  versions.
  - Pubdata accounting: BlockOutput.pubdata becomes an enum (Bytes for legacy versions, Length for
  V7, since canonical pubdata is now produced at batch level); seal criteria track pubdata_used
  accordingly.
  - Chain config: chain id moves from per-block metadata into the per-batch ChainConfig on V7.
  - Telemetry: new pig_telemetry module recording per-batch and per-block PIG stats (mode, prover
  input size, elapsed).
  - Toolchain/deps: bumps to nightly-2026-02-10; older zksync-os/airbender lanes temporarily point
  at bot-fork compat branches until upstream tags are patched, and airbender-crypto is patched to
  v0.2.3 for the host bigint delegation fix.
  - Tests/docs: batch verifier integration tests, batch accumulator and native-PIG unit tests, and
  a new dependency_cleanup.md design doc.


lib/types/src/protocol/proving_version.rs
```diff
@@ -35,6 +36,7 @@ impl TryFrom<ProtocolSemanticVersion> for ProvingVersion {
             (31, 0) => Ok(ProvingVersion::V7),
             (31, 1) => Ok(ProvingVersion::V7),
             (32, 0) => Ok(ProvingVersion::V7),
```
EmilLuta: Shouldn't this be v8 as well?

node/bin/src/pig_telemetry.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Debug, Clone)]
+pub struct BlockPigTelemetry {
+    pub chain_id: u64,
+    pub block_number: u64,
+    pub proving_version: ProvingVersion,
+    pub prover_input_words: usize,
+    pub elapsed: Duration,
+}
+
+static BATCH_PIG_TELEMETRY: OnceLock<Mutex<Vec<BatchPigTelemetry>>> = OnceLock::new();
```
EmilLuta: These lines make me feel very uneasy. I'd really like to bind these vectors to a specific size, rather than allowing them to be "infinite". One day, a bug will turn this decision into OOM storms.

EmilLuta (commented):
I did a quick drive-by code review. Left minor points. Overall looks good to me.

With respect to PIG, based on comments, this is an intentional decision. FWIW, we're going to lose the chance to generate prover inputs ASAP => in slower proofs => slower commits (i.e. will make interop much slower).

If this has been discussed and agreed upon as acceptable damage, disregard the comment above.

lib/types/src/protocol/proving_version.rs
```diff
@@ -35,6 +36,7 @@ impl TryFrom<ProtocolSemanticVersion> for ProvingVersion {
...
+            (32, 1) => Ok(ProvingVersion::V8),
```
itegulov: There is no v0.32.1 yet, right? I think we should just map v0.32.0 to v8 as Emil pointed out above

node/bin/src/batcher/batch_builder.rs
```diff
@@ -24,30 +35,92 @@ pub(crate) fn seal_batch<ReadState: ReadStateHistory>(
...
-        pubdata_mode,
-        sl_chain_id,
-        multichain_root,
-        &protocol_version,
-        &last_replay_record.block_context.block_hashes.0,
-    );
+    let native_batch_run = match proving_version {
+        ProvingVersion::V8 => {
+            let started_at = std::time::Instant::now();
+            let batch_run = generate_batch_run(
```
itegulov: This is a CPU-heavy task right? Should be in `spawn_blocking` if so

lib/batch_verification/src/verifier/mod.rs
```diff
@@ -109,22 +117,58 @@ impl<Finality: ReadFinality, ReadState: ReadStateHistory>
...
-            self.chain_id,
-            request.batch_number,
-            request.pubdata_mode,
-            self.l1_state.sl_chain_id,
-            multichain_root,
-            &blocks.first().unwrap().1.protocol_version,
-            &last_replay_record.block_context.block_hashes.0,
-        );
+        let (batch_info, _) = if proving_version >= ProvingVersion::V8 {
+            let native_batch_run = generate_batch_run(
```
itegulov: Same thing here

node/bin/src/prover_api/fri_proof_verifier.rs
```diff
@@ -59,10 +60,128 @@ pub fn verify_fri_proof(
...
+    use verifier_common_0_4_0::SecurityModel;
+
+    static UNIFIED_LEVEL_DATA: std::sync::OnceLock<(
+        UnrolledProgramSetup,
+        execution_utils_0_4_0::setups::CompiledCircuitsSet,
+    )> = std::sync::OnceLock::new();
+
+    const SECURITY: SecurityModel = SecurityModel::Security80;
+
+    let (setup, layouts) = UNIFIED_LEVEL_DATA.get_or_init(|| {
```
itegulov: Hmm, and maybe here as well? Less confident here though, needs to be validated

node/bin/src/batcher/batch_builder.rs
```diff
@@ -145,15 +222,17 @@ fn compute_batch_prover_input(
...
-) -> anyhow::Result<ProverInput> {
+    native_batch_run: Option<NativeBatchRunOutput>,
+) -> anyhow::Result<(ProverInput, Option<BatchPigMeasurement>)> {
     use zk_os_forward_system::run::generate_batch_proof_input;
     use zk_os_forward_system_prev::run::generate_batch_proof_input as generate_batch_proof_input_prev;
 
-    if blocks
-        .iter()
-        .any(|(_, _, _, pi)| matches!(pi, ProverInput::Fake))
+    if proving_version < ProvingVersion::V8
```
itegulov, line 235: Was this a conscious decision to not skip fake proofs for v8? Because it's much cheaper computationally?

Worth at least leaving a comment if so

node/bin/src/prover_api/fri_proof_verifier.rs
```diff
@@ -59,10 +60,128 @@ pub fn verify_fri_proof(
...
+        let setup = compute_unified_setup_for_machine_configuration::<
+            IWithoutByteAccessIsaConfigWithDelegation,
+        >(&padded_bin_bytes, &padded_text_bytes);
+        let layouts = get_unified_circuit_artifact_for_machine_type::<
+            IWithoutByteAccessIsaConfigWithDelegation,
+        >(&padded_bin_u32);
+        (setup, layouts)
+    });
+
+    verify_proof_in_unified_layer(proof, setup, layouts, false, SECURITY)
```
itegulov: It looks like new version is doing verification on the async thread. If it panics the job is never `complete_job`'d so it stays assigned until `fri_job_timeout` (48h in the recommended V8 config) and no failed-proof artifact is saved.

node/bin/src/pig_telemetry.rs
itegulov: Yes, even worse am I crazy or do they grow unbounded even right now? `take_*`/`clear_*` methods have zero callers anywhere in the workspace.

lib/multivm/build.rs
```diff
@@ -4,19 +4,39 @@ use reqwest::blocking::Client;
...
 }
 
-fn proving_version_from_tag(tag: &str) -> Option<String> {
-    match tag {
-        "v0.2.10-interface-v0.1.3" => Some(String::from("V6")),
-        "v0.3.1-interface-v0.1.3" => Some(String::from("V7")),
+fn binary_source_config(reference: &str) -> Option<BinarySourceConfig> {
+    match reference {
+        "v0.2.10-interface-v0.1.3"
+        | "antonio/compat-nightly-2026-02-10-v0.2.10-interface-v0.1.3" => {
```
itegulov: I assume most changes to this file are temporary and will be reverted once we have proper tags, correct?

lib/types/src/protocol/proving_version.rs
dimazhornyk: yes, changed it to V8
I initially used v32.1 for the integration and kept v32.0 on V7 because `(32, 0)` was added here before so I assumed it was used somewhere. I verified it is not the case, so it is safe to use v32.0 on V8

node/bin/src/pig_telemetry.rs
dimazhornyk: I removed the accumulation entirely, just kept the log lines and aggregate metrics

node/bin/src/batcher/batch_builder.rs
dimazhornyk: done, sealing now runs via spawn_blocking

node/bin/src/batcher/batch_builder.rs
dimazhornyk: it was deliberate, comment added

node/bin/src/prover_api/fri_proof_verifier.rs
dimazhornyk: good catch, added a few fixes here: 
- verification runs in `spawn_blocking`
- register extraction additionally wrapped in `catch_unwind`
- any rejected submission now unassigns the job so it's immediately re-pickable

lib/multivm/build.rs
dimazhornyk: correct, left a TODO here

node/bin/src/prover_api/fri_job_manager.rs
```diff
@@ -216,38 +223,55 @@ impl FriJobManager {
...
+                .unwrap_or([0u32; 8]);
+                Err(SubmitError::FriProofVerificationError {
+                    expected_hash_u32s,
+                    proof_final_register_values: [0u32; 16],
+                })
+            }
+            Err(join_error) => {
+                return Err(SubmitError::Other(format!(
+                    "proof verification task failed: {join_error}"
+                )));
```
itegulov, line 281: Nit: you now construct this variant here but below we only process `if let Err(SubmitError::FriProofVerificationError {`

node/bin/src/prover_api/prover_job_map/models.rs
```diff
@@ -79,6 +79,14 @@ impl JobMetadata {
         self.assigned_to_prover_id = Some(assigned_to_prover_id);
         self.current_attempt += 1;
     }
+
+    /// Clear the assignment so the job can be picked up again immediately
+    /// (e.g. after the assigned prover submitted a proof that failed verification).
+    /// `current_attempt` is preserved as assignment history.
+    pub fn unassign(&mut self) {
+        self.assigned_at = None;
+        self.assigned_to_prover_id = None;
```
itegulov, line 88: You don't clear `current_attempt` which I think is problematic as it can lead to `unwrap` panicking in `JobBatchStats::new`

integration-tests/tests/prover_v8.rs
```diff
@@ -0,0 +1,201 @@
...
+use std::collections::BTreeMap;
+use std::time::{Duration, Instant};
+use zksync_os_integration_tests::GatewayTester;
+use zksync_os_integration_tests::assert_traits::ReceiptAssert;
+use zksync_os_integration_tests::upgrade::UpgradeTester;
+use zksync_os_server::default_protocol_version::PROTOCOL_VERSION_V31_0;
+
+#[test_log::test(tokio::test)]
+#[ignore = "requires an externally built V8 zksync_os_fri_prover binary; run manually"]
+async fn v8_native_pig_real_fri_proof_e2e() -> anyhow::Result<()> {
```
itegulov, line 34: We should be able to get rid of this in favor of adding another entry to the regular `prover` test. I think the only two blockers are:
1. Prover binary as per your comment (will be done as soon as we merge the PR and create a tag)
2. v32 local chain setup

For latter I recently generated v31 state (see https://github.com/matter-labs/zksync-os-server/pull/1462) and I actually have a skill that you can use to generate v32 in a similar fashion: https://github.com/matter-labs/zksync-os-server/commit/67f6ed7e11e8738da7b1ffcb2a3f3c7a41b22796 (I'll raise on standup whether it makes sense to keep the skills inside this repo)

lib/batch_types/src/batch_info.rs
```diff
@@ -222,6 +284,34 @@ impl PendingBatchInfo {
...
     }
 
+    /// Batch output hash exactly as the zksync-os 0.4.0 (proving V8) batch program computes it
+    /// (`BatchOutput::hash` in `basic_bootloader/.../post_tx_op/public_input.rs`): unlike the
+    /// pre-V8 [`Self::public_input_hash`] layout, it does NOT include the leading `chain_id` —
+    /// the chain id is committed through the chain config hash in the outer batch public input
+    /// instead. Used for server-side verification of V8 FRI proofs; the L1-facing commitment
+    /// (`public_input_hash`) is intentionally left unchanged until v32.0 contracts define the
+    /// on-chain layout.
+    pub fn v8_batch_output_hash(&self) -> B256 {
```
itegulov: Nit: let's use v32 as opposed to v8 (we are trying to get rid of execution version/proving version abstraction)
Same thing in other places (I'd just ask AI of your choice to go through them)

lib/batch_types/src/batch_info.rs
```diff
@@ -29,6 +29,25 @@ pub struct PendingBatchInfo {
     pub protocol_version: ProtocolSemanticVersion,
 }
 
+#[derive(Debug, Clone)]
+pub struct CanonicalBatchCommitData {
```
itegulov, line 37: Can you add a comment why this new struct is needed?

lib/storage/src/in_memory/repository.rs
```diff
@@ -383,9 +383,15 @@ impl TransactionReceiptRepository {
             .map(|tx_hash| *tx_hash)
     }
 
-    pub fn remove_by_hashes(&self, tx_hashes: &[TxHash]) {
+    /// Removes tx data for `tx_hashes`, but only entries still owned by `block_number`.
+    /// A tx with the same hash can occur in a later block (e.g. the deterministic
+    /// `SetSLChainId` placeholder is injected both on the first block of a fresh chain and
+    /// on every protocol upgrade); that block's re-insert overwrites the entry, and removing
+    /// it here would lose data the later block still needs for persistence.
+    pub fn remove_by_hashes(&self, tx_hashes: &[TxHash], block_number: BlockNumber) {
```
itegulov, line 390: > `SetSLChainId` placeholder is injected both on the first block of a fresh chain and on every protocol upgrade

Actually this shouldn't be happening. This probably has to do with the code around that place not accounting for v32 upgrade correctly.

The idea is that such transactions have salt set to `u64::MAX` because we expect there to only be one of them for the entire chain's lifespan.

node/bin/src/prover_input_generator/mod.rs
```diff
@@ -79,6 +79,7 @@ impl<ReadState: ReadStateHistory + Clone + Send + 'static> PipelineComponent
                         record: replay_record,
                         prover_input: ProverInput::Fake,
                         tree_output: tree.output,
+                        tree_data: Some(tree),
```
itegulov: Nit: I think this can be `None` if PIG is disabled

lib/batch_verification/src/verifier/mod.rs
```diff
@@ -99,32 +107,90 @@ impl<Finality: ReadFinality, ReadState: ReadStateHistory>
...
-            blocks
+        let (batch_info, _) = if proving_version >= ProvingVersion::V8 {
+            // Native batch PIG re-executes the whole batch - run it on a blocking
+            // thread to avoid stalling the async runtime.
+            let native_run_inputs = blocks
                 .iter()
-                .map(|(block_output, replay_record, tree)| {
-                    (*block_output, replay_record.transactions.as_slice(), tree)
+                .map(|(_, replay_record, tree_data)| {
+                    ((*replay_record).clone(), (*tree_data).clone())
```
itegulov: Nit: there is a lot of heavy clones here, please rethink whether they are necessary. Also, it looks like you only construct `native_run_inputs` to then immediately deconstruct it into `Vec<NativeBatchBlock>` below

If this purely because of `spawn_blocking`, then I think wrapping into `Arc` might suffice here

lib/batch_verification/src/verifier/mod.rs
```diff
@@ -226,3 +292,550 @@ impl<Finality: ReadFinality, ReadState: ReadStateHistory> PipelineComponent
...
+            &read_state,
+            tree.clone(),
+            PubdataMode::Calldata,
+        )
+        .expect("V8 native batch run failed");
+
+        let words = native_batch_run.prover_input;
+
+        let out_dir = std::env::var("V8_PROVER_INPUT_OUT")
+            .unwrap_or_else(|_| "/home/claude/v8-prover-input".to_string());
```
itegulov: This looks strange

lib/sequencer/src/model/debug_formatting.rs
```diff
@@ -80,7 +90,8 @@ impl<'a> fmt::Debug for BlockOutputDebug<'a> {
             .field("storage_writes", &o.storage_writes)
             .field("account_diffs", &o.account_diffs)
             .field("published_preimages", &preimages)
-            .field("pubdata", &Hex(&o.pubdata))
+            .field("pubdata", &BlockPubdataDbg(&o.pubdata))
+            .field("pubdata_used", &o.pubdata_used())
             .field("computaional_native_used", &o.computational_native_used)
```
itegulov: Nit: note related to your PR but I just randomly noticed that `computational` is misspelled here. Could fix that please?

itegulov (commented):
LGTM overall, mostly nits other than the `SetSLChainId` issue. I think we can merge this soon (pending on other crates' PRs being merged and proper releases cut for them first)

node/bin/src/prover_api/fri_job_manager.rs
dimazhornyk: good catch! it was a nasty miss; it basically made this code return `Ok(())` for `SubmitError::Other`

node/bin/src/prover_api/prover_job_map/models.rs
dimazhornyk: fixed

integration-tests/tests/prover_v8.rs
dimazhornyk: I removed the `prover_v8.rs` for now, will extend the regular `prover` test  either before merging this PR or as a follow-up

lib/batch_types/src/batch_info.rs
dimazhornyk: done

lib/batch_types/src/batch_info.rs
dimazhornyk: done

lib/storage/src/in_memory/repository.rs
dimazhornyk: ok, fixed the comment

node/bin/src/prover_input_generator/mod.rs
dimazhornyk: yep, now it is `None` for <V8
it must stay `Some` for V8+ because the batcher's seal path runs the native batch run even with PIG disabled

lib/batch_verification/src/verifier/mod.rs
dimazhornyk: fixed

lib/batch_verification/src/verifier/mod.rs
dimazhornyk: it was an artifact from claude's tests, removed

lib/sequencer/src/model/debug_formatting.rs
dimazhornyk: fixed

integration-tests/tests/prover_v8.rs
itegulov: Looks like the file is still there btw

itegulov (dismissed):
LGTM, I think we can merge this as soon as prover stuff is settled

Cargo.toml
```diff
@@ -134,7 +136,11 @@ zk_os_forward_system_0_1_2 = { package = "forward_system", git = "https://github
 ], default-features = false }
 # v0.2.x is not historical yet, but we use a separate version for simulation so that `eth_estimateGas` can work with
 # 0-balance accounts. The fix was not a part of v0.2.5 and unfortunately cannot be included without changing `app.bin`.
-zk_os_forward_system_0_2_8 = { package = "forward_system", git = "https://github.com/matter-labs/zksync-os", tag = "v0.2.10-simulation-only-interface-v0.1.3", features = [
+zk_os_forward_system_0_2_8 = { package = "forward_system", git = "https://github.com/matter-labs/zksync-os", tag = "v0.2.10-simulation-only-interface-v0.1.3-2026-02-10", features = [
+    "production",
+    "no_print",
+], default-features = false }
+zk_os_forward_system_0_2_10 = { package = "forward_system", git = "https://github.com/matter-labs/zksync-os", tag = "v0.2.10-interface-v0.1.3-2026-02-10", features = [
```
itegulov, line 137: Give that you made v0.2.x historical, I think you no longer need v0.2.8 above, no?

itegulov (dismissed):

Cargo.toml
dimazhornyk: yes, done

hatemosphere (approved):

PR #1425, opened 2026-06-30
ci(ZKD-3716): add SBOM generation and upload pipeline

asmeirs:
## Summary

The PR adds a GitHub workflow that will collect SBOM of the source code and the container image after the publish of the latter.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -101,4 +101,4 @@ jobs:
     uses: ./.github/workflows/sbom-analysis-reusable.yaml
     secrets: inherit
     with:
-      version: ${{ inputs.tag_name || (github.ref_type == 'tag' && github.ref_name) || github.sha }}"
+      version: ${{ inputs.tag_name || (github.ref_type == 'tag' && github.ref_name) || github.sha }}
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -90,3 +90,15 @@ jobs:
...
+    name: SBOM
+    needs: [ 'build-images' ]
+    if: github.event_name != 'pull_request'
+    permissions:
+      contents: read
+      packages: read
+    uses: ./.github/workflows/sbom-analysis-reusable.yaml
+    secrets: inherit
+    with:
+      version: ${{ inputs.tag_name || (github.ref_type == 'tag' && github.ref_name) || github.sha }}"
```
LogvinovLeon: Stray `"` at the end of this line — version resolves to `<sha>"` (valid YAML, so nothing flags it). It won't show on PR CI (the job is gated to non-PR), but on the first push to main the sbom-image job will fail pulling `...:<sha>"` (invalid Docker tag), and the code/parent SBOMs get registered in Dependency Track under the malformed version.

LogvinovLeon (dismissed):

LogvinovLeon (dismissed):

onyxet (approved):

PR #1426, opened 2026-07-01
feat(replay_archive): native gcs archive support

antonbaliasnikov:
## Summary

Adds a native GCS replay-archive backend with optional GCP KMS encryption, and reworks the
archive layout and recovery path so every node writes into one shared, idempotent namespace.

### GCS backend

- New GCS storage backend built on `google-cloud-storage` / `google-cloud-auth`, alongside the
  existing S3 backend. Application Default Credentials (GKE Workload Identity, external workload
  identity federation, or local ADC) is the node auth mode; anonymous read is available for
  recovery from public buckets.
- Wired through `ReplayArchiveConfig`, the node production config (`Gcs` variant), and the
  recovery CLI (`--gcs-bucket-base-url`, `--gcs-anonymous`, `--download-concurrency`).

### GCP KMS encryption

- Archive records can be encrypted with an age recipient/identity backed by a GCP KMS asymmetric
  key (`ASYMMETRIC_DECRYPT`, `RSA_DECRYPT_OAEP_*_SHA256`), as an alternative to a local age X25519
  identity.
- Wrapping the per-record age file key with the KMS public key is fully local — the node fetches
  the public key once at startup and needs only `cloudkms.cryptoKeyVersions.viewPublicKey`, never
  `useToDecrypt`. Recovery unwraps the file key with one KMS `AsymmetricDecrypt` call per object;
  the private key never leaves KMS.
- Supported in the node config (`GcpKms` encryption variant) and the recovery CLI
  (`--kms-key-version`, `--decrypt-concurrency`). The key version resource name is embedded in each
  object's age header, so it is recoverable from the archive itself.

### Shared flat archive namespace

- All nodes and process restarts now write to one flat `<block_number>/<block_hash>` namespace
  (per-session prefixes removed). Writes are idempotent conditional creates (GCS
  `if_generation_match(0)`, S3 `If-None-Match: *`, filesystem temp-file + hard-link): the first
  writer wins and concurrent or duplicate writes are successful no-ops.
- Recovery lists keys page-by-page and interleaves concurrent downloads, resumes an interrupted
  download (skips complete objects, re-fetches `.partial` leftovers), and decodes records
  concurrently to keep the KMS-backed decrypt pipeline saturated.

### Replay record serialization

- `ZkTransaction` now (de)serializes via canonical EIP-2718 bytes in human-readable formats, and
  `SystemTxEnvelope` persists its `salt`, so JSON `ReplayRecord`s round-trip losslessly for every
  transaction type (required for archive records and block dumps). This matches the existing
  bincode encoding.

### Misc

- `cargo-audit` / `cargo-deny`: ignore `RUSTSEC-2023-0071` (Marvin timing side-channel in `rsa`
  private-key ops). Production code only performs public-key RSA-OAEP encryption; private-key
  decryption happens inside GCP KMS, and local `RsaPrivateKey` use is test-only.

## Rollout Instructions

Enabling GCS + KMS archiving is additive — it only changes where/whether the main node writes
replay records. There is no client-side or protocol change.

**Order of operations**

Roll out the config below to common configs for all nodes:

```yaml
replay_archive:
  type: Gcs
  bucket_base_url: <your-replay-archive-bucket>
  encryption:
    type: GcpKms
    # The node only fetches the public key (viewPublicKey); decryption stays IAM-gated in KMS.
    kms_key_version: projects/<project>/locations/<location>/keyRings/<ring>/cryptoKeys/<key>/cryptoKeyVersions/<version>
```

**Monitoring**

- Watch the `replay_archive_*` metrics (`archive_time`, `encryption_time`, `object_bytes`) and node
  logs for archive write failures. Archive writes apply backpressure to the pipeline if the queue
  fills, so a persistently failing backend can slow block production.

**Rollback**

- Set `replay_archive.type: Noop` (or remove the block) and redeploy. Archived objects are
  immutable and first-writer-wins, so leaving already-written objects in place is safe.
- Note: archives written in this flat layout are not readable by pre-flat-layout recovery tooling,
  and vice versa. Recovery from a KMS-encrypted archive requires `useToDecrypt` on the key version.


lib/types/src/transaction/mod.rs
```diff
@@ -62,11 +62,32 @@ impl ZkEnvelope {
...
+#[derive(Clone, Debug, PartialEq)]
 pub struct ZkTransaction {
     pub inner: Recovered<ZkEnvelope>,
 }
 
+// Serde uses the canonical EIP-2718 encoding (hex in human-readable formats like the JSON
+// replay archive records and block dumps). Field-level JSON projections of transactions
+// exist for RPC responses only and are lossy for non-Ethereum transaction types, while the
+// 2718 encoding is lossless by construction: the transaction hash commits to it.
+impl Serialize for ZkTransaction {
```
itegulov: Hmm, I am actually not sure that it's a good idea to override the entire `(De)Serialize` trait for this type. Can't we limit this logic to replay archive? e.g. using `serde_as` or something similar

lib/replay_archive/Cargo.toml
```diff
@@ -18,21 +18,32 @@ zksync_os_pipeline.workspace = true
...
 aws-config.workspace = true
 aws-runtime.workspace = true
 aws-sdk-s3.workspace = true
+bytes.workspace = true
 clap = { workspace = true, features = ["derive", "env"] }
 futures.workspace = true
+google-cloud-auth.workspace = true
+google-cloud-gax.workspace = true
+google-cloud-kms-v1.workspace = true
+google-cloud-storage.workspace = true
```
itegulov: Nit: could you gate GCP behind a feature flag that is disabled by default so we don't have to compile and link all this stuff locally?

lib/replay_archive/src/bin/replay_archive_recovery.rs
```diff
@@ -62,21 +73,36 @@ enum Command {
...
+        kms_key_version: Option<String>,
+        /// Number of replay records decoded concurrently. With KMS decryption every record decode
+        /// takes one KMS call, so this bounds in-flight KMS requests.
+        #[arg(long, default_value_t = zksync_os_replay_archive::DEFAULT_DECRYPT_CONCURRENCY)]
+        decrypt_concurrency: usize,
     },
 }
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
```
itegulov, line 120: Nit: I think it's better for this file to be in `node/replay-archive-recovery` so it's visible there's another entrypoint in the system. Or, alternatively, a new tool under `tools`. Not related to this PR though so feel free to leave for a follow-up.

lib/replay_archive/src/recovery.rs
```diff
@@ -27,16 +38,46 @@ where
...
+        let mut objects = futures::stream::iter(
+            page.keys
+                .into_iter()
+                .filter(|key| !existing.contains(key))
+                .map(|key| async move {
+                    let bytes = reader.fetch_object(&key).await?;
+                    anyhow::Ok((key, bytes))
+                }),
+        )
+        .buffer_unordered(download_concurrency.max(1));
```
itegulov: Nit: I think doing `.max(1)` here is a bit of an anti-pattern. I'd just enforce `NonZeroU64` in the CLI if possible.

node/bin/src/config/mod.rs
```diff
@@ -1495,6 +1503,14 @@ pub enum ReplayArchiveEncryptionConfig {
         /// age X25519 recipient public key. The node only needs this public key.
         recipient: String,
     },
+    /// GCP KMS encryption using Application Default Credentials. This supports GKE Workload
+    /// Identity, external workload identity federation, and local ADC.
+    GcpKms {
+        /// KMS key version (`projects/../cryptoKeyVersions/..`), purpose `ASYMMETRIC_DECRYPT`
```
itegulov: I'd use the full resource name here (without ellipsis). Otherwise, operator has to reconstruct the resource name format himself and AFAIU it's not validated anywhere on our end.

lib/replay_archive/src/recovery.rs
```diff
@@ -61,35 +101,55 @@ pub async fn recover_replay_records_to_rocksdb(
         anchor_block_number,
         anchor_block_hash,
         None,
+        DEFAULT_DECRYPT_CONCURRENCY,
     )
     .await
 }
 
+/// Default number of replay records decoded concurrently during recovery.
+pub const DEFAULT_DECRYPT_CONCURRENCY: usize = 32;
```
itegulov: Nit: move to top of the file where `DEFAULT_DOWNLOAD_CONCURRENCY` is

itegulov (dismissed):
LGTM overall, mostly nits

lib/types/src/transaction/mod.rs
antonbaliasnikov: Yes, good idea. Addressed that.

lib/replay_archive/Cargo.toml
antonbaliasnikov: Added the gcp feature, yeah. Makes total sense

lib/replay_archive/src/bin/replay_archive_recovery.rs
antonbaliasnikov: Let's defer this for now, we can address in a separate pr.

lib/replay_archive/src/recovery.rs
antonbaliasnikov: Fixed.

node/bin/src/config/mod.rs
antonbaliasnikov: Addressed

lib/replay_archive/src/recovery.rs
antonbaliasnikov: Sure, done.

itegulov (approved):

iluwaa (approved):

PR #1436, opened 2026-07-07
fix(l1-sender): retry nonce-class L1 rejections instead of crashing the node

vasyl-ivanchuk:
## Summary

`l1_sender_commit` occasionally panics with:

```
Critical task `l1_sender_commit` panicked: `pipeline segment failed: server returned an error response: error code -32000: nonce too high`
```

The nonce for each L1 transaction is derived from the node's `eth_getTransactionCount(pending)`. When a block containing our previous transactions is imported on the L1 node mid-burst, the node's pool/state views are momentarily inconsistent, and a freshly submitted transaction can be seen as nonce-gapped. Gapped **blob** transactions are rejected outright (reth: `-32000 "nonce too high"`) instead of being queued — and any send error was fatal for the whole node. The rejection is transient: the same transaction is accepted seconds later (which is why a restart always recovered).

## Fix

Classify nonce-class `eth_sendRawTransaction` rejections (`nonce too low` / `nonce too high` / `nonce gap`) and retry them with a bounded backoff. Such a rejection is a definitive refusal — the transaction was **not** admitted to the pool — so re-sending is safe. Each retry re-runs with the same nonce. All other errors (including ambiguous transport failures) still propagate and remain fatal.

## Config

Two new **optional** settings, available in both `l1_sender` and `gateway_sender` sections (shown with default values):

```yaml
l1_sender:
  # Total submission attempts per transaction on nonce-class rejections
  # (5 = 1 initial send + up to 4 retries). Set to 1 to disable retrying.
  nonce_error_max_attempts: 5
  # Backoff between attempts.
  nonce_error_retry_backoff: 2s

gateway_sender:
  nonce_error_max_attempts: 5
  nonce_error_retry_backoff: 2s
```

No action needed on rollout — defaults apply if unset.

### DIFF
--- a/lib/l1_sender/src/lib.rs
+++ b/lib/l1_sender/src/lib.rs
@@ -314,7 +314,30 @@ where
                     L1_SENDER_METRICS.balance_required_for_tx[&Input::COMPONENT_ID.as_str()]
                         .set(balance_required);
 
-                    let pending_tx = self.provider.send_transaction(tx_request).await?;
+                    // A nonce-class rejection is a definitive refusal (the tx was not admitted to the pool).
+                    // Each retry re-runs the nonce filler against a fresh pending-count, so bounded retries
+                    // self-heal instead of crashing the node.
+                    let mut send_attempt = 1;
+                    let pending_tx = loop {
+                        match self.provider.send_transaction(tx_request.clone()).await {
+                            Ok(pending_tx) => break pending_tx,
+                            Err(err)
+                                if is_nonce_error(&err)
+                                    && send_attempt < self.config.nonce_error_max_attempts =>
+                            {
+                                tracing::warn!(
+                                    command_name,
+                                    range,
+                                    send_attempt,
+                                    %err,
+                                    "L1 node rejected the transaction with a nonce error; retrying"
+                                );
+                                tokio::time::sleep(self.config.nonce_error_retry_backoff).await;
+                                send_attempt += 1;
+                            }
+                            Err(err) => return Err(err.into()),
+                        }
+                    };
                     let submitted_at = Instant::now();
                     let tx_hash = *pending_tx.tx_hash();
                     let receipt_fut = self.wait_for_confirmed_receipt(tx_hash);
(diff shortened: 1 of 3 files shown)


lib/l1_sender/src/lib.rs
```diff
@@ -314,7 +314,30 @@ where
                     L1_SENDER_METRICS.balance_required_for_tx[&Input::COMPONENT_ID.as_str()]
                         .set(balance_required);
 
-                    let pending_tx = self.provider.send_transaction(tx_request).await?;
+                    // A nonce-class rejection is a definitive refusal (the tx was not admitted to the pool).
+                    // Each retry re-runs the nonce filler against a fresh pending-count, so bounded retries
+                    // self-heal instead of crashing the node.
+                    let mut send_attempt = 1;
+                    let pending_tx = loop {
+                        match self.provider.send_transaction(tx_request.clone()).await {
```
itegulov, line 331: I am _fairly_ sure this will use a new nonce on retry (because of `CachedNonceManager` logic but please validate this first as I am not 100% sure). It's probably easier to fix this in transport layer by extending `RetryService::should_retry`

vasyl-ivanchuk: you're right, alloy fillers default to `CachedNonceManager`, so a plain re-send would have handed out an ever-higher nonce and never converged.

Reworked so the filler is out of the loop entirely: each cycle reads the pending nonce once — at the quiescent point after the previous cycle's receipts are confirmed — and assigns nonces explicitly (base_nonce + i), used for both gas simulation and the sends. A mid-cycle nonce rejection is then necessarily a transient torn view on the node, so the retry just re-sends the same fixed nonce after a backoff (bounded by nonce_error_max_attempts). Genuine divergence falls through to a restart + in-flight recovery rather than shifting nonces mid-cycle.

On extending RetryService::should_retry instead: I kept it in l1_sender because that layer wraps every provider call across all components with retry tuning shared with the 429/5xx path, and would bypass the per-sender nonce_error_* knobs. Nonce rejections are txpool verdicts rather than transport failures, so I'd rather not fold them into the generic transport retry.

itegulov (approved):

PR #1438, opened 2026-07-07
feat(network): restrict external nodes to trusted peers only

LogvinovLeon:
Enable reth `trusted_nodes_only` on external nodes, so an EN only connects to its configured boot/trusted peers — reth won't dial or accept anyone else. Main nodes keep it off (they must still accept untrusted ENs to serve replays).

**Why:** blocks are unsigned, so an EN syncing from an arbitrary discovered peer could be fed forged blocks. An EN only ever syncs/verifies against the main node anyway.

**Note:** breaking — an EN whose `boot_nodes` omits its main node will no longer connect/sync.

### DIFF
--- a/lib/network/src/service.rs
+++ b/lib/network/src/service.rs
@@ -346,6 +346,8 @@ impl NetworkService {
         let boot_nodes = resolve_boot_nodes_with_retry(config.boot_nodes.clone()).await?;
         tracing::info!(?genesis, ?fork_id, "initializing p2p network service");
         let (protocol_tx, protocol_rx) = mpsc::unbounded_channel();
+        // ENs only sync/verify against trusted peers; a main node must still accept untrusted ENs.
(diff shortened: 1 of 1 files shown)


EmilLuta (approved):
The change makes sense to me. Did you have a chance to manually test it? (I see no testes being added)

LogvinovLeon:
Thanks! Manually tested locally (anvil L1 + 1 MN + 2 ENs on this branch):

- **EN with the MN in `boot_nodes`** → syncs normally (stable connection to the MN, blocks applied).
- **EN without the MN in `boot_nodes`** → does **not** sync (0 blocks; connections to the MN are dropped as untrusted, even though it's reachable). That's the intended breaking behavior — I checked all our gitops chains already list their MN.

No unit test since `trusted_nodes_only` is reth connection-policy and isn't exercised by the handler-level e2e tests.

PR #1443, opened 2026-07-11
perf(l1_sender): improve fees and metrics usage

vasyl-ivanchuk:
## Summary

Under sustained compute-heavy load, `l1_sender_commit` is the pipeline bottleneck: the batch diff to upstream grows to `max_batch_diff_to_upstream` and the node suspends transaction acceptance (`pipeline backpressure: ... BatchDiffToUpstreamTooHigh`).

Profiling an incident showed each command send paying **~12 sequential RPC round-trips (~4.0s at ~330ms each against a hosted L1 endpoint)**:

- a full fee re-resolution per command, including **9 metrics-only `eth_feeHistory` calls** (3 windows × 3 percentiles, consumed by nothing but Prometheus gauges), 
- a per-command `eth_blobBaseFee` fetch (commit only),
- and finally the actual `eth_sendRawTransaction`.

### Changes

- **Resolve fee params once per cycle** (already done for gas simulation) and reuse them   for every send. Estimate freshness is not load-bearing here: the per-tx   `max_fee_per_gas` / `max_fee_per_blob_gas` are config-bound caps (`apply_fee_caps` only trims the priority tip by the estimate), and fee adequacy is decided at inclusion time — txs wait in the pool longer than any intra-cycle fee drift.

- **Fetch the blob base fee once per cycle**, and only when the cycle actually carries blob sidecars. The fetched value feeds a warning and a gauge; the tx always carries the configured cap, so this is behavior-neutral for submissions.
- **Report the 3×3 priority-fee comparison gauges from the existing 60s `report_operator_metrics_loop`**, non-fatally, instead of on every send. Previously a transient `eth_feeHistory` failure mid-send failed the whole cycle; now it logs a warning. Gauges keep updating — including while the sender is idle, which the old placement never did.

Per-command send cost drops to one `eth_sendRawTransaction` round-trip (**~206 → ~22 RPCs per 16-command commit cycle**).

| Metric | Before | After |
|---|---|---|
| Gap between consecutive submissions | 4.0s | 1.0s |
| Fee + gas-estimation phase per cycle | ~8s | ~1–2s |
| Commit throughput ceiling | ~1 batch/8.4s | ~1 batch/4.7s |

### Testing

- Crate unit tests, plus `cargo nextest run --workspace` (322 passed) and CI-exact clippy.
- Integration: `uncommitted_blocks_are_settled_after_batcher_reenabled` (full commit/prove/execute round-trip on a fresh L1, including node restart and in-flight recovery) and the `restart`/revert test group (6 tests), all green.

🤖 Generated with [Claude Code](https://claude.com/claude-code

### DIFF
--- a/lib/l1_sender/src/lib.rs
+++ b/lib/l1_sender/src/lib.rs
@@ -244,9 +244,12 @@ where
                 .pending()
                 .await
                 .context("get pending nonce for L1 sender cycle")?;
-            // Resolved once per cycle and reused for both simulation and every send below:
-            // L1 fees move at most once per L1 block, slower than a send cycle, and each
-            // resolution costs an RPC round-trip per command otherwise.
+            // Resolved once per cycle and reused for both simulation and every send below.
+            // Freshness is not load-bearing: the submitted max-fee/blob-fee caps are
+            // config-bound (`apply_fee_caps` only trims the priority tip by the estimate),
+            // and fee adequacy is decided at inclusion time — txs sit in the pool longer
+            // than any intra-cycle drift. Re-resolving per command would cost an RPC
+            // round-trip per tx without adding protection.
             let fee_params = self
                 .resolve_fee_params(fee_config, force_transaction_resubmission)
                 .await?;
--- a/lib/l1_sender/src/lib.rs
+++ b/lib/l1_sender/src/lib.rs
@@ -244,10 +244,12 @@ where
                 .pending()
                 .await
                 .context("get pending nonce for L1 sender cycle")?;
-            // Resolved once per cycle, not per command: fee caps are config-bound and
-            // whether a fee is high enough is decided when the tx is mined, not now. So
-            // re-reading fees between sends wouldn't change what we submit — it would only
-            // add an RPC round-trip per tx.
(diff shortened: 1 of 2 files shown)


lib/l1_sender/src/lib.rs
```diff
@@ -244,11 +244,17 @@ where
                 .pending()
                 .await
                 .context("get pending nonce for L1 sender cycle")?;
-            let sim_fee_params = self
+            // Resolved once per cycle and reused for both simulation and every send below.
+            // Freshness is not load-bearing: the submitted max-fee/blob-fee caps are
```
RomanBrodetski: this ai comment seems to be too lengthy to be useful. Please rewrite. 
This seems like a repeated issue, please add guidance on comment conciseness to AGENTS.md.

vasyl-ivanchuk: Updated `CLAUDE.md` (`AGENTS.md` is a symlink to it) and updated comments. For agent instructions I used reth repo for inspiration and kept then brief for now. Let's adjust along the way according to our needs.

RomanBrodetski (approved):

PR #1451, opened 2026-07-14
feat(interop)!: v32 atomic-interop chain batch root + IMT proof RPCs

AntonD3:
## Summary

Server side of the atomic-interop commitment structure (`protocol-docs/atomicity` in era-contracts): from protocol v32 a chain's batch commitment (`l2LogsTreeRoot`) is a fixed height-3 (8-leaf) keccak tree — leaf 0 = L2 logs root, leaf 1 = multichain root, leaves 2/3 = the interop commitment tree (IMT) roots at the batch boundaries. All changes are gated on `supports_l1_interop()` (v32+); released v31 chains are bit-for-bit unaffected.

  - `chain_batch_root`: bit-for-bit mirror of `ChainBatchRootTree` / the zksync-os `compute_chain_batch_root`, locked by unit tests. Used by the proof RPCs only — v32 sealing/verification takes the commitment from the native batch run, where the batch program computes the root itself.
  - `zks_getL2ToL1LogProof`: for v32+ batches the log-leaf path extends with the three chain-batch-root siblings; v31 keeps `keccak(localRoot, multichainRoot)`.
  - IMT proof RPCs return complete, on-chain-ready `AtomicInteropProof` consumables: `zks_getImtInclusionProof` anchors at the batch-END root (leaf 3) and carries the settlement half (metadata word, 3 siblings, batch-leaf / chain-tree / shared-tree hops + `l1Timestamp`); new `zks_getImtNonInclusionProof` serves the timeout BEGIN branch (leaf 2) via the low-nullifier leaf. Both self-verify and cross-check the reconstructed root against the tree contract's storage.
  - `read_commitment_tree_root` derives the root slot from the FullMerkle layout (`_height` at slot 0, `_nodes[_height][0]` from base slot 2), matching the bootloader; slot math locked by unit tests.
  - v32 L1-facing batch commitment is the chain-id-free batch output hash (`Committer.batchOutputHash`); the chain id moved into the chain config hash.
  - v32 batch proof public input folds `v32_chain_config_hash` between the state commitments and the batch commitment (`Executor._getBatchProofPublicInputZKsyncOS`); pre-v32 keeps the 3-word fold. SNARK groups never span a protocol-version boundary — a mixed group can verify on neither the pre- nor post-upgrade executor and wedges the chain at the upgrade, since executing the pre-boundary batches is a precondition of the upgrade that would release the rest.
  - zksync-os 0.4.0 lane pinned to `dev-20260724-atomic-interop`: the v32 lane (contracts, batch program, server-side commitment/proof code) must be one lineage — `v0.4.0-rc.1`'s `ChainConfig::hash` lacks the `pubdata_content` word the atomic executor folds. TODO: move to the release tag once cut and regenerate the V8 app binary / VK provenance.
  - Upgrade-test fixtures now vendor the v32 `CommitterFacetV32` / `ExecutorFacetV32` (from `ad-atomic-interop-da-v32-genesis`); the two `upgrade_to_v32` tests are `#[ignore]`d — the v31→v32 upgrade also needs the ecosystem-level contracts upgrade (`MessageRoot.addChainBatchRootV32` does not exist on a v31-deployed L1), which chain facet cuts cannot model. They un-ignore when the era-contracts v32 ecosystem-upgrade tooling exists.
  - `local-chains/v32.0-atomic/`: self-contained two-chain atomic preset regenerated from `ad-atomic-interop-da-v32-genesis` (genesis reports 0.32.0), plus the vendored atomic-swap driver. Main's mainline `v32.0` preset is untouched.
  - CI: cargo-deny/audit ignores for advisories published after main's last green run (`spin`/`core2` yanked, `bitmaps`/`lru` unsound — all frozen transitives; main needs the same patch).

  ## Breaking Changes

  - Protocol v32 semantics only (no v32 chain is released): the 8-leaf batch commitment, chain-id-free batch output hash and 4-word proof public input follow the atomic (v32) contracts. v31 chains, their operators and wire formats are unaffected; no DB migrations.
  - `zks_getImtInclusionProof` response type changed (adds `batchNumber` / `settlementBlockNumber` / `settlementProof`) and anchors at the batch-end root rather than the request block.
  - The zksync-os 0.4.0 lane pin moves from `v0.4.0-rc.1` to `dev-20260724-atomic-interop`; v32 chains built from this branch run the atomic OS line.

  ## Rollout Instructions

  - Deploy order: the era-contracts atomic (v32) stack first; the server activates the new commitment and proof paths only for chains reporting >= 0.32.0.
  - Nothing to monitor for v31 chains. For v32 chains watch batch commit/prove/execute and `zks_getImt*Proof` latencies.
  - Rollback: revert the release; v31 behavior is untouched, no state repair.


lib/types/src/transaction/system/utils.rs
```diff
@@ -43,13 +43,21 @@ pub(crate) enum SystemTxInput {
...
         match self {
-            Self::ImportInteropRoots(roots) => (
-                addInteropRootsInBatchCall {
-                    interopRootsInput: roots.clone(),
-                }
-                .abi_encode(),
-                0,
-            ),
+            Self::ImportInteropRoots(roots) => {
+                // Only timestamped roots reach the mempool (the watcher drops legacy ones);
```
dimazhornyk: @AntonD3 please confirm if this is an expected behavior

Cargo.toml
```diff
@@ -170,12 +170,14 @@ full_statement_verifier_prev = { package = "full_statement_verifier", git = "htt
...
-zk_os_forward_system_0_4_0 = { package = "forward_system", git = "https://github.com/matter-labs/zksync-os.git", tag = "v0.4.0-rc.1", features = [
+# TODO: dev pre-release of the atomic-interop zksync-os line; move to the release tag once cut
+# and regenerate the V8 app binary / VK provenance with it.
+zk_os_forward_system_0_4_0 = { package = "forward_system", git = "https://github.com/matter-labs/zksync-os.git", tag = "dev-20260724-atomic-interop", features = [
     "production",
     "no_print",
 ], default-features = false }
-zk_ee_0_4_0 = { package = "zk_ee", git = "https://github.com/matter-labs/zksync-os.git", tag = "v0.4.0-rc.1" }
-zk_os_basic_system_0_4_0 = { package = "basic_system", git = "https://github.com/matter-labs/zksync-os.git", tag = "v0.4.0-rc.1" }
+zk_ee_0_4_0 = { package = "zk_ee", git = "https://github.com/matter-labs/zksync-os.git", tag = "dev-20260724-atomic-interop" }
```
dimazhornyk: as per @AntonD3, atomic interop already includes the 0.4.0 changes, but this tag needs dependency bump like here before it can be merged: https://github.com/matter-labs/zksync-os/commit/8ef4749940db45b3752cbb823354bd85eaafa3cc

dimazhornyk (commented):
overall LGTM, would be great if @AntonD3 does another review round

before merging we have to:
- make sure era-contracts tag is finalized and fixtures are generated based off the stable version
- merge zksync-os into a more permanent branch like `dev` and update the dependency here (and in the prover)

lib/types/src/transaction/system/utils.rs
itegulov: I removed legacy interop roots transactions as there were never used in v31

PR #1454, opened 2026-07-14
fix(sequencer): retry GKMS failures

vasyl-ivanchuk:
## Summary

A hosted node crashed when a Google KMS signing operation failed to reach `https://sts.googleapis.com/v1/token` (OAuth token refresh):

```
Critical task `l1_sender_execute` panicked: `pipeline segment failed: local usage error:
status: 'Unknown error', self: "http error: error sending request for url (https://sts.googleapis.com/v1/token)"`
```

Every GKMS operation goes over the network (client creation, public key fetch, and each transaction signing, incl. the token refresh performed by `gcloud_sdk` under the hood), and there were no retries on that path — a single transient Google API failure took the whole node down.

This PR wraps all GKMS operations with a bounded exponential-backoff retry, governed by a single policy shared by all GKMS keys. Each retry is logged with a `warn`; once the budget is exhausted the error propagates as before, so a permanently broken key (revoked IAM, deleted version) still surfaces via a crash.

## Configuration

New optional `gkms` section — the defaults give a per-operation budget of ~15 seconds (5 attempts, delays 1s, 2s, 4s, 8s). **Existing deployments need no changes.**

```yaml
gkms:
  retry_max_attempts: 5    # attempts per operation, incl. the initial one (default: 5)
  retry_min_delay: 1s      # delay before the first retry, doubles each retry (default: 1s)
  retry_max_delay: 1min    # cap for the delay between retries (default: 1min)
```
or via environment variables: `GKMS_RETRY_MAX_ATTEMPTS`, `GKMS_RETRY_MIN_DELAY`, `GKMS_RETRY_MAX_DELAY`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16100,6 +16100,8 @@ version = "0.20.11"
 dependencies = [
  "alloy",
  "anyhow",
+ "async-trait",
+ "backon",
  "tokio",
  "tracing",
 ]
--- a/lib/operator_signer/Cargo.toml
+++ b/lib/operator_signer/Cargo.toml
@@ -17,6 +17,11 @@ alloy = { workspace = true, default-features = false, features = [
     "network",
     "consensus",
 ] }
-tokio = { workspace = true, features = ["sync"] }
+tokio = { workspace = true, features = ["sync", "time"] }
 tracing.workspace = true
 anyhow.workspace = true
+async-trait.workspace = true
+backon.workspace = true
+
+[dev-dependencies]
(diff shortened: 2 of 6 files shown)


hatemosphere:
Superseded by #1456, which addresses the same transient STS/KMS failure path by migrating the operator signer to the official Google Cloud Rust SDK. The official auth client handles token acquisition/refresh retries, and GAX handles bounded KMS retries, so the application-level retry wrapper and new node configuration in this PR are no longer needed.

PR #1456, opened 2026-07-14
refactor(operator_signer): use official GCP KMS SDK

hatemosphere:
## Summary

- replace the Alloy GCP signer, which is coupled to the retired `gcloud-sdk`, with a small Alloy signer adapter backed by the official `google-cloud-kms-v1` client
- use Application Default Credentials for GKE Workload Identity, external Workload Identity Federation, and local ADC without application-level credential parsing
- keep the existing `SignerConfig::gcp_kms` configuration and lazy client cache unchanged
- use bounded GAX retries for transient KMS failures and the auth library retry path for token acquisition and refresh
- validate the KMS key algorithm, resource identity, and CRC32C request/response integrity
- reuse `k256` and Alloy for DER parsing, low-S normalization, recovery ID derivation, transaction chain handling, and signature conversion
- add mock-KMS coverage for signing, idempotency, integrity failures, algorithm validation, retry bounds, and debug redaction

Supersedes #1454. This addresses the same transient STS/KMS failure path while removing the retired SDK instead of adding an application-level retry/configuration layer around it.

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --all-targets --all-features --workspace --exclude zksync_os_integration_tests -- -D warnings`
- [x] `cargo test -p zksync_os_operator_signer`
- [x] `cargo nextest run --workspace --exclude zksync_os_integration_tests` (337 passed)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -818,7 +818,7 @@ dependencies = [
  "lru",
  "parking_lot",
  "pin-project",
- "reqwest 0.13.3",
+ "reqwest 0.13.4",
  "serde",
  "serde_json",
  "thiserror 2.0.17",
@@ -886,7 +886,7 @@ dependencies = [
  "alloy-transport-ws",
  "futures",
  "pin-project",
- "reqwest 0.13.3",
+ "reqwest 0.13.4",
  "serde",
  "serde_json",
  "tokio",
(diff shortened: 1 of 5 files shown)


vasyl-ivanchuk (approved):

vasyl-ivanchuk:
LGTM! Checked that batches are being commited, executed and proven on stage.

antonbaliasnikov (approved):

PR #1458, opened 2026-07-15
perf(storage): stop depending on persisted block hashes 

RomanBrodetski:
## Summary

Every replay WAL row embeds the 256 previous block hashes (~8 KiB per block) inside the bincode-encoded `Context` value, even though the same hashes are derivable from the `CanonicalHash` CF (`block_number → hash`, 32 bytes, written atomically with every canonical block since #867). At the current block count that's ~111 GiB of redundant data.

This PR is the **expand phase** of an expand → migrate → contract removal, kept fully non-breaking in both directions:

- **Canonical rows are dual-written**: a stripped context (everything except the block hashes, ~120 B) goes to the new `ContextV2` CF; the full legacy row is still written to `Context`, so older binaries keep working on a DB touched by this version and rollback stays safe.
- **Reads prefer the stripped row** and reconstruct the hashes from `CanonicalHash` with a single `multi_get`. Slots belonging to blocks appended before `CanonicalHash` existed fall back to the row's legacy copy.
- **Rows without a stripped copy** (written before this version, or by an older binary during a rollback window) are served from `Context` as stored. Non-canonical (hash-keyed) rows live only there: `CanonicalHash` entries for their ancestors may be overwritten by the new canonical chain, so their embedded hashes stay authoritative.

The **contract phase** (follow-up, on top of proper migration infrastructure) backfills `CanonicalHash` for pre-#867 history and deletes the legacy canonical `Context` rows, reclaiming the ~8 KiB per block. After contraction, rollback to *this* version remains safe (reads here never require the legacy copy once `CanonicalHash` is complete).

## Testing

- New unit tests: corrupting the embedded hashes of stored rows doesn't affect reads (proof nothing depends on the persisted copy); deleting `CanonicalHash` entries exercises the pre-#867 fallback; deleting stripped rows simulates a rollback window; the override test covers reverted rows (readable by hash with original hashes, never dual-written, subsequent block chains from the replacement).
- Integration tests pass locally: `node_stop_and_restart_preserves_state`, `external_node::transaction_replay`, `rebuild_after_emptying_historical_block_preserves_unrelated_l2_txs`, `encrypted_replay_archive_recovers_node_storage_end_to_end`.

### DIFF
--- a/lib/storage/src/db/replay.rs
+++ b/lib/storage/src/db/replay.rs
@@ -30,6 +30,8 @@ use zksync_os_types::{BlockStartCursors, ProtocolSemanticVersion};
 #[derive(Clone, Debug)]
 pub struct BlockReplayStorage {
     db: RocksDB<BlockReplayColumnFamily>,
+    /// Shared by all blocks; stripped rows don't persist it (see [`StoredBlockContextV2`]).
+    chain_id: u64,
 }
 
 /// Column families for storage of block replay commands.
@@ -111,7 +113,10 @@ impl BlockReplayStorage {
             .expect("Failed to open BlockReplayStorage")
             .with_sync_writes();
 
-        let this = Self { db };
+        let this = Self {
+            db,
+            chain_id: genesis.state().await.context.chain_id,
+        };
         let inserted_genesis = if this.latest_record_checked().is_none() {
             let genesis_tx = genesis.genesis_upgrade_tx().await;
             let genesis_context = &genesis.state().await.context;
@@ -142,11 +147,11 @@ impl BlockReplayStorage {
     ///
     /// This is intended for recovery tooling that rebuilds the DB from archived replay records and
     /// writes the recovered chain from genesis upward.
-    pub fn new_without_genesis(db_path: &Path) -> Self {
+    pub fn new_without_genesis(db_path: &Path, chain_id: u64) -> Self {
         let db = RocksDB::<BlockReplayColumnFamily>::new(db_path)
             .expect("Failed to open BlockReplayStorage")
             .with_sync_writes();
-        Self { db }
+        Self { db, chain_id }
     }
 
     fn write_replay_unchecked(&self, sealed_record: Sealed<ReplayRecord>, is_canonical: bool) {
@@ -393,7 +398,9 @@ impl BlockReplayStorage {
                 stored.block_number, block_number,
                 "block number mismatch when reading context"
             );
-            return Some(stored.into_context(self.read_block_hashes(block_number, key)));
+            return Some(
+                stored.into_context(self.chain_id, self.read_block_hashes(block_number, key)),
+            );
         }
         // No stripped row: either a canonical row written before `ContextV2` existed (or by an
         // older binary during a rollback), or a non-canonical (hash-keyed) row holding a
@@ -678,17 +685,16 @@ pub struct StorageForcePreimages {
     pub preimages: Vec<(B256, Vec<u8>)>,
 }
 
-/// [`BlockContext`] as persisted in the `ContextV2` CF: without the 256 previous block hashes.
-/// The hashes are derivable data (hashes of the previous 256 canonical blocks), so they are not
-/// persisted per block (~8 KiB each); [`BlockReplayStorage::read_block_hashes`] reconstructs them
-/// from the `CanonicalHash` CF on read.
+/// [`BlockContext`] as persisted in the `ContextV2` CF: without the 256 previous block hashes
+/// (derivable from the `CanonicalHash` CF, ~8 KiB per block; see
+/// [`BlockReplayStorage::read_block_hashes`]) and without the chain id (shared by all blocks,
+/// stored once in the `Meta` CF).
 ///
 /// The exhaustive destructuring in the conversions below keeps this struct in sync with
 /// [`BlockContext`]: a field added there fails compilation here, prompting a decision on whether
 /// and how to persist it.
 #[derive(Debug, bincode::Encode, bincode::Decode)]
 struct StoredBlockContextV2 {
-    chain_id: u64,
     block_number: u64,
     timestamp: u64,
     #[bincode(with_serde)]
@@ -711,7 +717,7 @@ struct StoredBlockContextV2 {
 impl StoredBlockContextV2 {
     fn strip(context: BlockContext) -> Self {
         let BlockContext {
-            chain_id,
+            chain_id: _,
             block_number,
             block_hashes: _,
             timestamp,
@@ -726,7 +732,6 @@ impl StoredBlockContextV2 {
             blob_fee,
         } = context;
         Self {
-            chain_id,
             block_number,
             timestamp,
             eip1559_basefee,
@@ -741,9 +746,8 @@ impl StoredBlockContextV2 {
         }
     }
 
-    fn into_context(self, block_hashes: BlockHashes) -> BlockContext {
+    fn into_context(self, chain_id: u64, block_hashes: BlockHashes) -> BlockContext {
         let Self {
-            chain_id,
             block_number,
             timestamp,
             eip1559_basefee,
@@ -874,7 +878,7 @@ mod tests {
     #[tokio::test]
     async fn reconstructs_hashes_from_canonical_hash_cf() {
         let dir = tempfile::tempdir().unwrap();
-        let storage = BlockReplayStorage::new_without_genesis(dir.path());
+        let storage = BlockReplayStorage::new_without_genesis(dir.path(), 270);
         // Crosses the 256-hash window boundary: early blocks are zero-padded, later ones aren't.
         let chain = make_chain(300);
         for sealed in &chain {
@@ -894,7 +898,7 @@ mod tests {
     #[tokio::test]
     async fn falls_back_to_embedded_hashes_when_canonical_hash_missing() {
         let dir = tempfile::tempdir().unwrap();
-        let storage = BlockReplayStorage::new_without_genesis(dir.path());
+        let storage = BlockReplayStorage::new_without_genesis(dir.path(), 270);
         let chain = make_chain(300);
         for sealed in &chain {
             storage.write(sealed.clone(), false).await.unwrap();
@@ -908,7 +912,7 @@ mod tests {
     #[tokio::test]
     async fn rows_without_stripped_copy_are_served_from_legacy() {
         let dir = tempfile::tempdir().unwrap();
-        let storage = BlockReplayStorage::new_without_genesis(dir.path());
+        let storage = BlockReplayStorage::new_without_genesis(dir.path(), 270);
         let chain = make_chain(300);
         for sealed in &chain {
             storage.write(sealed.clone(), false).await.unwrap();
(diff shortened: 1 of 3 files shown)


lib/storage/src/db/replay.rs
```diff
@@ -40,7 +40,21 @@ pub struct BlockReplayStorage {
...
+    /// Full [`BlockContext`], including the 256 previous block hashes (~8 KiB per block).
+    ///
+    /// For canonical rows this is a compatibility copy: reads prefer the stripped row in
+    /// [`Self::ContextV2`] and only fall back here (for rows written before `ContextV2`
+    /// existed, or by an older binary during a rollback). Non-canonical (hash-keyed) rows live
+    /// only here.
+    ///
+    /// TODO(RocksDB migration): once `ContextV2` has proven itself in production, stop writing
+    /// canonical rows here and migrate pre-existing ones (backfilling `CanonicalHash` for
+    /// history predating that CF along the way), leaving this CF to non-canonical rows only.
```
RomanBrodetski: make comment shorter and to the point

lib/storage/src/db/replay.rs
```diff
@@ -265,44 +294,129 @@ impl BlockReplayStorage {
...
+            .db
+            .multi_get_cf(BlockReplayColumnFamily::CanonicalHash, keys.iter());
+        let mut embedded = None;
+        for (number, result) in (first..block_number).zip(results) {
+            let index = (number + 256 - block_number) as usize;
+            hashes[index] = match result.expect("Failed to read from CanonicalHash DB") {
+                Some(bytes) => U256::from_be_slice(&bytes),
+                None => {
+                    let embedded = embedded.get_or_insert_with(|| {
+                        self.get_legacy_context(key)
```
RomanBrodetski, line 360: add a warning here - we don't expect that tthere are still such blocks that need to be read

RomanBrodetski:
Review comments addressed (shorter `Context` CF doc, `tracing::warn!` on the embedded-hashes fallback) and rebased onto main.

**Root cause of the CI failure — not related to this PR.** `upgrade_to_v32_with_deployments` fails identically on this branch's merge-base (`61c5a3f5`, without any of these changes): the pre-#1462 canned `v31.0` state was a time bomb. Its L1 dump carries anvil's clock warped −30 days, and its L2 blocks were stamped 2026-06-15 08:57 UTC; once wall time passed 2026-07-15 ~08:57 (30 days later), anvil's clock overtook the canned block timestamps and committing the canned uncommitted tail (batch 12) started reverting with `TimeNotReached(1781513827, <anvil now>)`. This PR's CI ran at 09:45 that morning — 53 minutes past expiry, likely the first run to trip it. #1462 removed the canned chain DB the same day (the test now starts from an empty chain), so the rebase resolves it: the test passes locally on the rebased branch in 42s.

lib/storage/src/db/replay.rs
```diff
@@ -165,6 +176,19 @@ impl BlockReplayStorage {
         // Batch writes: replay entry, latest pointer and canonical hash mapping
         let mut batch: WriteBatch<'_, BlockReplayColumnFamily> = self.db.new_write_batch();
         if is_canonical {
+            // Expand phase of dropping the persisted block hashes: canonical rows are
+            // dual-written. Reads prefer this stripped row; the full legacy row (below) keeps
+            // rollback to older binaries safe until a migration removes it.
+            let stripped_context_value = bincode::encode_to_vec(
```
itegulov, line 187: IMHO let's also stop using `bincode` while we are at it (the crate is unmaintained)

lib/storage/src/db/replay.rs
```diff
@@ -265,44 +289,136 @@ impl BlockReplayStorage {
...
+            .multi_get_cf(BlockReplayColumnFamily::CanonicalHash, keys.iter());
+        let mut embedded = None;
+        for (number, result) in (first..block_number).zip(results) {
+            let index = (number + 256 - block_number) as usize;
+            hashes[index] = match result.expect("Failed to read from CanonicalHash DB") {
+                Some(bytes) => U256::from_be_slice(&bytes),
+                None => {
+                    let embedded = embedded.get_or_insert_with(|| {
+                        // Only expected for blocks appended before `CanonicalHash` was
+                        // introduced, i.e. only on chains that haven't produced a block since.
```
itegulov, line 354: Since... what? Or am I misreading the sentence somehow?

lib/storage/src/db/replay.rs
```diff
@@ -572,3 +677,293 @@ pub struct StorageForcePreimages {
...
+/// The hashes are derivable data (hashes of the previous 256 canonical blocks), so they are not
+/// persisted per block (~8 KiB each); [`BlockReplayStorage::read_block_hashes`] reconstructs them
+/// from the `CanonicalHash` CF on read.
+///
+/// The exhaustive destructuring in the conversions below keeps this struct in sync with
+/// [`BlockContext`]: a field added there fails compilation here, prompting a decision on whether
+/// and how to persist it.
+#[derive(Debug, bincode::Encode, bincode::Decode)]
+struct StoredBlockContextV2 {
+    chain_id: u64,
```
itegulov: Nit: I'd remove this as well if not too much hassle - chain id is the same for all blocks, no need to store it per block...

lib/storage/src/db/replay.rs
```diff
@@ -311,11 +427,7 @@ impl ReadReplay for BlockReplayStorage {
         db_key: Option<Vec<u8>>,
     ) -> Option<ReplayRecord> {
         let key = db_key.unwrap_or_else(|| block_number.to_be_bytes().to_vec());
-        let Some(block_context) = self
-            .db
-            .get_cf(BlockReplayColumnFamily::Context, &key)
-            .expect("Failed to read from Context CF")
-        else {
+        let Some(block_context) = self.get_context_by_key(block_number, &key) else {
```
itegulov: I didn't think this through fully yet but isn't there an issue with overrides if you call `write(..., true)` on block N, then on block N + 1. Canonical parent hash for block N + 1 will be different from the one used during original `write` which can lead to some unexpected issues.

lib/storage/src/db/replay.rs
```diff
@@ -265,44 +289,136 @@ impl BlockReplayStorage {
...
-    fn get_context(&self, block_number: BlockNumber) -> Option<BlockContext> {
-        let key = block_number.to_be_bytes();
+    /// Reconstructs the 256 previous block hashes for the block stored under `key` from the
+    /// `CanonicalHash` CF. The hash of block `M` lands at index `M + 256 - block_number` (most
+    /// recent last); slots before genesis stay zero, mirroring how the genesis context is
+    /// initialized.
+    ///
+    /// Entries can only be missing for blocks appended before `CanonicalHash` was introduced;
+    /// those slots are taken from the hashes embedded in the row's legacy copy.
+    fn read_block_hashes(&self, block_number: BlockNumber, key: &[u8]) -> BlockHashes {
```
itegulov, line 337: This will be a pretty expensive operation I think. For normal operation I think it makes sense for sequencer to maintain rolling struct with last processed 256 block hashes (maybe ring buffer or smth like that). Can be a follow-up of course.

lib/storage/src/db/replay.rs
RomanBrodetski: I was thinking we'd consider this when we do `step 3` - after we have the simple migration machinery. We'll need to rewrite every row then anyway. Doing that now would mean we'll have differently serialized data in different CFs.. 
I can reconsider if that doesn't make sense - LMK.

lib/storage/src/db/replay.rs
RomanBrodetski: since the introduction of `CanonicalHash` :). I don't think we have such cases - hence warning.

lib/storage/src/db/replay.rs
RomanBrodetski: agreed!

lib/storage/src/db/replay.rs
RomanBrodetski: Yeah I considered this, also asked AI yesterday:
```
  As for "isn't it slow" — the numbers say no, in every path that pays it:

  - The cost is one multi_get of ≤256 keys against a dense ~600 MB CF (bloom-filtered, cached): roughly tens to a couple hundred
  µs.
  - eth_call: one reconstruction per call, followed by VM execution measured in milliseconds — low single-digit % overhead at
  worst.
  - Startup replay / EN serving: one per block, next to VM re-execution or RLP-encoding-plus-shipping a record that itself
  contains those same 8 KiB of hashes on the current wire format.
  - Historical blocks (pre-upgrade legacy rows) don't reconstruct at all — they decode their embedded copy.

  And there's a cheap escape hatch we deliberately haven't taken: sequential readers (startup, EN streams) request windows that
  overlap 255/256 with the previous one, so a one-entry rolling cache in BlockReplayStorage would make it O(1) per block. I'd add
  that only if profiling of EN sync throughput ever asks for it — the step-0 PR keeps the read path dumb and measurable instead.
```
still I agree it doesn't feel right we read those 256 keys for each block that we need for strartup/stream.. The alternative is to have som stateful intermediate when reading - which would I think also require a stateful version of 
```
    fn into_context(self, chain_id: u64, block_hashes: BlockHashes) -> BlockContext {
```
I'm on the fence. I'll have another look at this in a follow-up

lib/storage/src/db/replay.rs
RomanBrodetski: Done — dropped from the stored row entirely. `BlockReplayStorage` now carries the chain id (from genesis state; recovery tooling passes it from the archived records) and fills it in on read, so nothing chain-constant is persisted per block.

lib/storage/src/db/replay.rs
RomanBrodetski: Great catch — this was real, and worse than a transient read glitch: the override path itself would *archive* the corrupted record. `write(N, true)` repoints `CanonicalHash[N]`; `write(N+1, true)` then read old N+1 via reconstruction, splicing new N's hash into the copy re-stored under the old hash key — permanently. (Canonical data was unaffected: its hashes come from the in-memory chain and the index is consistent for the new chain. The damage was confined to the archived reverted rows for the 2nd+ blocks of a range override — but those are exactly what `en_replay_record_overrides` replays, so it matters.)

Fixed in e6feea8c: the override path now takes the old record's context from the **legacy row** (guaranteed present — canonical rows are dual-written), whose embedded hashes are the ones the block was executed with. Added `multi_block_override_archives_original_hashes`, which fails without the fix.

lib/storage/src/db/replay.rs
```diff
@@ -100,7 +113,10 @@ impl BlockReplayStorage {
             .expect("Failed to open BlockReplayStorage")
             .with_sync_writes();
 
-        let this = Self { db };
+        let this = Self {
+            db,
+            chain_id: genesis.state().await.context.chain_id,
```
itegulov: Nit: this now forces loading genesis on every startup (which can be costly). `Genesis` already holds `chain_id`, we should just expose it through a getter method

lib/storage/src/db/replay.rs
```diff
@@ -572,3 +710,336 @@ pub struct StorageForcePreimages {
     #[bincode(with_serde)]
     pub preimages: Vec<(B256, Vec<u8>)>,
 }
+
+/// [`BlockContext`] as persisted in the `ContextV2` CF: without the 256 previous block hashes
+/// (derivable from the `CanonicalHash` CF, ~8 KiB per block; see
+/// [`BlockReplayStorage::read_block_hashes`]) and without the chain id (shared by all blocks,
+/// stored once in the `Meta` CF).
```
itegulov: Nit: this comments is not accurate, we don't store chain id in `Meta` CF

lib/storage/src/db/replay.rs
RomanBrodetski: The fix is not the simplest, but once we get to step-3, we'll be able to do it much cleaner

itegulov (dismissed):

itegulov (approved):

PR #1460, opened 2026-07-15
perf(l1_watcher): batch-sweep startup commit discovery

itegulov:
## Summary

Speeds up external-node startup by reworking how `CommittedBatchProvider` discovers committed batches on L1. On a stage EN this cuts the batch-discovery phase (`L1 state` → `Node state on startup`) from **~114s to ~13s**; together with the logs-cache fix (#1452) total startup dropped from ~136s to ~26s.

Previously every batch in the startup range got an independent binary search over the full L1 history (~21 sequential archive `eth_call`s each) plus its own chunked `BlocksRevert` log scan — ~500 sequential deep-history RPCs for a typical ~24-batch range whose commits all sit within a few hundred blocks of the tip.

Three changes, sharing one idea — **verify commit liveness via `storedBatchHash` instead of scanning revert logs**:

- **Hash-based commit lookup.** `find_l1_commit_block_by_batch_number` now searches with that predicate and returns the live commit block directly
- **One sweep instead of N searches.** Startup now runs a single search for the *oldest* batch, then one chunked `eth_getLogs` pass collecting every batch's `ReportCommittedBatchRangeZKsyncOS`+`BlockCommit` events. Note: the sweep is linear in the swept window, so it can take minutes if execution finalization lags days behind the tip (still cheaper than the per-batch scans it replaces).
- **Gallop from the tip.** `find_l1_block_by_predicate` brackets exponentially from `latest` before binary-searching: ~12 probes at recent (warm) blocks instead of ~21 at deep history.

### DIFF
--- a/integration-tests/tests/node/rebuild.rs
+++ b/integration-tests/tests/node/rebuild.rs
@@ -22,7 +22,7 @@ use zksync_os_integration_tests::wallets::load_operator_private_key;
 use zksync_os_integration_tests::{
     CURRENT_TO_L1, StoppedTester, TestEnvironment, Tester, test_multisetup,
 };
-use zksync_os_l1_watcher::{fetch_batch, fetch_batch_commit_tx_hash};
+use zksync_os_l1_watcher::fetch_live_committed_batch;
 use zksync_os_operator_signer::SignerConfig;
 use zksync_os_server::config::{RebuildBounds, RebuildConfig};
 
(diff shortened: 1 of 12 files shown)


LogvinovLeon (dismissed):
This is awesome!

LogvinovLeon (approved):
The push was a mistake in branch tracking. Force-push fixed it so the PR contains only your commits again

LogvinovLeon (approved):

antonbaliasnikov (approved):

PR #1471, opened 2026-07-17
feat(storage)!: schema migration runner + contract replay WAL block hashes

RomanBrodetski:
## Summary

Retargeted to `main` now that #1458 (expand phase) has landed and soaked.

Step 2 of the replay WAL cleanup: the contract phase that reclaims the redundant persisted block hashes (measured against the real testnet WAL: ~16.7M blocks, ~146 GiB total, ~137 GiB of which is this redundancy), plus the schema-migration machinery it runs on.

### Migration runner (`lib/rocksdb`)

- `Migration<CF>` trait + `RocksDB::run_migrations`: ordered, idempotent, blocking migrations executed at open, before the database serves anything.
- Schema version stamped in a reserved `__schema_meta` CF managed by the wrapper, outside each database's CF enum — any database adopts the runner with zero schema changes and an empty migration list, gaining a clean refuse-to-start error (instead of a decode panic) when opened by a binary that doesn't know its version.
- Blocking only for now — see "Step 3" below for why background/resumable mode isn't needed yet.

### Contract migration (`contract_block_contexts`, WAL schema v1)

Per canonical row, in atomic 1000-row chunks: ensure the `CanonicalHash` entry (pre-#867 entries recovered from the successor's embedded hashes), ensure the stripped `ContextV2` row, delete the legacy ~8 KiB row; final `compact_range` reclaims the space. Idempotent and crash-resumable — each chunk is one atomic write, so a crash leaves every row either fully migrated or fully untouched; a restart safely reruns the whole scan from block 0, at the cost of cheaply re-verifying already-done rows rather than skipping them.

Timed end-to-end against a real copy of the testnet WAL (16,791,125 blocks, 148 GiB on disk): **~144 seconds** (open + migration, single pass) — ~142s for the per-row conversion loop (~8.5 µs/row) plus ~1.9s for the final compaction. Reclaimed 134 GiB on disk (148 GiB → 14 GiB), matching the ~137 GiB estimate. Comfortably fast enough that this phase stays blocking-only — see "Step 3" for the migration that will actually need background mode.

### Write/read paths after contraction

- Canonical rows are written stripped-only; the dual write from the expand phase is gone.
- Archived rows (a block displaced by an override) stay fully embedded, exactly as before this PR. Stripping them too was tried during review — a `parent_hash`-per-row scheme with a dedicated `ArchivedContext` CF, reconstructing the window by walking a chain of these on read — and dropped: overrides are rare enough that the extra CF and read-path complexity wasn't worth it.
- The one thing full embedding still needs help with: correctly archiving a block that was already displaced earlier in the same override wave (by the time it's block N+1's turn, the live canonical chain has moved on, so it no longer has the parent block N+1 was actually built on). Solved with a single in-memory `last_archived` cache — the most recently archived row, reused via the same shift-and-append the sequencer already does to build any new block's window, if the next override turns out to continue the same wave; cleared on any plain append. Not a durability mechanism: a restart landing between two overrides of the same wave loses it, and the next override's archived window ends up wrong at exactly the position the earlier override already repointed — accepted, since overrides are already very rare, multi-block waves rarer still, and a restart hitting the middle of one rarer still than that (see `override_wave_across_restart_has_known_imprecision`).
- `previous_block_timestamp` isn't persisted for any row, canonical or archived — it's always recomputed live from the current canonical neighbor on read. Correct by construction for canonical rows; for an archived row it's a pre-existing quirk (predates this PR), wrong whenever the block at its parent height has since been overridden — independent of the above, wave or no wave, restart or no restart.

## Breaking Changes

- Who is affected: node operators (main nodes and ENs).
- What breaks: after the migration runs, legacy canonical context rows are gone; binaries older than the expand phase (#1458) cannot read the database. The expand-phase binary remains a safe rollback target.
- Migration: automatic at first start, blocking. See "Step 3" for the downtime discussion.

## Rollout Instructions

- #1458 is already deployed and soaked; this lands on top of it.
- Each node (main node / EN) migrates its own WAL independently at startup; order doesn't matter.
- Watch startup logs for `running schema migration` / `schema migration finished` and the reclaimed-space compaction; alert on `schema migration failed`.
- Rollback: to the expand-phase binary at any time, including mid-migration. Rolling back further is refused explicitly by stamped databases.

## Step 3 (unification) — plan, not part of this PR

This PR is deliberately blocking-only; see the timing above for why that's fine for the contract phase. Step 3 is a different kind of migration: not a bounded delete + index-backfill, but a full re-key of *every* row under a composite key (old key tombstoned, new key inserted, for the whole history) — meaningfully more write-amplified, and the place the roadmap always expected to need background execution.

What's left for it:
- Composite `number ++ hash` key (preserves append locality vs. a pure hash key).
- Fold `NodeVersion` / `ProtocolVersion` / `BlockOutputHash` / the four `Starting*` cursors into one per-row metadata blob — 13 CFs → ~5.
- Make `previous_block_timestamp` a genuinely stored field for every row for the first time — closes the pre-existing archived-row quirk above for good.
- With composite keys, "revert" becomes a pure index repoint: the old row simply stays at its own key, never touched, forever fully correct — no separate archived-row write path, no wave-tracking cache, none of the quirks above.
- Needs: a background/resumable migration mode in the runner (not built yet — it only supports blocking today), given the re-key touches the whole history rather than a bounded set of rows.

## Testing

- Runner: version stamping, pending-only execution, refuse-on-future, failed-migration rerun, contiguity validation.
- Contract: legacy-only + pre-`CanonicalHash` history conversion (incl. successor-recovery), idempotent double-run, archived rows untouched, unrecoverable-tip error path.
- Override correctness: multi-block override wave archives the original window correctly via `last_archived` (`previous_block_timestamp`'s pre-existing quirk is asserted as such, not conflated with a regression); a restart between two overrides of the same wave is asserted to lose exactly that precision, not silently produce something worse.
- Integration (local, re-run against this design): all 9 `rebuild`/`revert_l1` tests (incl. a multi-block override via L1 revert), plus `external_node` (incl. transaction replay), `restart` (incl. recovery from an L1 batch revert), and `replay_archive` (encrypted recovery) — 23 tests, all green.

### DIFF
--- a/lib/rocksdb/src/db.rs
+++ b/lib/rocksdb/src/db.rs
@@ -670,10 +670,23 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
     /// so that tombstoned space is actually reclaimed instead of waiting for background
     /// compaction to get to it.
     pub fn compact_range_cf(&self, cf: CF) {
-        let cf = self.column_family(cf);
+        let cf_name = cf.name();
+        tracing::info!(
+            db_name = CF::DB_NAME,
+            cf_name,
+            "starting column family compaction"
+        );
+        let started_at = Instant::now();
+        let cf_handle = self.column_family(cf);
         self.inner
             .db
-            .compact_range_cf(cf, None::<&[u8]>, None::<&[u8]>);
+            .compact_range_cf(cf_handle, None::<&[u8]>, None::<&[u8]>);
+        tracing::info!(
+            db_name = CF::DB_NAME,
+            cf_name,
+            elapsed = ?started_at.elapsed(),
+            "column family compaction finished"
+        );
     }
 
     /// Returns size stats for the specified column family.
--- a/lib/rocksdb/src/db.rs
+++ b/lib/rocksdb/src/db.rs
@@ -32,6 +32,14 @@ use crate::metrics::{
 /// Not properly dropped RocksDB instances can lead to DB corruption.
 static ROCKSDB_INSTANCE_COUNTER: (Mutex<usize>, Condvar) = (Mutex::new(0), Condvar::new());
 
+/// Name of the reserved column family holding wrapper-managed metadata (currently the schema
+/// version stamped by [`crate::migrations`]). Managed outside [`NamedColumnFamily::ALL`] so that
+/// per-database CF enums don't need to declare it.
+pub(crate) const SCHEMA_META_CF_NAME: &str = "__schema_meta";
+
+/// Key under [`SCHEMA_META_CF_NAME`] holding the database schema version.
+pub(crate) const SCHEMA_VERSION_KEY: &[u8] = b"schema_version";
+
 /// Describes column family used in a [`RocksDB`] instance.
 pub trait NamedColumnFamily: 'static + Copy {
     /// Name of the database. Used in metrics reporting.
@@ -416,6 +424,7 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
                 // The default CF is created on RocksDB instantiation in any case; it doesn't need
                 // to be explicitly opened.
                 let is_obsolete = cf_name != rocksdb::DEFAULT_COLUMN_FAMILY_NAME
+                    && cf_name != SCHEMA_META_CF_NAME
                     && !cfs_and_options.contains_key(cf_name);
                 is_obsolete.then_some(cf_name)
             })
@@ -433,7 +442,9 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
         let cf_names = cfs_and_options.keys().copied().collect();
         let all_cfs_and_options = cfs_and_options
             .into_iter()
-            .chain(obsolete_cfs.into_iter().map(|name| (name, (false, None))));
+            .chain(obsolete_cfs.into_iter().map(|name| (name, (false, None))))
+            // Created on first open thanks to `create_missing_column_families`.
+            .chain([(SCHEMA_META_CF_NAME, (false, None))]);
         let cfs = all_cfs_and_options.map(|(cf_name, (requires_tuning, prefix_extractor_len))| {
             let mut block_based_options = BlockBasedOptions::default();
             block_based_options.set_bloom_filter(10.0, false);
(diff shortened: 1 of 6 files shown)


RomanBrodetski:
Rollback audit (prompted by review question): after the migration runs, a rollback to the #1458 binary previously would have **panicked on ENs** — they issue override writes for existing blocks on restart replay, and #1458's override path expected the legacy row the migration deletes. Fixed on the #1458 branch (`3b78553f`): the override read now falls back to reconstruction when the legacy copy is absent, with a regression test (`override_write_tolerates_contracted_rows`). This branch is rebased on top; the rollback claims in the description now hold for both node roles.

lib/rocksdb/src/db.rs
```diff
@@ -620,6 +631,51 @@ impl<CF: NamedColumnFamily> RocksDB<CF> {
...
+            SCHEMA_VERSION_KEY,
+            version.to_be_bytes(),
+            &options,
+        )
+    }
+
+    /// Compacts the whole key range of the column family. Used by migrations after bulk deletes
+    /// so that tombstoned space is actually reclaimed instead of waiting for background
+    /// compaction to get to it.
+    pub fn compact_range_cf(&self, cf: CF) {
```
RomanBrodetski, line 672: let's add `info` logs to this

lib/rocksdb/src/migrations.rs
```diff
@@ -0,0 +1,234 @@
...
+    fn run(&self, db: &RocksDB<CF>) -> anyhow::Result<()>;
+}
+
+impl<CF: NamedColumnFamily> RocksDB<CF> {
+    /// Runs all migrations pending for this database and stamps the schema version after each.
+    ///
+    /// Consumes and returns `self` to fit the open-time builder chain. `migrations` must be
+    /// sorted by strictly increasing `target_version` with no gaps from 1; a database can then
+    /// upgrade from any older version, including 0 (never migrated).
+    pub fn run_migrations(self, migrations: &[&dyn Migration<CF>]) -> anyhow::Result<Self> {
```
RomanBrodetski, line 44: let's add an info! log here that will state what migration versions are passed and wht's already migrated/found on disc. Think what else we can/should log at this point

lib/storage/src/db/replay.rs
```diff
@@ -42,16 +44,25 @@ pub struct BlockReplayStorage {
...
     ContextV2,
+    /// Hash-keyed archived rows (blocks displaced by an override), holding the same
+    /// [`StoredBlockContextV2`] as [`Self::ContextV2`]. Its `parent_hash` field is what makes
+    /// archived rows self-sufficient: the window and `previous_block_timestamp` are reconstructed
+    /// on read by walking `parent_hash` pointers through this CF until reaching a hash still on
+    /// the live canonical chain, then finishing with one `CanonicalHash` lookup (see
+    /// [`BlockReplayStorage::resolve_window_and_previous_timestamp`]). Unlike [`Self::Context`],
+    /// this doesn't depend on `CanonicalHash` staying unchanged, so it's correct on any read, no
+    /// matter how long after the override or in how different a process.
+    ArchivedContext,
```
RomanBrodetski: comments to the three CFs above can be improved. IIUC, we now never write to `Context` - we write canonical to `ContextV2` and archived to `ArchivedContxt`. whereas before we were writing all to Context.

Try to express this concisely. Perhaps one comment for all three? Or one common and then small ones for each.

lib/storage/src/db/replay.rs
```diff
@@ -150,22 +165,20 @@ impl BlockReplayStorage {
...
         // Prepare record
         let (record, block_hash) = sealed_record.split();
         // TODO: We want to change the key to be block_hash for all blocks
-        let db_key = if is_canonical {
-            record.block_context.block_number.to_be_bytes().to_vec()
-        } else {
-            block_hash.0.to_vec()
+        let db_key = match kind {
+            WriteKind::Canonical => record.block_context.block_number.to_be_bytes().to_vec(),
+            WriteKind::Archived => block_hash.0.to_vec(),
```
RomanBrodetski, line 205: is there a value in having this as a separate match? `db_key` is different for the two kinds - and the only place it's used is below inside another match. So, sounds like we can just use different values depending on Kind.

RomanBrodetski:
Closing in favor of #1548: we decided to go straight for the full replay-WAL schema unification (composite `(number, hash)` keys, folding all the per-row CFs into one) rather than shipping this intermediate canonical-row-only contraction first — the latter would get discarded almost immediately once the bigger unification lands, at the cost of touching every canonical row twice. See #1548 for the new plan and phase 1 (dual-write).

PR #1474, opened 2026-07-20
feat(sequencer): rate limiter on gas/s

vasyl-ivanchuk:
## Summary

Adds an optional rate limiter that gates based on how much gas the node has executed. The feature is fully opt-in. Omitting `rpc.tx_gas_rate_limit` entirely disables it — zero behavior change, zero overhead. It only runs on the main node.

## Approach

Implemented as `RateLimitedL2Subpool`, a decorator in the mempool crate that's an alternative implementation of the `L2Subpool` trait. It gates the mempool's own `add_transaction` (the single choke point every admission path funnels through — local RPC today, gossip once consensus lands) and drains from `on_canonical_state_change`, which the mempool already receives for every block it commits or replays. Because the gate lives at this level, it applies uniformly regardless of how a transaction enters the pool, rather than only covering `eth_sendRawTransaction`.

A single shared "gas bank":

- **Drain**: every sealed block subtracts the gas it actually executed.
- **Refill**: wall-clock time adds gas back at `gas_per_second`.
- **Gate**: while the bank is empty, non-exempt submissions are rejected with JSON-RPC code `-32005` and a `retryAfterMs` hint. The gate reopens only once the bank has recovered some credit, so acceptance flips in windows of seconds rather than flapping per transaction.
- **Deficit**: the bank can dip slightly negative when a block overshoots the budget; that debt is repaid before the gate reopens, so the long-run average still converges on `gas_per_second`.

Declared `gas_limit` is never consulted — it's commonly padded 2-3x, and using it would let an attacker submit huge-limit/low-cost transactions to close the gate almost for free. Charging only *executed* gas means the only
way to drain the bank is to pay for real execution.

Blocks sealed *before* the limiter started (WAL replay after a restart, external-node catch-up) are deliberately skipped — replaying old history at replay speed must not slam the gate shut for the whole sync window.

## Config

New optional section under `rpc`. Omit the whole `tx_gas_rate_limit` key to disable the limiter.

```yaml
rpc:
  tx_gas_rate_limit:               # omit this whole key to disable (default)
    gas_per_second: 64000000
    max_credit_seconds: 2.0         # default: 30.0
    reopen_credit_seconds: 1.0      # default
    deficit_floor_seconds: 2.0      # default
    exempt_senders: "0xAbc...,0xDef..."   # default: empty
```

| Field | Meaning | Default |
|---|---|---|
| `gas_per_second` | Target sustained throughput, in gas/s. Set to ~80% of the chain's measured max throughput. **Required if the section is present at all.** | — |
| `max_credit_seconds` | Bank capacity, in seconds' worth of `gas_per_second` — how much idle burst headroom is allowed. | `30.0` |
| `reopen_credit_seconds` | How much credit the bank must recover before the gate reopens after closing. | `1.0` |
| `deficit_floor_seconds` | How far into debt the bank may go before clamping; `0` disables debt tracking (clamp at zero). | `2.0` |
| `exempt_senders` | Comma-separated addresses that bypass the gate entirely (still consume bank capacity). | *(empty)* |

All three window values are expressed in *seconds' worth of `gas_per_second`*, so they scale automatically if the target rate is retuned — in practice only `gas_per_second` needs to be chosen per chain.

### Example

Chain benchmarked at ~80 MGas/s max throughput, targeting 80%, with two exempt partner addresses:

```yaml
rpc:
  tx_gas_rate_limit:
    gas_per_second: 64000000
    exempt_senders: "0x36615Cf349d7F6344891B1e7CA7C72883F5dc049,0x1234...5678"
```

Everything else uses the defaults above.

## Observability

New metrics under `tx_gas_rate_limiter_*`: current bank level (`bank_level_gas`, goes negative while a deficit is being repaid), gate open/closed state (`gate_open`), and a count of gate closures (`gate_closes`) — enough to see
the duty cycle live and tell whether `gas_per_second` is tuned too tight. `bank_level_gas` and `gate_open` update on every admission check as well as every block drain, so they stay current between blocks rather than only moving when one seals.

## Testing

Unit tests cover the bank's own refill/drain/hysteresis/deficit math, plus the `RateLimitedL2Subpool` wrapper's gating and draining behavior directly (exempt-sender bypass under a closed gate, the disabled-limiter no-op path, and that draining only applies to blocks sealed after the limiter is armed). An integration test (`gas_rate_limiter_closes_gate_and_recovers`) launches a real node, drives it into the closed state with a burst of transfers, and verifies the `-32005`/`retryAfterMs` response, that an exempt sender still lands transactions while closed, and that acceptance resumes after recovery.


lib/rpc/src/tx_gas_limiter.rs
```diff
@@ -0,0 +1,271 @@
...
+/// Rate limiter for incoming L2 transactions based on executed gas throughput.
+///
+/// A shared "gas bank" is drained by each sealed block's executed gas and refilled by
+/// wall-clock time at `gas_per_second`. The gate closes when the bank is exhausted and
+/// reopens once it recovers `reopen_credit`. Non-obvious properties:
+/// - The bank goes negative down to `-deficit_floor`: overshoot is repaid before
+///   reopening, keeping the long-run executed average at `gas_per_second`.
+/// - The drain is block-granular; overshoot admitted within a block self-corrects via
+///   the deficit. There is no per-transaction bookkeeping.
+pub struct TxGasRateLimiter {
```
itegulov: Thinking ahead, we'll have transactions gossiping once consensus gets merged and we'll start accepting transactions not only from RPC but also from p2p stack. So I think it makes more sense to make `TxGasRateLimiter` a part of the `mempool` crate. Consider making it an alternative implementation of `L2Subpool` trait - I think it will look pretty clean.

Moreover, we should consider how this policy will be enforced with consensus. Not sure if you've already had any thoughts on it, but anyway this is way more general and ambiguous than this PR so feel free to start a discussion somewhere else.

integration-tests/tests/rpc/gas_rate_limit.rs
```diff
@@ -0,0 +1,219 @@
...
+            &provider,
+            &params,
+            limited,
+            sink,
+            U256::from(1),
+            first_nonce + i,
+        )
+        .await
+        .expect("burst txs are admitted while the gate is open");
+    }
```
itegulov: Isn't there a race condition here that a block might get sealed before you send all 10 txs and hence you get -32005 early?

lib/rpc/src/tx_gas_limiter.rs
```diff
@@ -0,0 +1,271 @@
...
+            tracing::warn!(
+                from = expected,
+                to = number,
+                "tx gas rate limiter: missed block notifications, bank under-drained for this gap"
+            );
+        }
+        next_expected = Some(number + 1);
+
+        if notification.block.header.timestamp >= started_at {
+            limiter.on_block(notification.block.header.gas_used);
```
itegulov: This also takes any L1->L2 priority txs, upgrade txs and (in the future) interop txs into account. Maybe that's okay but just noting as the semantics might be a bit different from what you expect.

node/bin/src/config/mod.rs
```diff
@@ -1116,6 +1121,76 @@ impl From<RpcRateLimitsConfig> for zksync_os_rpc::RateLimits {
...
+/// Rate limiter for incoming L2 transactions based on executed gas throughput.
+/// Only effective on the main node.
+#[derive(Clone, Debug, DescribeConfig, DeserializeConfig, ConfigValidate)]
+#[config(validate(
+    Self::check_credit_windows,
+    "max_credit_seconds must be positive, other windows non-negative, and reopen_credit_seconds must not exceed max_credit_seconds"
+))]
+pub struct TxGasRateLimitConfig {
+    /// Target sustained executed-gas throughput, in gas per second.
+    pub gas_per_second: NonZeroU64,
```
itegulov, line 1135: Does this mean this field is mandatory? Is there no reasonable testnet default to be put here?

itegulov, line 1135: Also, just double checking: is "per second" a good mental model here? It's often harder to think in terms of such small periods of time.

node/bin/src/config/mod.rs
```diff
@@ -1116,6 +1121,76 @@ impl From<RpcRateLimitsConfig> for zksync_os_rpc::RateLimits {
...
+
+    /// Credit required to resume acceptance after the bank was exhausted, in seconds'
+    /// worth of `gas_per_second`.
+    #[config(default_t = 1.0)]
+    pub reopen_credit_seconds: f64,
+
+    /// Max remembered deficit, in seconds' worth of `gas_per_second`. Capacity consumed
+    /// above target while closing is repaid before reopening; `0` clamps the bank at zero.
+    #[config(default_t = 2.0)]
+    pub deficit_floor_seconds: f64,
```
itegulov: Are these defaults reasonable? They look too conservative to me

Also, if not too difficult I'd love to see how the system behaves on a load similar to what we have on GenLayer testnet for example

node/bin/src/config/mod.rs
vasyl-ivanchuk: I have a default value for `gas_per_second` in mind based on the load when backpressure kicked in. However, I want a kill switch for the feature to turn it off if it misbehaves. Another option is to keep a default for `gas_per_second`, but add an `enabled` field that defaults to `false`. Wdyt?

vasyl-ivanchuk: I couldn't come up with anything better than this. I was thinking about `gas per block` measures, but that conversion assumes blocks seal on a `block_time` cadence, which isn't always true. We could specify values `per_minute`, but I don't think that helps much. Happy to rename if you have better suggestions.

lib/rpc/src/tx_gas_limiter.rs
vasyl-ivanchuk: Great catch! I've implemented `RateLimitedL2Subpool` using the `L2Subpool` trait. As for consensus and rate limiting, I'll bring that up separately.

integration-tests/tests/rpc/gas_rate_limit.rs
vasyl-ivanchuk: Right, there is room for a race condition here. I've updated the test to send burst transfers in parallel and increased `block_time` to ensure the block isn't sealed early.

lib/rpc/src/tx_gas_limiter.rs
vasyl-ivanchuk: I think that's OK, I've added this to the comments. Lmk if you have anything in mind on how to make it clearer.

node/bin/src/config/mod.rs
vasyl-ivanchuk: I agree! I increased the `max_credit_seconds` default to `30`. It should tolerate bursts better while still handling steady load as expected. I'll play with the config more once it's deployed and maybe tweak a few things.

integration-tests/tests/rpc/gas_rate_limit.rs
itegulov: The fix shrank the window but didn't fix it. I asked my AI to reproduce try and reproduce it locally and there were 4 failures out of 11 runs. So this will make the test flaky.

I think a proper way to fix this is treating early -32005 as success and break the loop

vasyl-ivanchuk: Updated the test, now it waits for the rate limiter error (with a deadline as a safety measure).

node/bin/src/config/mod.rs
itegulov: I think adding `enabled` makes sense in this case yeah

lib/mempool/src/subpools/l2/mod.rs
```diff
@@ -61,6 +62,14 @@ pub trait L2Subpool:
         )
     }
 
+    /// Arms the (optional) executed-gas rate limiter: from this moment on, canonical blocks
+    /// drain its bank. Call once the node is caught up and about to start serving live
+    /// traffic — never before, since blocks committed during WAL replay / EN catch-up arrive
+    /// at replay speed and would otherwise double-charge the bank and pin the gate shut.
+    ///
+    /// No-op unless the limiter is enabled; safe to call multiple times or never.
+    fn arm_gas_rate_limiter(&self) {}
```
itegulov, line 71: The semantics of this method feel a little weird. I think it would be cleaner if we just had a mark whether a certain is replay or not. In any case, happy to merge as is for now.

itegulov (approved):

PR #1477, opened 2026-07-21
test: kill leaked anvil/prover processes

itegulov:
## Summary

Adds a helper process `leash` that ties the lifetime of a spawned process to the root process.

This should fix cases when anvil could leak and take a lot of RAM space just by continuously producing blocks every 250ms

### DIFF
--- a/integration-tests/src/bin/leash.rs
+++ b/integration-tests/src/bin/leash.rs
@@ -9,6 +9,10 @@
 //! On EOF: verify the target is still the process we were asked to kill (guards against
 //! PID reuse), send SIGTERM, wait up to `grace_secs`, then SIGKILL.
 //!
+//! This is intentionally a Rust binary rather than a shell script. Prover test artifacts are
+//! built on a non-GPU machine and transferred to a GPU runner, so keeping the watchdog in the
+//! Cargo build makes the transferred test archive self-contained.
+//!
 //! See `src/leash.rs` for the spawning side.
 
 fn main() {
--- a/integration-tests/src/bin/leash.rs
+++ b/integration-tests/src/bin/leash.rs
@@ -11,125 +11,122 @@
 //!
 //! See `src/leash.rs` for the spawning side.
 
-#[cfg(unix)]
 fn main() {
-    let args: Vec<String> = std::env::args().collect();
-    let (pid, grace_secs, expected_name) = match &args[..] {
-        [_, pid, grace, name] => (
-            parse_pid(pid).expect("pid must be a positive process ID"),
-            grace.parse::<u64>().expect("grace_secs must be an integer"),
-            name.clone(),
-        ),
-        _ => {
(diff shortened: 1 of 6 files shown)


integration-tests/src/bin/leash.rs
```diff
@@ -0,0 +1,135 @@
...
+//! write end is held (and never written to) by the process whose lifetime we track. The
+//! kernel closes that pipe on any kind of process death, so reading EOF here is a reliable
+//! death signal that needs no polling and no cooperation from the tracked process.
+//!
+//! On EOF: verify the target is still the process we were asked to kill (guards against
+//! PID reuse), send SIGTERM, wait up to `grace_secs`, then SIGKILL.
+//!
+//! See `src/leash.rs` for the spawning side.
+
+#[cfg(unix)]
```
popzxc: Instead of adding `cfg` on each item, maybe it's better to add `#[cfg(unix)] mod impl`?

And then inside of main have something 

```rust
#[cfg(unix)]
impl::run_main();
#[cfg(not(unix))]
panic!("Only unix is supported");
```

popzxc: Also not sure if this could've been a bash script to avoid having to compile it.

integration-tests/src/bin/leash.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// Best-effort check that `pid` still refers to the process we were told to kill.
+#[cfg(unix)]
+fn name_matches(pid: libc::pid_t, expected: &str) -> bool {
+    observed_name_matches(observed_name(pid).as_deref(), expected)
+}
+
+/// Matches an observed process name without guessing when no name could be read.
+#[cfg(unix)]
+fn observed_name_matches(observed: Option<&str>, expected: &str) -> bool {
```
popzxc: Single-use helper, can be inlined.

popzxc (commented):
Approach is OK, some pass-by nitpicks

integration-tests/src/bin/leash.rs
itegulov: Indeed, gated implementation in `mod unix`.

> Also not sure if this could've been a bash script to avoid having to compile it.

I also considered this but it makes test artifact portability more difficult (for prover tests we pre-build it on a non-GPU machine and then transfer it to a GPU machine to save on CI time). This _should_ be a one-time compilation for most cases, so hopefully not a big deal.

popzxc: Makes sense. I guess maybe it's worth a comment in the crate root doc comment -- otherwise other folks might be tempted to "optimize" it the same way I thought about it.

antonbaliasnikov (approved):

antonbaliasnikov:
I would think if unix protections are required, as we're not supporting Windows anyway.

PR #1482, opened 2026-07-22
feat(network)!: drop support for zks v1-v4

itegulov:
## Summary

Follow-up to #1409. This drops all pre-v5 `zks` versions thus getting rid of 2FA handshake and all 2FA-related messages in the `zks` subprotocol.

## Breaking Changes

This affects ENs: they must be updated to v0.20.12 to be able to still sync from MN.

## Rollout Instructions

Ensure all ENs (both internal and run by third parties) are upgraded to v0.20.12. Otherwise they will stop syncing.


lib/network/src/version.rs
```diff
@@ -86,79 +39,21 @@ impl ZksProtocolVersionSpec for ZksProtocolV5 {
...
     /// The `zks` protocol version 0. Only used for testing.
     Zks0 = 0,
-    /// The `zks` protocol version 1.
-    Zks1 = 1,
-    /// The `zks` protocol version 2.
-    Zks2 = 2,
-    /// The `zks` protocol version 3.
-    Zks3 = 3,
-    /// The `zks` protocol version 4.
-    Zks4 = 4,
```
popzxc, line 101: Probably it's worth documenting anywhere that legacy versions were present but are now removed.
Also not sure but _maybe_ it makes sense to document the version deprecation/removal policy, so that agents don't go "well if this can be removed, we're now gonna be removing stuff".

popzxc (commented):
Pass-by review; generally LGTM but would recommend getting a review from someone who works with the codebase day to day.

One optional addition could be an integration test where a peer only knows legacy protocols (so there is no intersection), if that's not hard to implement.

lib/network/src/protocol/handler.rs
```diff
@@ -162,14 +147,20 @@ impl<P: ZksProtocolVersionSpec, Replay: ReadReplay + Clone> ConnectionHandler
...
     ) -> OnNotSupported {
+        // `supported` is the negotiated intersection of both hello messages (exact name+version
+        // match on both sides), so a `zks` entry here means another locally-registered zks
+        // version owns this connection.
         if supported.iter_caps().any(|c| c.name() == ZKS_PROTOCOL) {
-            // Keep connection alive if there is at least one other common zks protocol version.
             OnNotSupported::KeepAlive
         } else {
-            // Disconnect otherwise.
+            tracing::debug!(
```
EmilLuta, line 171: Not sure how I feel about this. On one end, I'd like to keep it debug, because people can just choose not to update. On the other hand, it'd be very nice troubleshooting setup to know that at any given time we have outdated ENs connected. Maybe worthwhile to add DEBUG level for handler, whilst keeping it as DEBUG?

WDYT?

EmilLuta (dismissed):
Overall LGTM.

I know you have a section on rollout, but I'd do capitals and in red -- all ENS must be upgraded before protocol upgrade. I'm thinking of release procedure carrying on and us forgetting to notify an externally running EN. (think GenLayer)

lib/network/src/protocol/handler.rs
itegulov: Added a metric, this should strike a nice balance between having this log and not having it IMHO

EmilLuta (approved):

PR #1483, opened 2026-07-22
ci: gpg-sign release commits

dimazhornyk:
## Summary

Provide a GPG private key and a passphrase to a release-please workflow to support verified commits

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -27,7 +27,6 @@ jobs:
     uses: matter-labs/zksync-ci-common/.github/workflows/release-please.yaml@905a2a2dffe79187430cc0bc597db2bb0d2f1f9d # release-please-signed-commits
     secrets:
       slack_webhook: ${{ secrets.SLACK_WEBHOOK_RELEASES }}  # Slack webhook for release notifications
-      gh_token: ${{ secrets.RELEASE_TOKEN }}                # GitHub token for creating PRs and releases
       # Org-level secret; the App ID comes from the org-level MATTERLABS_BOTS_APP_ID
       # variable, resolved by the reusable workflow itself. API commits are auto-signed.
       MATTERLABS_BOTS_PRIVATE_KEY: ${{ secrets.MATTERLABS_BOTS_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -24,10 +24,13 @@ concurrency:
 jobs:
   # Prepare the release PR with changelog updates and create github releases
   release-please:
-    uses: matter-labs/zksync-ci-common/.github/workflows/release-please.yaml@c3b31c6f11bada204069c8b4621044e4c1cbc1df
+    uses: matter-labs/zksync-ci-common/.github/workflows/release-please.yaml@ba562a558063af2b224dab48798a99a56de8e8aa # release-please-signed-commits
     secrets:
       slack_webhook: ${{ secrets.SLACK_WEBHOOK_RELEASES }}  # Slack webhook for release notifications
       gh_token: ${{ secrets.RELEASE_TOKEN }}                # GitHub token for creating PRs and releases
```
antonbaliasnikov: I think we should be able to remove this secret with this PR already.

antonbaliasnikov (changes requested):

.github/workflows/release-please.yml
dimazhornyk: it is used as a fallback in case `MATTERLABS_BOTS_PRIVATE_KEY` is unset

antonbaliasnikov: Let's remove this fallback, it's insecure, and we should not have it anyway if we have gh app option

.github/workflows/release-please.yml
```diff
@@ -24,10 +24,13 @@ concurrency:
 jobs:
   # Prepare the release PR with changelog updates and create github releases
   release-please:
-    uses: matter-labs/zksync-ci-common/.github/workflows/release-please.yaml@c3b31c6f11bada204069c8b4621044e4c1cbc1df
+    uses: matter-labs/zksync-ci-common/.github/workflows/release-please.yaml@ba562a558063af2b224dab48798a99a56de8e8aa # release-please-signed-commits
```
antonbaliasnikov: Let's update to the latest sha 905a2a2dffe79187430cc0bc597db2bb0d2f1f9d

antonbaliasnikov (approved):

PR #1484, opened 2026-07-23
feat(zks): add stable getBatchByNumber and getLatestBatchNumber RPC

JackHamer09:
## Summary

Adds stable `zks_` RPC methods for enumerating batches:

- `zks_getBatchByNumber(batch_number) -> Option<PersistedBatch>`
- `zks_batchNumber() -> u64`
- `zks_getBatchByBlockNumber(block_number) -> Option<PersistedBatch>` — stable replacement for `unstable_getBatchByBlockNumber`, which keeps working for now

All three expose storage functions that already exist (`ReadBatch::get_batch_by_number`, `ReadBatch::latest_batch`, `ReadBatch::get_batch_by_block_number`) — no new indexing. This lets a consumer enumerate batches (latest number + fetch-by-number), unblocking re-adding a Batches page to the block explorer. Approved by protocol (Daniyar Itegulov).

`zks_batchNumber` mirrors `eth_blockNumber`. The batch getters reuse the existing `PersistedBatch` type (same one `unstable_getBatchByBlockNumber` returns) rather than introducing a new DTO, and return `Option` (`null` when absent) instead of erroring, so all three batch reads behave uniformly.

`ZksyncApi::wait_batch_number_by_block_number` now polls the stable `zks_getBatchByBlockNumber` and retries while it returns `null`, replacing the error-message string match against `unstable_`.

Includes an integration test (`rpc::batches::enumerate_batches`) covering: latest number covers a freshly finalized batch, fetch-by-number returns the expected batch, block-range membership, and out-of-range number → `None`.

### DIFF
--- a/lib/rpc_api/src/zks.rs
+++ b/lib/rpc_api/src/zks.rs
@@ -9,6 +9,7 @@ use alloy::rpc::types::Index;
 use jsonrpsee::core::RpcResult;
 use jsonrpsee::proc_macros::rpc;
 use zksync_os_genesis::GenesisInput;
+use zksync_os_storage_api::PersistedBatch;
 
 #[cfg_attr(not(feature = "server"), rpc(client, namespace = "zks"))]
 #[cfg_attr(feature = "server", rpc(server, client, namespace = "zks"))]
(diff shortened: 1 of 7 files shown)


lib/rpc_api/src/zks.rs
```diff
@@ -35,6 +33,12 @@ pub trait ZksApi {
     #[method(name = "getBlockMetadataByNumber")]
     fn get_block_metadata_by_number(&self, block_number: u64) -> RpcResult<Option<BlockMetadata>>;
 
+    #[method(name = "getBatchByNumber")]
+    fn get_batch_by_number(&self, batch_number: u64) -> RpcResult<Option<PersistedBatch>>;
+
+    #[method(name = "getLatestBatchNumber")]
+    fn get_latest_batch_number(&self) -> RpcResult<u64>;
```
itegulov: Maybe just `zks_batchNumber` to match Ethereum's `eth_blockNumber`?

lib/rpc_api/src/zks.rs
itegulov, line 1: Could you also move `unstable_getBatchByBlockNumber` to `zks_getBatchByBlockNumber`? Or at least add an alias for now, we'll drop support for `unstable_` RPC endpoint later.

itegulov (approved):

PR #1485, opened 2026-07-24
feat(l1_sender)!: pipelined sending with a bounded in-flight window

dimazhornyk:
## Summary

  The L1 sender used to send a wave of transactions and idle until all of them were included and
  confirmed — capping settlement at ~890 batches/hour on mainnet. It now keeps up to command_limit (16,
  the L1 pool's per-account cap) transactions in flight and tracks confirmations in a separate task;
  downstream forwarding is still confirmation-gated in batch order. A/B benchmark: 4x faster
  settlement.

  The old behavior is still available: l1_sender.pipelining_enabled = false.
  
  Living with a window of unconfirmed transactions required handling what the pool does to them:
  fee-priced-out sends wait for the market instead of crashing, evicted transactions are resent at
  their nonce with a 2x bump, and restart recovery pairs pending transactions with re-queued commands —
  replacing mismatched or dropped ones at their own nonces. force_transaction_resubmission now
  replaces stuck transactions instead of appending behind them.

  New integration tests cover the benchmark plus fee spikes, eviction, inclusion stalls, restarts
  mid-window, forced resubmission, and L1 connection loss.

 ## Breaking Changes
  
  - Main-node operators only; no protocol, DB, or wire changes.
  - Shutdown leaves in-flight transactions in the mempool — recovery picks them up on restart. Recovery
  pairing needs eth_getTransactionBySenderAndNonce (reth-family) and a single consistent L1 RPC view.

## Rollout Instructions
  
  - Ensure the L1 node's per-account pool limits cover command_limit (geth blobpool maxTxsPerAccount,
  reth max-account-slots; both default 16).
  - Deploy normally, no migrations. Watch the new txs_in_flight gauge; fee-stall and eviction-resend
  warnings during L1 fee spikes are expected.
  - Rollback: l1_sender.pipelining_enabled = false (no redeploy). If mismatched transactions are pooled
  at that point, restart once with force_transaction_resubmission.enabled = true.

### DIFF
--- a/lib/l1_sender/src/pipelined.rs
+++ b/lib/l1_sender/src/pipelined.rs
@@ -1,8 +1,8 @@
 //! Pipelined ("two-watermark") L1 sending.
 //!
-//! The stop-and-wait sender ties up the whole cycle waiting for inclusion + confirmations
-//! (~40-70s on mainnet) during which nothing new is submitted. This module decouples the
-//! three phases into concurrently running tasks joined by bounded queues:
+//! Submission, inclusion tracking, and confirmation run as concurrent tasks joined by
+//! bounded queues, so new transactions go out while earlier ones are still waiting to be
+//! mined or confirmed:
 //!
 //! ```text
 //! submitter ──(submitted queue: entries hold an unmined-slot permit)──▶ inclusion watcher
@@ -16,8 +16,8 @@
 //!   observed mined. This matches the L1 pool's *per-account* cap (geth blobpool
 //!   `maxTxsPerAccount` = reth `max-account-slots` = 16), which counts pooled (unmined)
 //!   transactions only. Each submission takes a semaphore permit that the inclusion watcher
-//!   releases at the *first receipt sighting* — never at confirmation, which would waste
-//!   `(required_confirmations-1)` blocks of window budget per tx and cap throughput at ~2x.
+//!   releases at the *first receipt sighting* — never at confirmation, which would keep each
+//!   window slot occupied for `required_confirmations - 1` extra blocks per transaction.
 //! * **Confirmation depth** — downstream forwarding still waits for `required_confirmations`,
 //!   handled by a separate FIFO task so it delays forwarding by a constant without ever
 //!   blocking submission.
@@ -307,7 +307,15 @@ where
             anyhow::Ok(())
         };
 
-        tokio::try_join!(submitter, inclusion_watcher, confirmation_forwarder)?;
+        // Boxed for stack safety, not style: these three state machines (the submitter
+        // embeds the whole wave-submission and recovery-seeding machinery) otherwise live
+        // inline in this future, which itself is moved by value when the pipeline spawns the
+        // component — that spike overflowed an 8 MiB thread stack in debug builds.
+        tokio::try_join!(
+            Box::pin(submitter),
+            Box::pin(inclusion_watcher),
+            Box::pin(confirmation_forwarder)
+        )?;
         Ok(())
     }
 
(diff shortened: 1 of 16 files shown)


lib/l1_sender/src/pipelined.rs
```diff
@@ -0,0 +1,1062 @@
...
+                    command_name,
+                    in_flight_count,
+                    "eth_getTransactionBySenderAndNonce is not supported by the current L1 \
+                     provider, so in-flight transactions cannot be paired with queued commands. \
+                     Waiting for them to settle before sending anything. For crash-tolerant \
+                     operation use a provider that supports the method (reth/Erigon family), or \
+                     restart with `force_transaction_resubmission.enabled = true` to replace \
+                     them.",
+                );
+                let settled_nonce = self.wait_for_account_quiescence(operator_address).await?;
```
itegulov: > quiescence

Nit: looks like AI pulled a pretty insane word here 😅

lib/l1_sender/src/pipelined.rs
```diff
@@ -0,0 +1,1062 @@
...
+        );
+
+        // Probe whether the provider supports `eth_getTransactionBySenderAndNonce` before
+        // iterating over all pending nonces.
+        if let Err(TransportError::ErrorResp(ref e)) = self
+            .provider
+            .get_transaction_by_sender_nonce(operator_address, latest_nonce)
+            .await
+        {
+            if e.code == METHOD_NOT_FOUND_CODE {
```
itegulov, line 841: This flow is not tested currently, right? I think adding one would be worth it for the sake of self-hosted environments

node/bin/src/batcher/mod.rs
```diff
@@ -245,11 +245,21 @@ impl<ReadState: ReadStateHistory + Clone + Send + 'static> Batcher<ReadState> {
...
+                        Some(true) if !blocks.is_empty() => {
                             // some of the limits was reached, start sealing the batch
                             break;
                         }
-                        Some(false) => {
+                        Some(seal_after_adding) => {
+                            // `seal_after_adding` means the batch is still empty and the peeked
+                            // block alone exceeds a seal limit. A batch must contain at least
+                            // one block — refusing it would replay the same block forever — so
+                            // accept it as a single-block batch.
```
itegulov, line 256: Is this something you encountered during the work on this PR? If so, then it must be a bug somewhere else, no? Block seal criteria should be more strict than the batch seal criteria.

itegulov (dismissed):
I have to be honest, I did not read the whole thing but overall it looks sane to me. Especially nice to see new integration tests, we are severely lacking in testing with flaky L1 scenarios.

node/bin/src/batcher/mod.rs
dimazhornyk: yes, it is a preexisting bug encountered when working on this PR

> If so, then it must be a bug somewhere else, no?

if you mean it is a bug that was in the code before this PR - that's true

> Block seal criteria should be more strict than the batch seal criteria

I remember that from the zksync-os perspective we are ok if block limits and batch limits are equal, and that's pretty much how we were setting block limits initially (the only limitation for block was that it has to fit in a batch)

lib/l1_sender/src/pipelined.rs
dimazhornyk: added a test

lib/l1_sender/src/pipelined.rs
dimazhornyk: yep, good catch, renamed

itegulov (dismissed):

itegulov (approved):

PR #1494, opened 2026-08-03
fix(config): refuse local block rebuilds under consensus

RomanBrodetski:
## Summary

Two recovery paths repair a bad block by re-running the chain with it rewritten or emptied — `sequencer.revm_consistency_checker_revert_on_divergence` (automatic) and `sequencer.rebuild`'s `BlockRebuild` / `DangerBlockRebuildWithL1Revert` modes (manual). Both change the block's hash and every descendant. Fine for a single sequencer, impossible under consensus: the block already carries a finalization certificate, so the rewrite forks this node off the chain instead of healing it.

Nothing rejected the combination, and it failed silently. A divergence panics the node, restart synthesizes a `BlockRebuild` from `internal_config.json`'s `failing_block`, `starting_block` moves back to the divergent block — but `blocks_to_empty` is only consumed by `ConsensusNodeCommandSource`, which the consensus pipeline replaces. So the node re-executes the same block unchanged, re-detects the same deterministic divergence, and panics again.

Two declarative validators now refuse it at startup. `load_internal_config` runs before `validate()` (`main.rs:191-192`), so the `rebuild` guard covers the synthesized case as well as an operator-set one.

**Unaffected:** detection (`revm_consistency_checker_enabled`, `revm_divergences_detected`) still alerts. `L1Revert` stays allowed — it touches only L1, and the disaster-fork runbook needs it.

**Test fallout:** the chaos rig and the integration tests inherit the auto-revert from the `local_dev.yaml` base layer, so both were already running the forbidden combination. The rig's generated overlay sets it false; the tests turn it off in `build_node_config` next to the existing `idle_heartbeat` override. Test-wide rather than per consensus scenario, because tests never call `validate()` — the guard can't catch a regression there, so it has to be structural.

## Notes

- Rejects a previously-accepted config, but nothing working can depend on it. Say the word if you'd rather it warn for one release.
- Worth checking whether the deployed consensus env sets this flag.
- `truncate-to` runs after `validate()` (`main.rs:196`), so a node with a stale `failing_block` marker can't be truncated until it's cleared by hand — friction on the recovery path the error message points at. Happy to exempt the subcommand from the `rebuild` guard here or in a follow-up.
- Longer term this probably wants deleting rather than guarding: a REVM divergence is committee-wide by construction, so a per-node self-heal was never the right shape. The manual modes still have users on non-consensus chains, so this is the interim.
- Unrelated: `cargo fmt --all` touches six files this PR doesn't, so the base branch isn't `--check` clean. Left alone to keep the diff scoped.

## Testing

5 new unit tests (auto-revert refused, detection-only allowed, local rebuild refused, `L1Revert` allowed, local rebuild still allowed without consensus). `cargo test -p zksync_os_server --lib config::` 27 passed; `cargo test -p zksync_os_chaos` 32 passed; clippy clean on both; `cargo check -p zksync_os_integration_tests --all-targets` clean.

### DIFF
--- a/integration-tests/src/test_config.rs
+++ b/integration-tests/src/test_config.rs
@@ -43,6 +43,14 @@ pub(crate) async fn build_node_config(
     // blocks): existing scenarios time their waits around continuous block
     // production. Dedicated idle-policy tests opt into heartbeats explicitly.
     config.consensus_config.idle_heartbeat = std::time::Duration::ZERO;
+    // `local_dev.yaml` arms the REVM auto-revert, which repairs a divergence by rewriting the
+    // offending block — refused under consensus, where the block carries a finalization
+    // certificate. Off for every test rather than per consensus scenario: the config validator
+    // that enforces this cannot catch a regression here, since tests never call `validate()`.
+    // Detection is untouched, so a divergence still logs and bumps `revm_divergences_detected`.
+    config
+        .sequencer_config
+        .revm_consistency_checker_revert_on_divergence = false;
     config.l1_provider_config =
         ProviderConfig::new(l1.address.clone(), TEST_PROVIDER_POLL_INTERVAL);
     if let Some(gateway_provider_config) = &mut config.gateway_provider_config {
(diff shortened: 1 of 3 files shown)


itegulov (approved):

RomanBrodetski:
It turned out this logic is already implemented in the newer version of consensus branch

PR #1498, opened 2026-08-04
fix(provider): fail hung L1 RPC requests with a retryable timeout

antonbaliasnikov:
## Summary

Fixes a silent external-node replay stall caused by hung L1 RPC requests.

**Root cause:** an L1 RPC request can hang forever on a half-dead connection (e.g. dropped by an intermediary without a RST). The RPC client stack has no request timeout, and `RetryService` only engages on errors it observes — a request that never returns is never retried. A hung call freezes the `priority_tx` watcher inside `process_event`, so the priority transaction is never inserted into the L1 subpool; when replay reaches the first block consuming it, `L1Subpool::pop_wait` blocks forever with zero log output, backpressure halts the whole pipeline, and the node keeps reporting `healthy: true` indefinitely.

**Changes:**

- `TimeoutService`: new innermost RPC client layer that fails any L1 request receiving no response within `l1_provider.request_timeout` (default 30s, per retry attempt). Timeout errors are classified retryable, so the retry lands on a fresh connection.
- `L1Subpool::pop_wait`: warns every 30s with the accumulated wait time while blocked on a missing priority transaction, so a stalled replay pipeline is visible in logs instead of silent. The periodic re-check also closes the lost-notification race between the queue check and `notified`.

**Note:** if all retry attempts time out, the error propagates to `Watcher::run`, which panics and restarts the node — loud and self-healing, versus an indefinite silent stall.

Validated during development with unit tests (hung request → timeout error → retryable classification; `pop_wait` warn cadence and wakeup semantics); tests were dropped from the final diff to keep it minimal, consistent with the rest of `provider/*`.

### DIFF
--- a/lib/mempool/src/subpools/l1.rs
+++ b/lib/mempool/src/subpools/l1.rs
@@ -3,9 +3,13 @@ use std::collections::VecDeque;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
+use std::time::Duration;
 use tokio::sync::{Notify, RwLock, mpsc};
 use tokio_stream::wrappers::ReceiverStream;
 use zksync_os_types::{L1PriorityEnvelope, L1TxSerialId, ZkTransaction};
+
+/// How often `pop_wait` warns while blocked on a priority transaction missing from the pool.
+const EMPTY_POOL_WARN_INTERVAL: Duration = Duration::from_secs(30);
 
 #[derive(Clone)]
 pub struct L1Subpool {
@@ -57,6 +61,7 @@ impl L1Subpool {
     }
 
     async fn pop_wait(&self) -> Arc<L1PriorityEnvelope> {
+        let started_at = tokio::time::Instant::now();
         loop {
             let notified = self.notify.notified();
             {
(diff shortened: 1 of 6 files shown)


lib/mempool/src/subpools/l1.rs
```diff
@@ -65,7 +70,17 @@ impl L1Subpool {
...
+            // and the calling replay/execution pipeline is blocked on it.
+            if tokio::time::timeout(EMPTY_POOL_WARN_INTERVAL, notified)
+                .await
+                .is_err()
+            {
+                tracing::warn!(
+                    waited = ?started_at.elapsed(),
+                    "L1 subpool has no pending priority transaction; blocked waiting for the L1 watcher to deliver it"
+                );
+            }
```
RomanBrodetski, line 83: hm.. does it also warn when there are genuinely no new priority transactions?

RomanBrodetski: seems like no - the `pop_wait` is only invoked when we know there is a block with pending priority transaction - so it must be somewhere there

RomanBrodetski (approved):

PR #1505, opened 2026-08-10
fix(prover_api): widen prover histogram buckets so quantiles stop clipping

LogvinovLeon:
`prove_time` clipped at 120s (`Buckets::LATENCIES`) and `prove_time_per_million_native` at 1.9s, so during an incident the mass lands in `+Inf` and quantiles saturate. Measured over 30d: FRI p99 is 2–15s but a fat batch took 601s; SNARK ~135s with a 208s tail.

| metric | before | after |
|---|---|---|
| `prove_time`, `prove_time_per_tx`, `prove_time_per_million_native` | top 120s / 1.9s | `PROVING_LATENCIES` — 10ms..10 000s, 1‑2‑5 boundaries |
| `computational_native_proven` | top 1.28G | 10M..100B (batches above 3.9G were clipping) |

Boundaries only — no renames, no label changes, `proved_after_attempts` untouched. No alert rule or dashboard reads these buckets, so nothing breaks.

Context: [PLA-1229](https://linear.app/matterlabs/issue/PLA-1229/increase-the-server-side-prover-job-timeouts) / [PLA-1209](https://linear.app/matterlabs/issue/PLA-1209/alert-p3-on-frisnark-proving-stall-batcher-vs-prover-queue-lag). Prover side: [zksync-airbender-prover#151](https://github.com/matter-labs/zksync-airbender-prover/pull/151).

### DIFF
--- a/node/bin/src/prover_api/metrics.rs
+++ b/node/bin/src/prover_api/metrics.rs
@@ -7,6 +7,11 @@ use vise::{
 const PROVER_JOB_LABELS: [&str; 3] = ["stage", "type", "id"];
 pub type ProverJobLabels = (ProverStage, ProverType, String);
 
+const PROVING_LATENCIES: Buckets = Buckets::values(&[
+    0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1.0, 2.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0, 500.0, 1000.0,
+    2000.0, 5000.0, 10_000.0,
+]);
+
 #[derive(Debug, Metrics)]
 #[metrics(prefix = "prover")]
 pub struct ProverMetrics {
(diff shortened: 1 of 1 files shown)


node/bin/src/prover_api/metrics.rs
```diff
@@ -7,6 +7,11 @@ use vise::{
 const PROVER_JOB_LABELS: [&str; 3] = ["stage", "type", "id"];
 pub type ProverJobLabels = (ProverStage, ProverType, String);
 
+const PROVING_LATENCIES: Buckets = Buckets::values(&[
+    0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1.0, 2.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0, 500.0, 1000.0,
```
RomanBrodetski, line 11: I don't think we need that low-valued buckets. 

I'd do `Buckets::exponential(0.5,2.0,10_000)` (I don't remember exact syntax)

LogvinovLeon: Those buckets are used in both per-tx and normal. That's why they have those low values

RomanBrodetski (approved):

Issue #77, opened 2025-07-21
Update to Rust 2024 edition


itegulov:
We should have been using 2024 edition from the start but let's fix it sooner rather than later

PR #1510, opened 2026-08-11
feat(prover_api): v32.0 / proving V8 - make real proofs settle on L1

dimazhornyk:
Everything the server needed for the v32.0 / V8 lane to settle, found by running it end-to-end.

**v32 ships at 100-bit.** `PROVING_SECURITY_LEVEL` and the zkos-wrapper feature moved together in matter-labs/zksync-airbender-prover#153 and matter-labs/zkos-wrapper#77; the server follows. The level selects the recursion verifier binaries, so it changes the recursion chain, the program commitment and the VK — the 80-bit values for the same binary are different and are not interchangeable.

**`V8_VK_HASH`** tracks the VK regenerated by #77. The server stamps this on batches and never reads the VK from L1, so a stale value silently mismatches the chain. Regenerated at 100-bit against `multiblock_batch.bin` md5 `8128c18a3b7145366b184e027d0e0f34`:

```
0xa81dc850a0724bcd62b7e5fbe60c62be32b4b45e33dd0d950f9c313e4684605a
```

**`V8_APP_END_PARAMS` and the verifier's security model.** `end_params` is derived from the app binary alone, so it is level-independent — but it was computed over a superseded binary (`3e19df8c…`), and `SECURITY` was still `Security80` while the prover emits 100-bit proofs. Either alone makes `expected_chain` wrong and every honest proof is rejected as *"proves a different program"*. Both now match the shipped binary and level.

**v32 batch public input** now matches on-chain (mirrors matter-labs/era-contracts#2381): commitment drops the leading `chain_id`; `chain_config_hash` folded in; multi-batch public input is the concatenation of all per-batch hashes, hashed once (a rolling fold coincides for N<=2 but diverges from N==3 on). `chain_config_hash` is byte-identical to the contract's (`0xa5bc74cb…3cf58e32` both sides).

**Explicit tokio stack size.** The v32.0 native batch PIG overflows the 2 MiB default and the server aborts with a stack overflow on its first batch. Only the in-process tests covered this path, so the standalone binary was never run against v32.0.

**`local-chains/v32.0` regenerated.** The snapshot predates era-contracts#2323 and shipped `verification_key: TBD`, with nothing registered at verifier version 8 — where the server routes V8 proofs. Now carries the V8 verifier at version 8, the #2381 Executor/Committer facets, and the patched dual verifier; `versions.yaml` records the real VK and contracts sha. `regenerate.sh` reproduces it.

### Validation

The batch public input, stack size and fixture changes were verified end-to-end at 80-bit: 2 batches, 43 txs, real FRI ×2 → SNARK → `proveBatches` on L1 (gas 527,261), `getTotalBatchesVerified() = 2`, both executed; and against the regenerated fixture, `committed=1, verified=1, executed=1` across 1/2/3-batch ranges (gas 519,001 / 527,261 / 534,697).

**The 100-bit constants above have not yet been through an end-to-end run** — the VK hash comes from `generate-vk --check-aux-params` and the `end_params` from airbender's `end_params` tool, both against the reproducibly-built V8 binary, and the program commitment was cross-checked against `ProgramProver::program_commitment()`. A 100-bit e2e is still outstanding.

Depends on matter-labs/era-contracts#2381 for the on-chain side.


lib/batch_types/src/batch_info.rs
```diff
@@ -267,7 +267,10 @@ impl PendingBatchInfo {
                 )
                     .abi_encode_packed(),
             )),
-            31 | 32 => B256::from(keccak256(
+            // v32 drops the leading chain_id - it is committed through the chain config hash
+            // in the outer public input instead (era-contracts#2323 does the same on-chain).
+            32 => self.v32_batch_output_hash(),
+            31 => B256::from(keccak256(
```
itegulov, line 268: Nit: lets swap the order of these two branches, current way looks counter-intuitive

node/bin/src/main.rs
```diff
@@ -88,8 +88,27 @@ fn load_config_defaults(config_sources: &mut ConfigSources, config_paths: Option
...
+/// overflows the 2 MiB default and aborts the server while sealing its first batch.
+/// Only a ceiling - nothing is allocated up front.
+const WORKER_THREAD_STACK_SIZE: usize = 256 * 1024 * 1024;
+
+pub fn main() {
+    // `RUST_MIN_STACK`, when set, wins, so constrained environments can still lower this.
+    let stack_size = std::env::var("RUST_MIN_STACK")
+        .ok()
+        .and_then(|value| value.parse::<usize>().ok())
+        .unwrap_or(WORKER_THREAD_STACK_SIZE);
```
itegulov, line 101: This gets overriden by `RUST_MIN_STACK` from `cargo/config.toml` when you run the node through `cargo run` (and also in tests). Should we bump it to 256MiB too for consistency?

node/bin/src/prover_api/fri_proof_verifier.rs
```diff
@@ -169,11 +169,14 @@ mod v8_verifier {
...
     use verifier_common::transcript::Blake2sBufferingTranscript;
 
-    const SECURITY: SecurityModel = SecurityModel::Security80;
-
-    /// `end_params` of the zksync-os v0.4.0 multiblock batch program, built reproducibly from
-    /// draft-0.4.0 @ 8ef47499 (md5 `8128c18a3b7145366b184e027d0e0f34`), computed with the
-    /// airbender `end_params` tool (`tools/cli`) at the pinned tag v0.6.0-rc.1.
+    /// v32 proves at 100-bit; this selects the recursion verifier binaries the chain below is
+    /// continued through, so it must match the prover's `PROVING_SECURITY_LEVEL`.
+    const SECURITY: SecurityModel = SecurityModel::Security100;
```
itegulov, line 174: I think v8 tests in the end of this file still assert 80-bit security constants

local-chains/v32.0/regenerate.sh
itegulov, line 1: I don't understand why we need this file. Shouldn't we patch zksync-os-integration-tests instead to generate v32.0 with the right verified from the very start?

lib/batch_types/src/batch_info.rs
```diff
@@ -267,7 +267,10 @@ impl PendingBatchInfo {
                 )
                     .abi_encode_packed(),
             )),
-            31 | 32 => B256::from(keccak256(
+            // v32 drops the leading chain_id - it is committed through the chain config hash
+            // in the outer public input instead (era-contracts#2323 does the same on-chain).
+            32 => self.v32_batch_output_hash(),
```
itegulov: Nit: comment in `v32_batch_output_hash` below is no longer accurate: "(`public_input_hash`) is intentionally left unchanged..."

lib/l1_sender/src/commands/prove.rs
```diff
@@ -92,7 +92,39 @@ impl ProofCommand {
...
+    /// `zksync_os_native_pig::v32_chain_config_hash`. Middle word is
+    /// `fri_proof_verification_enabled`, always disabled from L1.
+    fn zksync_os_chain_config_hash(chain_id: u64) -> B256 {
+        // EIP-7825 default cap (2^24), matching L1 and zk_ee.
+        const DEFAULT_MAX_TX_GAS_LIMIT: u64 = 1 << 24;
+        let mut bytes = Vec::with_capacity(32 * 3);
+        bytes.extend_from_slice(&U256::from(chain_id).to_be_bytes::<32>());
+        bytes.extend_from_slice(&U256::ZERO.to_be_bytes::<32>());
+        bytes.extend_from_slice(&U256::from(DEFAULT_MAX_TX_GAS_LIMIT).to_be_bytes::<32>());
+        keccak256(&bytes)
```
itegulov, line 106: Why not use `native_pig::v32_chain_config_hash` directly instead of copy-pasting code from there?

Overall, there is a lot of overlap in this file between this PR and https://github.com/matter-labs/zksync-os-server/pull/1451. But latter contains a different implementation (which is a bit cleaner IMHO, although I probably wouldn't save `chain_config_hash` inside `ProofCommand` is possible). Not sure if this was intentional or not

lib/batch_types/src/batch_info.rs
dimazhornyk: totally, fixed

node/bin/src/main.rs
dimazhornyk: bumped to 256

node/bin/src/prover_api/fri_proof_verifier.rs
dimazhornyk: yep, it is also ignored by default, so it wasn't visible in CI; fixed

local-chains/v32.0/regenerate.sh
dimazhornyk: makes sense, removed

itegulov (approved):

lib/l1_sender/src/commands/prove.rs
dimazhornyk: > Why not use native_pig::v32_chain_config_hash directly instead of copy-pasting code from there

that would require importing the whole zksync_os execution into L1 sender, not sure if it is worth it

> Why not use native_pig::v32_chain_config_hash directly instead of copy-pasting code from there?
> 
> Overall, there is a lot of overlap in this file between this PR and https://github.com/matter-labs/zksync-os-server/pull/1451. But latter contains a different implementation (which is a bit cleaner IMHO, although I probably wouldn't save chain_config_hash inside ProofCommand is possible). Not sure if this was intentional or not


the implementations are not interchangeable: the one from #1451 uses a rolling fold, while the correct v32 multi-batch public input is the full per-batch hashes concatenated and hashed once (because we are doing the aggregation differently)

and we have to keep both the new and the old one to support the previous version

PR #1513, opened 2026-08-12
fix(network): disconnect and re-request when the EN replay stream stalls

antonbaliasnikov:
## Summary

An external node's replay connection currently has no liveness protection. If the stream of `BlockReplays` messages silently dies while the RLPx session stays established, the EN parks on `conn.next()` forever — no warning, no reconnect — and stops syncing until the pod is restarted.

This is exactly what happened to `sequencer-c` on mainnet-memento on Aug 11–12: after a ~3-minute network disruption on the EN's node (22:12–22:17 UTC), the EN froze at block 176637 for **6 hours** until a manual restart. Evidence from the incident:

- The main node (`sequencer-a`) never logged a session close, kept the peer registered, and silently stopped advancing its replay cursor (visible via verify-request dispatch eligibility dropping from 2 peers to 1).
- The EN logged nothing after the (unrelated, already-fixed) L1 header watcher retries — no reconnect attempt, no `requesting block replays from main node` until the restart.
- A dead replay flow can hide behind a healthy-looking RLPx session: p2p-level pings keep flowing while the `zks` lane is dead, so reth never tears the session down and trusted-peer redial never kicks in.

## The fix

- After the EN sends `GetBlockReplays`, the connection task applies an inactivity timeout. Any inbound message resets the deadline.
- On expiry — or when the message stream terminates while the session lingers (e.g. after a decode error), a replay record cannot be recovered, or a record arrives out of sequence — the task reports the stall through a new `ProtocolEvent::ReplayStreamStalled { peer_id, next_block }`.
- The network service reacts by calling `disconnect_peer_with_reason(peer_id, PingTimeout)`, forcing the session down. The deliberately short peer backoffs (1–10s) redial, and the fresh session re-requests replays from the shared `starting_block`, i.e. exactly where the EN left off.
- New config knob: `sequencer.en_replay_inactivity_timeout`, default **300s**. It must comfortably exceed the chain's block cadence; on chains that can legitimately idle longer, operators should raise it to avoid periodic reconnect churn (a reconnect is cheap: one warn line + a re-request).

Worst-case sync stall shrinks from "until someone notices and restarts the pod" to the configured inactivity timeout.

## Testing

- New unit tests in `protocol/en.rs` (written first, watched fail):
  - idle stream → stall reported with the correct `next_block`;
  - inbound replays reset the deadline (cumulative time far beyond the timeout without a false positive), stall fires only after real silence;
  - stream termination → stall reported;
  - out-of-sequence record → stall reported instead of panicking the connection task.
- Full `zksync_os_network` unit + e2e suites and `zksync_os_server` tests pass; clippy `-D warnings` clean.
- Known gap: the disconnect → redial → re-request round trip runs through `NetworkService`, which the e2e harness doesn't instantiate, so that last mile is covered by reth's trusted-peer redial behavior rather than an automated test. After deployment, an inactivity timeout should be followed within seconds by `requesting block replays from main node` in the EN logs.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -15,6 +15,21 @@ ignore = [
     # never locally. Local `RsaPrivateKey` usage is test-only. No patched
     # version exists yet (tracked in RustCrypto/RSA#19).
     "RUSTSEC-2023-0071",
+    # `LruCache::pop()` in `lru` < 0.18.2 is not panic-safe: a cache key whose
+    # `Drop` panics under `catch_unwind` leaves dangling list pointers (potential
+    # use-after-free). Our only `lru` dependents are `alloy-provider` (keys:
+    # `u64`, `B256`) and `aws-sdk-s3` (string-like credential cache keys); no
+    # key type has a panicking `Drop`, so the unsound path is unreachable. The
+    # fix exists only in `lru` >= 0.18.2 while both dependents require `^0.16`.
+    # Drop this once they move their `lru` requirement past 0.18.2.
+    "RUSTSEC-2026-0253",
+    # `Bitmap::try_from(&[u8])` and `AsMut<[u8]>` in `bitmaps` can construct
+    # invalid backing-store values (UB); no patched version exists. Our only
+    # path to `bitmaps` is `imbl` (via `reth-transaction-pool`), which builds
+    # bitmaps exclusively through `new()`/`default()`/`from_value()` and never
+    # touches the unsound byte-slice conversions, so the UB is unreachable.
+    # Drop this if a patched `bitmaps` release appears or reth drops `imbl`.
+    "RUSTSEC-2025-0167",
 ]
 
 [bans]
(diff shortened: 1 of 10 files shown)


vasyl-ivanchuk (dismissed):

vasyl-ivanchuk:
@antonbaliasnikov LGTM, I made a small adjustment, feel free to check the last commit.

vasyl-ivanchuk (dismissed):

itegulov (commented):
Overall this makes sense to me but IMHO the wording around the fix should not mention watchdog

vasyl-ivanchuk:
> Overall this makes sense to me but IMHO the wording around the fix should not mention watchdog

@itegulov renamed _watchdog_ to _inactivity_:
`replay_watchdog_timeout` -> `replay_inactivity_timeout`
`watchdog_deadline` -> `inactivity_deadline`
etc.

itegulov (approved):

PR #1517, opened 2026-08-17
fix(sequencer): seal block instead of crashing on pubdata limit

vasyl-ivanchuk:
## Summary

Seal block instead of crashing when L1 tx hits a block resource limit:

Resource-exhaustion errors (`BlockPubdataLimitReached` et al.) are block-level seal signals, but the L1/Upgrade arm of the invalid-tx dispatch treated every `InvalidTransaction` as fatal. Since L1 priority txs are strict FIFO and re-derived from L1 on restart, a priority tx breaching the pubdata seal criterion produced a deterministic crashloop: dump -> exit 1 -> restart -> same tx -> same dump.

Now a seal-class error on a non-empty block seals the block and leaves the tx at the subpool head for the next block, mirroring the L2 seal path. Liveness is guaranteed by L1-enforced per-priority-tx caps (priorityTxMaxPubdata = 99k, priorityTxMaxGasLimit = 72M), which are below the fresh-block budgets. Genuinely invalid priority txs and seal-class errors on an empty block (tx can never fit; only reachable with misconfigured block limits) remain fatal.

### DIFF
--- a/lib/sequencer/src/execution/execute_block_in_vm.rs
+++ b/lib/sequencer/src/execution/execute_block_in_vm.rs
@@ -244,13 +244,36 @@ pub async fn execute_block_in_vm<V: ViewState>(
 
                         match (tx.tx_type(), command.invalid_tx_policy) {
                             (ZkTxType::L1 | ZkTxType::Upgrade, _) => {
-                                return Err(
-                                    BlockDump {
-                                        ctx,
-                                        txs: all_processed_txs.clone(),
-                                        error: format!("invalid {} tx: {e:?} ({})", tx.tx_type(), tx.hash()),
+                                match rejection_method(&e) {
+                                    // Block-resource exhaustion is a seal signal, not a verdict on
+                                    // the tx: seal what we have and let the tx retry from the
+                                    // subpool head in the next block. L1 caps per-priority-tx
+                                    // pubdata/gas below the block limits, so a lone tx in a fresh
+                                    // block always fits.
+                                    TxRejectionMethod::SealBlock(reason) if !executed_txs.is_empty() => {
+                                        tracing::info!(
+                                            block_number = ctx.block_number,
+                                            "Sealing block {} before {} tx {} because it hit a sealing criterion: reason={reason:?}, error={e:?}",
+                                            ctx.block_number,
+                                            tx.tx_type(),
+                                            tx.hash(),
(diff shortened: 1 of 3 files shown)


lib/sequencer/src/execution/execute_block_in_vm.rs
```diff
@@ -244,13 +244,36 @@ pub async fn execute_block_in_vm<V: ViewState>(
...
-                                    BlockDump {
-                                        ctx,
-                                        txs: all_processed_txs.clone(),
-                                        error: format!("invalid {} tx: {e:?} ({})", tx.tx_type(), tx.hash()),
+                                match rejection_method(&e) {
+                                    // Block-resource exhaustion is a seal signal, not a verdict on
+                                    // the tx: seal what we have and let the tx retry from the
+                                    // subpool head in the next block. L1 caps per-priority-tx
+                                    // pubdata/gas below the block limits, so a lone tx in a fresh
+                                    // block always fits.
```
itegulov: The comment is too verbose, I'd reduce it to just:
```
Seal what we have and let the tx retry from the subpool head in the next block.
L1 caps per-priority-tx pubdata/gas below the block limits, so a lone tx in a fresh block always fits.
```

lib/sequencer/src/execution/execute_block_in_vm.rs
```diff
@@ -244,13 +244,36 @@ pub async fn execute_block_in_vm<V: ViewState>(
...
+                                            tx.tx_type(),
+                                            tx.hash(),
+                                        );
+                                        break reason;
                                     }
-                                )
+                                    // Everything else is fatal: a genuinely invalid priority tx is
+                                    // a protocol violation (the FIFO cannot skip it), and a
+                                    // seal-class error on an empty block means the tx can never
+                                    // fit (block limits misconfigured below L1's per-tx caps).
```
itegulov: There are two different cases here:
1. Invalid priority tx which is a protocol violation
2. Block limits misconfigured below L1's per-tx caps

Let's do two different branches, in case of misconfigured block limits we don't need a block dump IMHO, just a error log

vasyl-ivanchuk: Good idea, added different branches as suggested.

itegulov (approved):

PR #1520, opened 2026-08-18
fix: shutdown teardown, L1 watcher revert classification and fake-SNARK lease return

antonbaliasnikov:
## Summary

Three independent robustness fixes in different subsystems. None of them changes
behaviour on the happy path — each turns a case that currently reports a crash,
trusts a wrong answer, or idles on a stale lease into the correct outcome. They
are independently revertable and can be reviewed in any order.

### 1. `fix(pipeline,batcher)`: survive an ordinary shutdown

Two ways a clean stop was reported as a crash.

- A pipeline segment that fails once graceful shutdown is under way has almost
  certainly just lost a downstream receiver that was dropped ahead of it.
  Panicking there reported a critical-task failure and left the runtime unable
  to finish shutting down. Both `select!` branches are ready at that moment, so
  which one wins is a coin flip rather than a rare interleaving. A segment that
  fails while the node is still *running* keeps panicking — `now_or_never` on
  the shutdown future is what separates the two cases.
- The blob sidecars the batcher hands the gas adjuster are fee-estimation
  samples. Failing that send failed the whole batcher segment, and the channel
  closes whenever the adjuster's task is gone — always on an external node, and
  during shutdown while batches are still being sealed. A lost sample now
  degrades a fee estimate instead of killing the segment.

### 2. `fix(l1_watcher)`: only an empty revert selects the configured supplier

A pre-v31 CTM has no `L1_BYTECODES_SUPPLIER()`, so the call reverts on an
unknown selector — which providers report as a JSON-RPC error response, not a
transport failure. The previous check treated *every* error response as "method
missing", so a rate limit (`-32005`), a provider internal error (`-32603`), or a
deliberately paused contract silently pinned the configured supplier as
canonical.

`is_revert_response` now matches the revert shape specifically, and a revert
that carries return data — a real revert reason, i.e. a contract deliberately
refusing — is **not** treated as a missing method. This is stricter than the
shared `is_method_missing` on purpose: there a wrong answer costs a retry, here
it decides which bytecode supplier the node treats as canonical.

### 3. `fix(prover_api)`: return the fake-SNARK lease on downstream backpressure

`pick_jobs_while_with_limit` assigns jobs before the loop reserves downstream
capacity, so a full channel abandoned the pick with the range still leased to
`fake_prover` — invisible to every other prover until the 300 s assignment
timeout, while the stage that applied the backpressure drains in milliseconds.
The lease is handed back instead.

## Testing

New unit tests:

- `only_reverts_select_the_configured_supplier` — covers all five error shapes
  (revert with and without data, rate limit, internal error, invalid params, and
  a transport failure).
- `backpressure_returns_the_fake_snark_lease` — one-slot downstream channel; the
  range must become pickable as soon as capacity returns rather than after the
  assignment timeout.

Full workspace verification is listed in a comment below.

## Notes for the reviewer

- `node/bin/src/prover_api/test_util.rs` is new shared test scaffolding
  (`create_test_batch_envelope`). The `prover_job_map` tests previously built the
  same fixture inline and now call it.
- `lib/l1_sender/src/commands/prove.rs` gains a single read-only accessor
  (`batches()`) so the backpressure test can assert on the emitted command. No
  encoding behaviour changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17201,6 +17201,7 @@ dependencies = [
  "futures",
  "rangemap",
  "reth-tasks",
+ "serde_json",
  "thiserror 2.0.17",
  "tokio",
  "tracing",
--- a/lib/l1_sender/src/commands/prove.rs
+++ b/lib/l1_sender/src/commands/prove.rs
@@ -27,6 +27,11 @@ impl ProofCommand {
         );
         Self { batches, proof }
     }
+
+    /// The batches this command settles, in range order.
+    pub fn batches(&self) -> &[SignedBatchEnvelope<FriProof>] {
+        &self.batches
+    }
 }
 
 impl SendToL1 for ProofCommand {
--- a/lib/l1_watcher/Cargo.toml
+++ b/lib/l1_watcher/Cargo.toml
@@ -31,5 +31,6 @@ backon.workspace = true
 rangemap.workspace = true
 
 [dev-dependencies]
+serde_json.workspace = true
 zk_os_api.workspace = true
 zk_os_basic_system.workspace = true
--- a/lib/l1_watcher/src/upgrade_tx_watcher.rs
+++ b/lib/l1_watcher/src/upgrade_tx_watcher.rs
@@ -566,9 +566,15 @@ impl L1UpgradeTxWatcher {
                 );
             }
             Err(e) => {
-                // Transport errors (503, timeout, etc.) should propagate.
-                // Contract-level reverts (function not found) are expected on pre-v31 CTMs.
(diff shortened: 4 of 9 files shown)


antonbaliasnikov:
### Local verification

Run on `aba-shutdown-watcher-lease-fixes` rebased on `main` (`331956187`):

- `cargo check --workspace --all-targets` — clean
- `cargo clippy --all-targets --all-features --workspace --exclude zksync_os_integration_tests -- -D warnings` — clean
- `cargo nextest run --workspace --exclude zksync_os_integration_tests` — **414 passed**, 4 skipped
- `cargo fmt --all -- --check` — clean

Integration tests were not run locally; CI covers them.

PR #1525, opened 2026-08-18
feat(prover): drop pre-0.3 proving support

vasyl-ivanchuk:
## Summary

This PR removes real proving support for protocols backed by ZKsync OS versions older than 0.3.x. It:
- Removes `ProvingVersion::V1` through `V6`, while preserving the numeric values of `V7` and `V8`.
- Removes legacy VK hashes and protocol-to-proving mappings.
- Removes pre-0.3 branches from prover-input generation, batch construction, FRI verification, prover job handling, and L1 verifier selection.

This is the first step toward replacing the independent `ProvingVersion` domain with an authoritative registry keyed by exact protocol versions.

## Backwards Compatibility

This change is backwards compatible under the assumption that no pre-0.3 batches still require proving or L1 submission. The intentional compatibility break is limited to real proving for pre-0.3 protocols. Such batches can no longer enter the proving pipeline, have their proofs verified, or select their former L1 verifier.

## Rollout Instructions

No coordinated prover-client deployment or data migration is required because the Prover API remains unchanged. Existing provers supporting the retained `V7` and `V8` VK hashes can continue operating normally.

### DIFF
--- a/lib/l1_sender/src/commands/prove.rs
+++ b/lib/l1_sender/src/commands/prove.rs
@@ -141,9 +141,6 @@ impl ProofCommand {
         let verifier_version = match self.proof.proving_execution_version() {
             // Use default verifier for fake proofs.
             None => 0,
-            Some(4) => 4,
-            Some(5) => 5,
-            Some(6) => 6,
             Some(7) => 0,
             // Switch to 0 once the L1 default verifier becomes the V8 one (as done for V7).
             Some(8) => 8,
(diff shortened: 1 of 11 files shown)


itegulov (approved):
Looks good overall, but this made me realize we actually mistakenly removed support for v30.1/v30.2 proving in #1421. It's still active on mainnets due to protocol upgrade delays there. So, we should bring it back otherwise new server releases will not be deployable there. Already flagged with @dimazhornyk, he will handle it.

I think we can merge this PR as is, @dimazhornyk please take it into account when you'll be bringing it back

PR #1527, opened 2026-08-19
feat(prover): bring back v30.x (V6) proving support

dimazhornyk:
## Summary

Some chains still run protocol v30.1/v30.2 and cannot upgrade yet, so the V6 proving lane removed in #1421 is restored:

- ProvingVersion::V6 (v30.1/v30.2, VK from zksync-os v0.2.5 + airbender v0.5.2 + zkos-wrapper v0.5.4) is back; v30.1/v30.2 are live again.
- The 0.2.10 lane regains zk_ee/basic_system deps for per-block PIG; app binaries are downloaded from the original v0.2.5 release whose binaries the V6 VK was generated from.
- Per-block and batch prover input generation, FRI proof verification (shared airbender 0.5.2 backend with V7) and the L1 verifier selection handle V6 again.
- The prover integration test regains a v30.2 lane (prover release v0.7.1).


Cargo.toml
```diff
@@ -134,10 +134,13 @@ zk_os_forward_system_0_1_2 = { package = "forward_system", git = "https://github
     # It was used with this feature, so it should stay here forever.
     "testing",
 ], default-features = false }
+# zksync-os 0.2.x proving lane (protocol v30.1/v30.2, proving version V6).
```
itegulov, line 137: This turned into a bit of a mess. Ideally in our model we were supposed to set:
`*_prev` deps to `v30`
`*` deps to `v31`
`*_dev` deps to `v32-rc.1`

But I really hope v30 won't be with us for much longer so let's merge as is

itegulov (approved):

PR #1532, opened 2026-08-20
feat(prover_api)!: move the V8 lane to the zksync-os v0.4.0 release

dimazhornyk:
## Summary

- zk_os_forward_system_0_4_0 / zk_ee_0_4_0 / zk_os_basic_system_0_4_0: v0.4.0-rc.2 -> v0.4.0 (69bc4305). The final tag's STF differs from rc.2 (EVM copy-opcode charge order #735, native-charging corrections #715, preimage-cache retention bound #720, modexp), so execution and the proved binary move together.
- V8_APP_END_PARAMS: recomputed for the release multiblock_batch.bin (md5 31cb9cb3b42d4a183fb858594eeb8706, built reproducibly from the tag) with the airbender end_params tool.
- V8_VK_HASH + local-chains v32.0 verification_key: 0xa81dc850… ->
  0x9f7576b911e7d3f528d49f894208682c81800814db9e3beac7fc3b1c4d626e7a (check_aux_params VK at 100-bit security binds the new binary).

Matches zksync-airbender-prover dz-v8-settlement-fixes @ 3bc29b0. The new VK must be re-registered on ZKsyncOSDualVerifier (verifier version 8) before real V8 proofs can settle.


itegulov (dismissed):

lib/batch_verification/src/verifier/mod.rs
dimazhornyk: I really don't like that we have to make the changes in verifier, but the final v0.4.0 added `prewarm_l1_postprocessing` to the pre-tx loop of every block, which fails to find the asset id of the base token, so we have to add some workarounds to genesis to make the tests work

itegulov (approved):

PR #1534, opened 2026-08-20
refactor(prover_api): replace proving versions with protocol registry

vasyl-ivanchuk:
## Summary

Replace the independent `ProvingVersion` domain with an authoritative proving registry keyed by exact `ProtocolSemanticVersion`.

Each registered protocol has one `ProvingConfiguration` containing:

- the API-facing verification-key hash;
- the prover-input generation strategy;
- the FRI proof format and application-binding parameters;
- the L1 verifier selector.

The registry currently supports:

- `0.30.1`
- `0.30.2`
- `0.31.0`
- `0.31.1`
- `0.32.0`

Unregistered protocol versions fail closed instead of falling back to a proving configuration selected by minor-version ranges.

## Naming

`V6` / `V7` / `V8` previously mixed several independent concepts under one ordinal. The new registry names each axis by what it controls:

| Previous model | Registry replacement | Meaning |
| --- | --- | --- |
| `ProvingVersion::V6/V7/V8` | exact `ProtocolSemanticVersion` lookup | which protocol configuration is active |
| version-based prover-input dispatch | `ZkOs0_2BlockInputs`, `ZkOs0_3BlockInputs`, `ZkOs0_4NativeBatch` | which zksync-os implementation produces the input |
| version-based FRI dispatch | `ProgramProof`, `UnrolledProof` | which proof format and verification path is used |
| `ProvingVersion::vk_hash()` | `ProvingConfiguration::verification_key_hash` | the VK advertised through the existing Prover API |
| carried `proving_version` values | `&'static ProvingConfiguration` | the resolved configuration and aggregation identity |

These names are introduced as part of the new registry; this PR deliberately does not perform a repository-wide terminology cleanup.

Existing names such as `apps::v6` / `apps::v7`, build-script environment variables, `v8_verifier`, ignored-test environment variables, and related local comments remain unchanged to keep this refactor focused.

### Runtime and wire compatibility

- **Prover API unchanged.** Provers continue exchanging `vk_hash` strings and proof bytes in the same request and response fields.
- **Prover deployment unchanged.** Existing prover binaries and VK hashes remain valid.
- **L1 calldata unchanged.** Registered configurations retain verifier selectors `6`, `0`, and `8`.
- **No database migrations.**
- **No runtime configuration or environment-variable changes.**
- **No proof-storage directory change is required.**

## Deployment

Deploy the server only. Provers do not need to be redeployed.

## Rollback

The server can be rolled back without changing prover deployments or L1 configuration.

Proof-storage files written by this version omit `proving_execution_version`. A pre-registry server cannot deserialize those files through the diagnostic `GET /SNARK/{from}/{to}/peek` endpoint, which may
return an error for recently stored batches after rollback.

This does not affect startup, prover job assignment, proof submission, or the commit/prove/execute path. The incompatible files will eventually rotate out of the bounded proof store.

## Validation

- `cargo test -p zksync_os_types proving_registry`
- `cargo test -p zksync_os_server prover_api::prover_job_map::map::tests`
- FRI verifier tests compile successfully


node/bin/src/prover_api/prover_job_map/models.rs
```diff
@@ -137,8 +144,9 @@ impl JobBatchStats {
 
         JobBatchStats {
             min_batch_number: min_batch.batch_number,
-            max_batch_number,
-            proving_version: min_batch.proving_version,
+            max_batch_number: max_batch.batch_number,
+            first_protocol_version: min_batch.protocol_version.clone(),
```
antonbaliasnikov, line 148: How does it work? Multiple protocol versions at the same time?

vasyl-ivanchuk: `JobBatchStats` is a logging summary over a slice of jobs. A slice can still hold two protocol versions on one proving stack (0.31.0 + 0.31.1 are both V7) - now they are reported accurately.

lib/types/src/protocol/proving_registry.rs
```diff
@@ -0,0 +1,342 @@
...
+    fri: FriProofConfiguration::V8 {
+        application_end_params: &V8_APP_END_PARAMS,
+    },
+    l1_verifier_selector: 8,
+};
+
+static REGISTRY_ENTRIES: [ProvingRegistryEntry; 5] = [
+    ProvingRegistryEntry {
+        protocol_version: ProtocolSemanticVersion::new(0, 30, 1),
+        configuration: &V6_PROVING_STACK,
```
itegulov: I think if you inline `V6_PROVING_STACK` here (and in other places), it would already look much cleaner - you'd essentially avoid calling this intermeidary struct as `V6`. I see that you reuse `V6_PROVING_STACK` in two places but this is actually the confusing part for me: v0.30.1 and v0.30.2 are not supposed to have the same VK and hence must have different configurations.

itegulov (commented):
Only leaving one general comment as I'd love to delegate the review to @antonbaliasnikov since he is already involved here

Ping me if you need a deeper review once general direction is ironed out

PR #1542, opened 2026-08-21
feat!: move the V8 lane to zksync-os v0.5.0 and the fix/87680 private…

dimazhornyk:
## Summary

Bump zksync-os to `v0.5.0`; use private crypto repos. 
NOTE: this PR completely disables interop traffic, and #1451 has to be merged to enable it back.


integration-tests/tests/prover_v8.rs
itegulov, line 1: Can we remove this whole file in favour of normal `prover.rs`? i.e. just add a new protocol version to the existing test

PR #1546, opened 2026-08-25
feat(v32): run the chain config with the pubdata content the chain holds on L1

StanislavBreadfulAI:
## Summary

v32 folds `pubdata_content` into the chain config hash and therefore into every batch's public input
(`Executor._getBatchProofPublicInputZKsyncOS`), but the node builds its config with
`ChainConfig::new(chain_id, false, DEFAULT_MAX_TX_GAS_LIMIT)` — always `FullPubdata`. The value was
never read from L1, so a chain configured `LOGS_ONLY` there would produce batches whose public input
can never verify, with nothing to say so. The zksync-os side already implements the mode fully
(`net_pubdata_used` → `net_committed_log_records_pubdata_used`, the logs-only block intrinsic, …);
only the server never selects it.

`LOGS_ONLY` is what a **ZKsync OS logs-only validium** runs with: it drops the state diffs and
message preimages but keeps publishing the mandatory L2→L1 log region — which carries the interop
commitment tree (IMT) leaves — through the same blobs a rollup uses. Same DA mechanism, narrower
content.

This PR makes the node run with the chain's own value:

- **`l1_discovery`** reads `getPubdataContent()` (and the chain's `l2DACommitmentScheme`, see below)
  from the diamond into `L1State`. The getter only
  exists on v32+ diamonds, so the protocol version is checked first rather than letting an unknown
  selector fail; anything older reports `FullPubdata`, which is exactly how a pre-v32 chain behaves —
  its batch public input carries no chain config hash at all.
- **The node pins that value once at startup**, before anything executes a block, and
  `v32_chain_config` builds every config with it.

## The startup check no longer uses pricing as a proxy for the DA mechanism

The `pubdata_mode` check compared the node's mode against the pubdata *pricing* mode read from L1
(`da_input_mode`), which was a stand-in for "what DA does this chain use". The two come apart in v32:
a logs-only validium publishes its (small) pubdata through blobs while pricing as a validium, and the
old check rejected that outright — the node would refuse to start on a perfectly well-configured
chain.

The invariant that actually matters is the one `Committer` enforces on every batch: a committed batch
must declare exactly the chain's `l2DACommitmentScheme`. The node now reads that scheme from
`getDAValidatorPair()` and requires its pubdata mode to produce it, which is both stricter (it catches
a genuine mechanism mismatch that pricing could not) and correct for a validium that publishes blobs.
Pricing goes back to being what it is — a fee-policy choice — and `da_input_mode` stays on `L1State`
for its metric.

## Why a process-wide pin

A node serves a single chain, and the value has to be identical in block execution, batch generation
and proof verification alike — `multivm::run_block` / `simulate_tx`, `generate_batch_run`, the
public-input reconstruction, the FRI/SNARK verifiers and the EN verifier all build the same config.
Threading it through each of those paths means touching the sequencer, the RPC sandbox and the prover
API; putting it in `BlockContext` (which already carries `chain_id` "temporarily … so that it can be
easily passed from the oracle") means a replay-record/wire version bump. Pinning it once keeps every
path consistent by construction, and L1 refuses to change the value while committed-but-unverified
batches exist, so a running node never observes it change.

Follow-up worth having on the record: a chain that ever *changes* its pubdata content would need the
value per block — i.e. in `BlockContext` — so that re-executing history stays exact. Nothing can
change it today without first verifying every committed batch, so this is not reachable yet.

## Behaviour for existing chains

Unchanged. v31 chains skip the read entirely; rollups read `FULL_PUBDATA`, which is what the node
already assumed. The only new work at startup is one `getProtocolVersion` call (plus one
`getPubdataContent` on v32+ chains).

## Contract side

matter-labs/era-contracts#2420 is the other half: it configures a ZKsync OS logs-only validium as
`BLOBS_ZKSYNC_OS` + the ZKsync OS blobs L1 DA validator + `pubdataContent = LOGS_ONLY`, leaving its
pubdata pricing free. matter-labs/zksync-os-integration-tests#36 carries the `zk-deployer` half.

## Testing

- `cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets` and
  `cargo fmt --check` clean.
- New `lib/native_pig/tests/chain_pubdata_content.rs` (own test binary, since the pin is
  process-wide): the default is `FullPubdata`, pinning `LogsOnly` reaches the chain config and
  changes its hash, re-pinning the same value is a no-op and a conflicting pin is refused.

### DIFF
--- a/lib/native_pig/src/lib.rs
+++ b/lib/native_pig/src/lib.rs
@@ -5,6 +5,7 @@
 
 use alloy::consensus::BlobTransactionSidecar;
 use alloy::primitives::{B256, keccak256};
+use std::sync::OnceLock;
 use zksync_os_batch_types::{BlockMerkleTreeData, CanonicalBatchCommitData, PendingBatchInfo};
 use zksync_os_merkle_tree::{MerkleTree, RocksDBWrapper};
 use zksync_os_storage_api::{ReadStateHistory, ReplayRecord};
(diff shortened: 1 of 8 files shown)


lib/native_pig/src/lib.rs
```diff
@@ -13,6 +14,40 @@ use zksync_os_types::{ProtocolSemanticVersion, ProvingVersion, PubdataMode};
...
 mod v32;
 
+pub use zk_ee_0_4_0::common_structs::da_commitment_scheme::PubdataContent;
+
+/// The chain's pubdata content, set once at node startup from the value the chain holds on L1.
+///
+/// The node serves a single chain, and the value is part of every v32 batch's public input through
+/// the chain config hash, so it has to be identical in block execution, batch generation and proof
+/// verification alike. Rather than threading it through every one of those call paths, it is pinned
+/// here once — the chain cannot change it while committed-but-unverified batches exist (the L1
```
StanislavBreadless: @itegulov note that a batch might not have been sent to l1, but be in progress. IIUC it is not an issue for zksync os, but FYI

StanislavBreadfulAI:
Superseded by #1530, which derives the pubdata content from L1 state and threads it explicitly through every chain-config construction site — the proper version of what this PR pinned in a process-wide `OnceLock`.

The one piece #1530 leaves inconsistent is the startup check, which still compares the node's pubdata mode against the pubdata *pricing* mode from L1 and therefore rejects a logs-only validium that publishes through blobs while pricing as a validium. That is now #1547, based on `di/chain-config-pubdata`.

PR #1547, opened 2026-08-25
feat: derive the node pubdata mode from the chain DA commitment scheme on L1

StanislavBreadfulAI:
## Summary

Builds on **#1530** (`di/chain-config-pubdata`), which derives the pubdata content from L1 state.
This PR does the same for the other DA axis — the publishing *mechanism* — so an operator no longer
configures either of them.

Two things were wrong with `l1_sender.pubdata_mode` as a required knob.

**It was checked against the wrong thing.** The startup check compared it with the pubdata **pricing**
mode read from L1 (`da_input_mode`), using pricing as a proxy for "what DA does this chain use":

```rust
(PubdataMode::Calldata | PubdataMode::Blobs, BatchDaInputMode::Validium)
| (PubdataMode::Validium, BatchDaInputMode::Rollup) => panic!(...)
```

The two axes come apart exactly in the configuration #1530 enables. A logs-only validium publishes
its (small) pubdata through blobs while pricing as a validium, so `(Blobs, Validium)` — a correct
setup — panicked at startup and the node refused to run.

**It did not have to be configured at all.** The invariant that matters is the one `Committer`
enforces on every batch: a committed batch must declare exactly the chain's `l2DACommitmentScheme`,
otherwise `MismatchL2DACommitmentScheme`. That scheme is on L1, so the node reads it from
`getDAValidatorPair()` into `L1State` and *derives* the pubdata mode from it. Leave
`l1_sender.pubdata_mode` unset and the chain's own configuration decides.

This is also what makes the interface honest for a logs-only validium: its operator would have had to
write `pubdata_mode: Blobs`, the same value a rollup writes, because the two differ in pubdata
*content*, not in mechanism. Now nobody writes either.

The config value remains as an override — for a chain whose scheme no ZKsync OS server produces
(`PUBDATA_KECCAK256`, third-party DA), and as an escape hatch. When set it must still agree with L1,
which is strictly more accurate than the old proxy in both directions:

- it accepts a blob-publishing chain regardless of how it prices pubdata;
- it catches mismatches pricing could not see — `Blobs` vs `Calldata` (both "rollup priced", and today
  that mismatch only surfaces when the first commit reverts).

`L1State::l2_da_commitment_scheme` is an `Option`: a diamond that predates the scheme either has no
getter or returns the L2 DA validator address where the enum now sits, and neither decodes. Such a
chain falls back to the configured mode, exactly as before this PR — which is why the pre-v32
`local-chains/v30.2` fixtures keep their `pubdata_mode` while the v31/v32 ones drop it.

`da_input_mode` stays on `L1State` for its metric; pricing goes back to being what it is, a fee-policy
choice.

## Contract side

matter-labs/era-contracts#2420 configures a ZKsync OS logs-only validium as `BLOBS_ZKSYNC_OS` + the
ZKsync OS blobs L1 DA validator + `pubdataContent = LOGS_ONLY`, and leaves its pubdata pricing free
(`DAValidatorType::NoDA` is renamed `LogsOnlyValidium` there for the same reason this knob needed
fixing — pricing and mechanism are not the same axis). Its `protocol-ops` side follows the same rule:
callers pick a DA mode and the commitment scheme is derived.
matter-labs/zksync-os-integration-tests#36 carries the `zk-deployer` half, where the whole choice is
one `da_mode: rollup | logs_only_validium | avail` field in `intent.yaml`.

## Testing

`cargo check --workspace --all-targets`, `cargo clippy --workspace --all-targets` and
`cargo fmt --check` clean on top of #1530; `cargo test -p zksync_os_types` covers the new
scheme↔mode round trip and the schemes that have no mode.

### DIFF
--- a/lib/batch_verification/src/verifier/mod.rs
+++ b/lib/batch_verification/src/verifier/mod.rs
@@ -754,7 +754,7 @@ mod tests {
             finalized_l1_block_number: 0,
             da_input_mode: BatchDaInputMode::Rollup,
             pubdata_content: PubdataContent::FullPubdata,
-            l2_da_commitment_scheme: DACommitmentScheme::BlobsZKsyncOS,
+            l2_da_commitment_scheme: Some(DACommitmentScheme::BlobsZKsyncOS),
             l1_chain_id: SL_CHAIN_ID,
         }
     }
--- a/lib/batch_verification/src/verifier/mod.rs
+++ b/lib/batch_verification/src/verifier/mod.rs
@@ -312,7 +312,9 @@ mod tests {
     use zksync_os_batch_types::BlockMerkleTreeData;
     use zksync_os_batch_types::PendingBatchInfo;
     use zksync_os_batch_types::batcher_model::{BatchEnvelope, BatchMetadata, ProverInput};
-    use zksync_os_contract_interface::models::{BatchDaInputMode, StoredBatchInfo};
+    use zksync_os_contract_interface::models::{
+        BatchDaInputMode, DACommitmentScheme, StoredBatchInfo,
+    };
     use zksync_os_contract_interface::{Bridgehub, ZkChain};
     use zksync_os_genesis::{FileGenesisInputSource, GenesisState, build_genesis};
     use zksync_os_interface::traits::{PreimageSource, ReadStorage};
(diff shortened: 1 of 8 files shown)


StanislavBreadfulAI:
Superseded by #1551, which carries the same fix on top of `main` rather than stacked on #1530 — the derivation does not depend on the pubdata-content work, and the check it replaces is wrong on `main` today. Closing this one.

PR #1548, opened 2026-08-26
feat(storage): dual-write phase 1 of replay WAL unification

RomanBrodetski:
## Summary

Phase 1 (dual-write) of unifying the replay WAL's ~13 column families into 4. Adds the target schema and writes into it unconditionally; nothing reads it yet, and no existing CF or read path changes.

## Target schema (end goal)

```
Block          number(8B) ++ hash(32B)  ->  everything but the 256-hash window:
                                             BlockContext fields, parent_hash, transactions,
                                             node/protocol version, block_output_hash,
                                             force_preimages, starting_cursors,
                                             previous_block_timestamp
CanonicalHash  number(8B)               ->  hash(32B)     -- canonical chain index (exists today)
HashToNumber   hash(32B)                ->  number(8B)    -- reverse index, write-once
Latest         fixed key                ->  latest number
```

Reads:
- canonical block N: `CanonicalHash[N] -> hash`, then `Block[(N, hash)]`. Window: one batched `CanonicalHash[N-256..N-1]` (unchanged mechanism).
- a specific `(number, hash)` (override sync, historical `eth_call`): direct point lookup, no index needed.
- hash alone: `HashToNumber[hash] -> number`, then `Block[(number, hash)]`.

Writes:
- append or override are the same operation: write the row at `(number, hash)`, repoint `CanonicalHash[number]`. The old row simply stays at its own key, untouched — no archival step, no wave-tracking cache, no window-reconstruction walk.

### Why `number ++ hash`, not just `hash`

Hash bears no relation to block number, so a pure hash key would turn range scans over canonical history (bulk EN sync, `replay_archive` recovery) from one sequential pass into scattered point lookups. The composite key preserves that locality; `HashToNumber` buys back hash-only lookup for ~40B/row. Nothing actually needs hash-only lookup today — `eth_getBlockByHash` reads from `RepositoryDb`, not the replay WAL — but it's cheap enough to keep.

## This PR

Just the dual-write: `write()` additionally writes `Block` and `HashToNumber` alongside every existing CF. A bug here can't affect production since nothing reads the new schema yet.

Next: backfill existing history into the new schema + verify old/new decode equivalently, then cutover reads, then drop the old CFs.

## Testing

- `dual_write_populates_unified_block_row`: every written block's `Block`/`HashToNumber` entries decode back correctly.
- `dual_write_override_keeps_old_row_and_adds_new`: overriding a block leaves the old row's entry intact at its own key; both old and new hash resolve to the same number.
- `cargo fmt`, `clippy -D warnings`, full workspace test suite (419 tests) green.


RomanBrodetski:
Moving to matter-labs/zksync-os-server-private#33 — the team's origin for this repo moved there.

